PR #61, opened 2026-02-07
improve prune handling for edit tool calls

### LINKED ISSUE #59: further improve pruning for edit tool calls
edit tool calls currently require old + new text which are then string-replaced. this can eat tokens a good amount. when pruning (`/prune:*`), we should always prune edit tool calls so that we replace old and new text arguments with a short truncation marker and replace the tool result with a minimal diff between the old/new text.

should also trim unchanged chunks like we do for compaction now.

markusylisiurunen:
- redact `edit` tool call `oldText` and `newText` arguments during `/prune:*` runs
- replace matching `edit` tool results with compact diffs and collapse unchanged chunks with 4-line context
- keep prune fraction `0` as a no-op and update prune messaging/hints to cover edit pruning
- add chat controller tests for edit pruning behavior and fraction `0` no-op

fixes #59


markusylisiurunen:
@codex review

PR #169, opened 2026-02-21
wrap default subagent prompt around inherited main prompt

### LINKED ISSUE #162: should the default subagent inherit the main agent's system prompt?
perhaps with some additions about it being a subagent etc.

markusylisiurunen:
- make default subagent prompt composition explicit by wrapping inherited main persona system prompt in a dedicated section
- keep full prompt rebuild path unchanged so skills, AGENTS context, sandbox info, and environment tags are resolved from the subagent cwd/risk
- preserve custom subagent behavior, only the built-in `default` subagent gets inherited main-prompt wrapping
- add tests that assert inherited main prompt inclusion in default subagent prompts
- update README.md and AGENTS.md to document default subagent prompt inheritance and precedence rules

fixes #162

### DIFF
--- a/src/core/static/prompts/default-subagent-wrapper.md
+++ b/src/core/static/prompts/default-subagent-wrapper.md
@@ -1,0 +1,20 @@
+You are a subagent supporting the main agent.
+
+Treat the instructions in <inherited-instructions> as baseline behavior.
+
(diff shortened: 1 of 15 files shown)


src/core/static/prompts/default-subagent-wrapper.md
```diff
@@ -0,0 +1,15 @@
+You are a subagent supporting the main agent.
```
markusylisiurunen: double-check this

PR #185, opened 2026-03-26
add diff review demo tool

markusylisiurunen:
- add blocking `/diff` review sessions with a dedicated Tau diff-review protocol and frozen git diff snapshots
- ship a built-in browser demo tool as `tau diff-tool`, used by default when `diffTool` is not configured
- cover the new config, protocol, controller flow, built-in tool, and lifecycle edge cases with tests

fixes #181


src/core/diff_review/review_thread.ts
```diff
@@ -0,0 +1,195 @@
...
+
+export class DiffReviewThread implements DiffReviewThreadSession {
+  private readonly session: CoreSession;
+  private readonly runtime: ConversationTurnRuntime;
+  private readonly personaId: string;
+  private readonly reasoningEffort: string;
+
+  constructor(options: CreateDiffReviewThreadOptions) {
+    const deps = options.deps ?? createDefaultCoreDeps();
+    const backend =
```
markusylisiurunen: [P1] Keep diff-review threads inside the session sandbox.

This path always falls back to `createLocalToolExecutionBackend(...)`, and the controller never passes `toolExecutionBackend`, so `/diff` review threads run their bash/view-image tools against the host filesystem even when the main Tau session was started with `--sandbox`. On top of that, the prompt bootstrap and `CoreSession` both hard-code `sandboxEnabled: false`, so the model also sees host paths instead of the mounted sandbox paths. That breaks the project sandbox guarantees for exactly the workflow that is supposed to review repository code. We should thread the existing sandbox backend and sandbox cwd through the diff-review session instead of constructing a fresh local backend here.

src/core/diff_review/snapshot.ts
```diff
@@ -0,0 +1,423 @@
...
+}
+
+export async function captureDiffReviewSnapshot(
+  options: CaptureDiffReviewSnapshotOptions,
+): Promise<DiffReviewSnapshot> {
+  const deps = createDiffSnapshotDeps(options.deps);
+  const cwd = resolve(options.cwd);
+  const diffArgs = [...(options.diffArgs ?? [])];
+  const repoRoot = await resolveGitRepoRoot(cwd, deps);
+  const patch = await runGitCommand(cwd, ["diff", ...diffArgs], deps);
```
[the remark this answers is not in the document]
markusylisiurunen: we can rely on the user doing the right thing for now

.github/workflows/ci.yml
```diff
@@ -17,7 +17,7 @@ jobs:
           registry-url: https://registry.npmjs.org
           cache: npm
 
-      - run: npm ci
+      - run: npm ci && (cd src/diff_tool/app && npm ci)
```
markusylisiurunen, line 20: hey

PR #192, opened 2026-04-07
add anthropic token counting backend

### LINKED ISSUE #191: allow using Anthropic's token counting endpoint instead of `num_bytes / 5` heuristic
see https://platform.claude.com/docs/en/build-with-claude/token-counting.md for docs. this should be used mainly for tool result token counting where we now use the `num_bytes / 5` (or whatever the denominator is). we should allow setting token counting to use this via a setting. if set to use this endpoint and no anthropic api key, error early.

markusylisiurunen:
- add a shared token counting abstraction with heuristic and anthropic backends
- use Anthropic's count_tokens endpoint via fetch for token-bounded truncation and sizing
- fail startup and reload early when anthropic token counting is configured without an Anthropic API key
- update tests and config docs

fixes #191


markusylisiurunen:
Closing without merging. This adds too much complexity for the value it provides, and token counts are model-specific, so using Anthropic's endpoint would not give us an accurate general-purpose count across the models Tau supports.

PR #237, opened 2026-06-27
preserve compaction user messages

### LINKED ISSUE #228: compact (or at least auto-compact) tends to lose the initial user message or the "goal" of the work
it should obviously be preserved in order to be able to pursue the same goal.

markusylisiurunen:
- tighten compaction prompts around continuation and current/original goals
- append selected original user messages verbatim inside compaction summaries by history id
- cover manual and automatic compaction metadata and docs

fixes #228


src/core/session/compaction.ts
```diff
@@ -17,20 +18,29 @@ export type CompactionHistoryEntry = {
...
   messagesToSummarize: Message[];
 };
 
 export type SessionCompactionMessageResult = {
+  summary: string;
   compactionMessage: string;
   includedLastAssistant: boolean;
 };
 
+const PRESERVED_USER_MESSAGE_MAX_TOKENS = 20_000;
```
markusylisiurunen, line 51: if the selected user messages exceed this, how are they truncated?

PR #259, opened 2026-07-06
add Nook v0

### LINKED ISSUE #240: a shopify-like quick implementation tau can use to deploy interactice sites or mini-apps
https://shopify.engineering/quick

markusylisiurunen:
## why

Issue #240 asks for a constrained, Tau-integrated V0 of Nook: a Cloudflare-backed static mini-app platform that Tau can deploy to and operate through both CLI commands and a model tool.

The V0 still needs to be trustworthy enough to operate against a real Cloudflare deployment, so the implementation also treats teardown, upload integrity, deployment limits, and Cloudflare Access validation as part of the initial contract rather than follow-up compatibility work.

## what

This adds the bundled Nook Worker stack, `tau nook` CLI, Tau config integration, and a single configured `nook` model tool. The Worker serves static deployments from R2, stores registry/site/KV metadata in Durable Objects, injects `window.nook`, exposes `/__nook/skill`, and enforces the V0 reserved-path, slug, manifest, visibility, and JSON KV rules.

Setup now writes the Cloudflare Access team domain and application audience into the Worker environment, and protected Worker routes validate Access JWTs against the Access JWKS, issuer, audience, expiry, and signature. Raw Access service-token headers are only used to pass Cloudflare Access; the Worker does not trust them directly.

Destroy now calls an authenticated Worker cleanup endpoint before deleting infrastructure, so site Durable Object state and R2 objects are removed before Wrangler deletes the Worker and R2 bucket. Deploy uploads are checked against manifest byte sizes and SHA-256 hashes, and each site has a small cap on non-expired pending deploy sessions.

It also adds focused tests for validation, setup/destroy parsing, config secret precedence, and model-tool risk/config gating, plus Nook-specific docs and package asset copying so the Worker docs ship with Tau.

## details

Nook setup and destroy remain infrastructure flows and take their own flags/env because they need Wrangler credentials and Cloudflare Access setup/cleanup inputs. Normal runtime operations use the single effective configured target. Deploys are private by default, `--public` is per deployment, and model-tool deploys read from the session execution backend so hosted execution environments use the same generic boundary as local sessions.

fixes #240


markusylisiurunen:
Cloudflare cost/spend exposure follow-up: public Nook deployments can generate Worker/R2/Durable Object usage, and Cloudflare does not provide a straightforward Tau-enforced spend cap/budget. The current branch has deploy/storage quotas and asset cache headers, but no request-rate or cost guardrails. Public browser KV is also public writable, so anonymous traffic can drive Worker + Durable Object work.

Before treating public Nook deployments as safe operationally, we should decide whether to add mitigations such as docs warnings, recommended Cloudflare WAF/rate limiting/bot protections, Worker-side request/write throttles, or a stricter public-KV opt-in/disable mode.

markusylisiurunen:
Consider using Cloudflare `HTMLRewriter` for browser SDK injection. The Worker currently reads HTML with `object.text()` and does a string replacement for exact lowercase `</body>`, appending the script otherwise. That is workable for small V0 apps, but `HTMLRewriter` would be more Worker-native, streaming-friendly, and less brittle for injecting `/__nook/client.js`.

markusylisiurunen:
Large cache correctness issue: non-index assets are served with `cache-control: public, max-age=31536000, immutable`, but their public URLs do not include the deployment id. R2 keys include the deployment id internally, but browsers/CDNs see the same URL across deployments, for example `/assets/app.js`.

If a new deployment uses the same asset filename with different content, clients may keep serving the old cached asset while `index.html` updates. This is only safe when all assets are content-hashed. We should either make the cache policy conservative for same-path assets, document/enforce content-hashed asset names, or otherwise include a deployment/content version in externally visible asset caching semantics.

PR #279, opened 2026-07-10
preserve cumulative session cost

### LINKED ISSUE #278: auto-compact zeroed the cumulative session cost

markusylisiurunen:
## why

Session cost was derived from the assistant messages still present in history. Manual or automatic compaction replaces part or all of that history, so the TUI could show a lower or zero cumulative cost after compaction.

## what

The host now owns and persists cumulative session cost in the canonical session snapshot. Main-assistant and subagent cost updates flow through a typed `cost.set` delta, and TUI and Telegram status rendering consume the snapshot total instead of recomputing it from retained messages.

fixes #278


markusylisiurunen:
Two changes before merge:

- Rebase/update against current `main` and adapt the newer tier-pricing footer test to the authoritative snapshot cost by setting `costTotal: 0.56`. I tested the merge result locally: the build passes and 661/662 tests pass, with only `includes tier-priced assistant usage in the session footer cost` failing (`$0.00` vs `$0.56`).
- Remove the now-obsolete client-side aggregate cost API: `SessionChatView.getToolUiCostTotal()`, its `TuiChatView` implementation, `SubagentPanel.getCostTotal()`, and the fake-view `toolCost`/method in the controller tests. Session cost now comes exclusively from `snapshot.costTotal`.

PR #309, opened 2026-07-17
improve telegram poll error logs

markusylisiurunen:
## why

Telegram polling warnings currently discard the structured failure cause at the runtime boundary, leaving every API, response, and network failure as the same `telegram poll failed` line. This makes recurring failures difficult to diagnose and can require code-level investigation to distinguish a duplicate poller from a transient connection problem.

## what

Preserve concise failure causes in Telegram runtime output and make API request errors identify the Telegram method, HTTP status and description, or nested network error code. Unexpected response details are bounded so warnings remain single-line and low-noise.

The adapter coverage now includes Telegram conflict responses and nested network failures, while runtime coverage verifies that structured causes reach the rendered log line.

### DIFF
--- a/src/core/telegram/runtime.ts
+++ b/src/core/telegram/runtime.ts
@@ -43,9 +43,26 @@ type RuntimeResources = {
   sessionManager: TelegramSessionManager;
 };
 
+type RuntimeLogEntry = {
+  level: string;
+  message: string;
+  data?: unknown;
+};
+
 const defaultDeps: TelegramRuntimeDependencies = {
   startTelegramAdapter: startTelegramAdapter,
 };
+
+function formatRuntimeLog(scope: string, entry: RuntimeLogEntry): string {
+  const cause =
+    typeof entry.data === "object" &&
+    entry.data !== null &&
+    "cause" in entry.data &&
+    typeof entry.data.cause === "string"
+      ? entry.data.cause.trim()
+      : "";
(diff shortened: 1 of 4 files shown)


markusylisiurunen (commented):
The structured causes now reach the CLI and the tested Telegram conflict/network cases look correct, but the new formatter does not yet uphold the promised single-line log contract.

src/core/telegram/runtime.ts
```diff
@@ -39,10 +39,27 @@ type RuntimeResources = {
...
   startTelegramAdapter: startTelegramAdapter,
 };
 
+function formatRuntimeLog(scope: string, entry: RuntimeLogEntry): string {
+  const cause =
+    typeof entry.data === "object" &&
+    entry.data !== null &&
+    "cause" in entry.data &&
+    typeof entry.data.cause === "string"
+      ? entry.data.cause.trim()
```
markusylisiurunen: Please normalize embedded line breaks before appending the cause. `trim()` only removes leading/trailing whitespace, so an unexpected multiline HTTP/proxy body (the fallback in `readTelegramHttpErrorDetail`) or a multiline network error is emitted as multiple physical log lines. That defeats the stated single-line behavior and can make response text look like separate runtime events. Collapsing CR/LF here (and covering it in the runtime/error-detail tests) would enforce the contract for both adapter and session-manager causes.

PR #310, opened 2026-07-18
add local session environment overrides

markusylisiurunen:
## why

WebSocket SDK clients need to isolate command-line tool configuration per hosted session without changing the environment of `tau serve` or unrelated sessions.

## what

Local execution environments can now carry optional environment overrides. Tau validates and persists those values, merges them into the sanitized environment used by tool processes, and preserves them when sessions are restored.


markusylisiurunen (commented):
No material findings. The local environment overrides are validated, scoped per session, sanitized after merging, and restored consistently from snapshots. The current CI failure is limited to formatter output.

PR #315, opened 2026-07-19
add session model sampling

### LINKED ISSUE #312: allow sampling the underlying model
when going through the sdk, it's often useful to easily be able to make "raw" LLM calls. basically like a raw LLM call to OpenAI Responses API or Chat Completions API. system prompt, reasoning effort, and so on. I would like to expose this capability in the session protocol to allow sampling the current persona's model.

markustherobot:
## why

SDK and protocol clients sometimes need a one-off call to the current persona's model without turning that call into Tau conversation state. The result also needs to retain the complete provider-neutral assistant message so callers can append it directly to a later sampling context.

## what

Adds a non-streaming `session.sample` request and `TauSdkSession.sample()` facade. The request accepts caller-owned system prompt, messages, optional tool schemas, and an explicit override allowlist containing only reasoning and max tokens. It inherits the active persona's model and remaining stream settings, returns the complete assistant message, and never executes returned tool calls.

Sampling runs with an isolated internal model session id and bypasses session history, deltas, snapshots, persistence, cost totals, and Tau usage logs. Calls can run concurrently with normal session work. `session.interrupt`, connection shutdown, and host shutdown abort active samples through transient handler-owned state.

Strict request and result validation covers the complete round-trippable message shape. The SDK exports dedicated sample input/result aliases, and the RPC, SDK, README, and repository guidance surfaces document the new method.

## details

The `options` object is required to keep one canonical request shape, while `reasoning` and `maxTokens` are individually optional and inherit persona defaults when absent. Temperature is intentionally not part of the contract. Sampling messages use Tau's existing provider-neutral message semantics, including timestamps, and optional tool schemas are supported from the first version because tool-call results are part of that same round-trippable assistant shape.

No unresolved implementation questions remain. Provider-side retention and billing remain governed by the selected provider even though Tau does not persist or account for the sample in session state.

fixes #312


src/host/local_session_host.ts
```diff
@@ -699,6 +701,15 @@ class LocalHostedSessionHandle implements LocalHostedSession {
     });
   }
 
+  async sample(
+    options: Omit<SessionProtocolSampleParams, "sessionId"> & {
+      signal?: AbortSignal;
+    },
+  ): Promise<SessionProtocolSampleResult> {
+    this.assertActive();
+    return { message: await this.session.sample(options) };
```
markustherobot: **[P2] Track samples in hosted-session shutdown**

`LocalHostedSessionHandle.sample()` delegates without registering the promise or owning an abort controller, while `waitForActiveWork()` only waits for turns and maintenance. A caller using `LocalSessionHost`/`TauHostedSession` directly can therefore start a sample and then call `session.dispose()` or `host.shutdown()`; disposal completes and tears down session resources while the provider request remains in flight. The protocol handler masks this only for protocol-originated calls. Track active samples on the hosted handle, abort them during disposal, and include them in `waitForActiveWork()` so the documented host-shutdown contract holds at the host boundary.

markustherobot (commented):
The protocol shape, state isolation, and round-trip validation look sound. The remaining issue is lifecycle ownership: sample work must participate in hosted-session shutdown independently of which adapter initiated it.

markusylisiurunen:
@markustherobot fix

src/host/local_session_host.ts
markustherobot: The hosted handle now owns an abort controller for each sample, combines it with the caller signal, tracks the promise in `waitForActiveWork()`, and aborts samples before direct disposal and host shutdown. Regression coverage verifies both paths settle before execution-environment teardown.

markusylisiurunen:
@markustherobot explain

markusylisiurunen:
@markustherobot deslop

markusylisiurunen:
@markustherobot deslop

markusylisiurunen:
@markustherobot deslop

PR #317, opened 2026-07-20
expose terminal turn outcomes

markustherobot:
## why

Session protocol clients could not distinguish a provider-error turn from a successful turn. Tau recorded the assistant message with `stopReason: "error"`, but `session.submit` only returned `aborted: false`, which allowed clients such as Cowork to silently mark failed work complete.

Clients also need the same outcome after reconnecting. Inferring it from assistant history duplicates Tau's turn semantics and is unreliable across tool subturns, compaction, and interruption.

## what

Replace the ambiguous turn flags with one discriminated terminal outcome covering completed, failed, aborted, and blocked turns. Failed outcomes expose the provider error message when available, while completed outcomes retain the model stop reason.

Persist each outcome on its submitted user message in session snapshots and export `getTauSdkSessionTurnOutcome()` for typed recovery by history entry ID. Existing snapshots without outcome metadata remain valid, and the SDK helper returns `undefined` when a turn has no persisted terminal outcome.

Update the TUI, Telegram session manager, protocol documentation, SDK examples, transport fixtures, and package type coverage for the new contract.

## details

The outcome is attached to the user history entry rather than a snapshot-wide latest-turn field so concurrent observers and reconnecting clients can correlate it with the idempotency ID they submitted. A turn that fails outside the normal model outcome path still produces a protocol error; if its response is lost, the absent persisted outcome lets a recovering client fail closed instead of assuming success.

This is the Tau half of markusylisiurunen/cowork#40. Cowork will consume the contract after the next Tau release.


markusylisiurunen:
@markustherobot review

markusylisiurunen:
@markustherobot explain update the explanation so that it makes it visible how this error is represented in the wire protocol as well as the snapshot with concrete examples

markustherobot (commented):
The outcome model is directionally sound, but two issues prevent it from being reliable across supported upgrade and history-mutation flows. The inline comments cover the required compatibility and stale-state fixes.

src/protocol/session_protocol.ts
```diff
@@ -272,14 +272,28 @@ export type SessionProtocolInitializeResult = {
   alreadyInitialized: boolean;
 };
 
-export type SessionProtocolTurnResult = {
-  turn: {
-    aborted: boolean;
-    blocked?: {
+export type SessionProtocolTurnOutcome =
```
markustherobot, line 275: **[P1] Bump the protocol version for the breaking result shape**

`SESSION_PROTOCOL_VERSION` remains 1 even though submit-family and retry responses now require this union instead of `{ aborted, blocked? }`. An existing v1 client connecting to an upgraded server passes ready/initialize and then rejects the submit result, while an upgraded client similarly rejects an older v1 server's result. Because remote attach and SDK peers can update independently, advertise a new protocol version (and update the versioned docs/fixtures) so this incompatibility fails during the handshake rather than after work has run.

src/host/local_session_host.ts
```diff
@@ -1308,14 +1314,16 @@ class LocalHostedSessionHandle implements LocalHostedSession {
         timestamp: 0,
       },
     };
-    const historyMessages = this.session.rawHistoryEntries.map(
-      (entry): SessionProtocolMessage => ({
+    const historyMessages = this.session.rawHistoryEntries.map((entry): SessionProtocolMessage => {
+      const turn = this.turnOutcomes.get(entry.id);
```
markustherobot, line 1335: **[P2] Remove outcomes when history entries are removed**

`turnOutcomes` is not pruned by `reconcileProjections()`, so rewind can remove a completed user entry while leaving its outcome in this map. A client may then reuse that explicit `historyEntryId` (the core allows it once the old entry is gone), and the newly recorded user message immediately inherits the old terminal result; if the replacement run fails before setting a new outcome, reconnect recovery falsely reports the stale result. Reconcile this map against the current message IDs whenever history is replaced.

markusylisiurunen:
@markustherobot fix

src/host/local_session_host.ts
markustherobot: `reconcileProjections()` now removes outcomes whose history entries no longer exist. The regression test rewinds a failed turn, reuses the same explicit history ID, and verifies the replacement message does not inherit the stale outcome.

markusylisiurunen:
@markustherobot deslop

PR #320, opened 2026-07-21
add telegram project selection and multi-repo workspaces

### LINKED ISSUE #319: move telegram towards a single bot helping with multiple repos instead of the current bot-per-repo approach
it's technically possible to work in multiple repos at the moment too but it's not optimal ux. I would basically want a more of a one bot and a set of repos it can work on, and the active repo can be selected with a slash command in telegram.

markustherobot:
## why

A Telegram bot can already be allowed to access several projects, but the conversation cannot choose among them. That leaves the runner effectively organized around one bot per repository and provides no supported workspace for work that spans repositories.

## what

Telegram now registers one `/use_<projectId>` command for every project available to the bot. The command persists that Telegram chat's preference for future `/new` sessions without changing its active Tau session. `/status` reports the active project, its composite members when applicable, and a different next-session preference.

Projects can now be repository projects or composites of at least two direct repository projects. Composite preparation reuses each member's existing cache and repository settings, places members under a generated non-Git root, runs member then composite bootstrap commands, and creates root `AGENTS.md` and `.tau/config.json` context files. The required composite persona remains authoritative for the root session, while child `.tau` configuration stays isolated and is resolved normally by subagents started in a child directory.

## details

Project ids are constrained to Telegram-safe command suffixes, nested composites and incompatible project fields are rejected, and each bot is checked against Telegram's command count limit. Project preferences are stored separately from recoverable session records at `<workspaceRoot>-project-preferences.json`.

The generated AGENTS file describes the repositories and their independent Git boundaries, plus configured composite instructions. It intentionally does not impose repository-by-repository reporting or verification rules. Synchronous workspace preparation now removes the complete workspace on failure for both composite and repository projects, keeping failed preparation all-or-nothing.

No unresolved implementation questions remain. Named composites are explicit config entries, child Tau configs are not implicitly merged, and `/use_*` affects only later `/new` calls as agreed in the issue discussion.

fixes #319


markusylisiurunen:
@markustherobot explain update the explanation to better demonstrate what this means: "The generated .tau/config.json explicitly includes member AGENTS.md files from each repository root down to its configured working directory" but keep it otherwise the same

markusylisiurunen:
@markustherobot explain amend the explanation to include a realistic example of the generated AGENTS.md file

markusylisiurunen:
@markustherobot fix so that the generated agents.md does not mention "Telegram", it's just as any other workspace for the agent. small change.

Issue #318, opened 2026-07-20
at least write tool still does not show queue or pending status while the tool arguments are being emitted by the model


markusylisiurunen:

markusylisiurunen:
@markustherobot is this a real issue?

markustherobot:
Yes. Confirmed on current `main`.

This is generic, not specific to `write`, but `write` makes it obvious because its `content` argument can take a while to stream. `runModelSubturn()` receives `toolcall_start`, but the accumulator only exposes a tool call after `toolcall_end`. `SessionEngine` then emits `tool_call_queued` only for calls present in that completed-call snapshot. As a result, there is no tool UI state at all while arguments are still arriving.

The coverage added in #298 does not catch this: its `toolcall_start` is immediately followed by `toolcall_end`, and it verifies that a fully emitted second call is visible while waiting behind the first call. It does not verify visibility during argument emission.

So this is a real, currently uncovered gap in the streamed tool-call UI, and the issue should remain open.

markusylisiurunen:
@markustherobot figure out how this should be implemented in a clean and robust (and generic) way

markustherobot:
### Summary

I recommend modeling this as a distinct, generic **tool-call streaming** phase rather than publishing `queued` early. The lifecycle should be:

`streaming arguments` → `queued for execution` → `running` → terminal state

`queued` should continue to mean that Tau has received the complete call and admitted it to the sequential runner. While the model is still producing arguments, the TUI should show a generic `preparing <tool>` card. This belongs in the shared session protocol path, so local TUI, remote attach, SDK observers, and snapshot hydration all see the same state. No `write`-specific behavior is needed.

I would not stream partial argument JSON into the session protocol. It can be large, temporarily invalid, and would create high-rate snapshot churn. The pending state only needs the provider tool-call id, tool name, originating assistant message id, and model content index.

### Plan

In `src/core/session/runner.ts`, inspect `toolcall_start` and `toolcall_delta` events through their `partial.content[contentIndex]`. Once that slot exposes a non-empty tool-call id and name, emit a new runner event such as `tool_call_streaming`. Track it by model content index and emit it only once unless the identity changes. Keep the current `toolcall_end` buffering and model-order admission logic unchanged, so no tool can execute before its full arguments exist and parallel/interleaved calls still enter the sequential runner in model order.

[...]

PR #321, opened 2026-07-21
show real targets for queued tool calls

### LINKED ISSUE #316: queued tool calls should show the real bash command
now it renders something like this:

```
 ⏵ running ssh -o BatchMode=yes astral-lily 'sudo -n du -x -h -d1 / 2>/dev/null | sort -h'

 ⏵ queued bash

 ⏵ queued bash
```

but I'd like it to show the real command. something similar probably applies to other tool calls as well.

markustherobot:
## why

Queued tool cards currently collapse every completed call to its tool name, so sequential bash calls render as `queued bash` even though Tau already has the full command. This also hides useful paths, URLs, objectives, and subagent targets for other tools.

## what

I added a required synchronous display-target method to tool definitions and use it when publishing the existing queued event. Built-in tools derive their target through their existing argument parsers, malformed calls use a stable invalid-arguments target, context-aware subagent calls resolve the current title when available, and client-provided tools explicitly fall back to their tool name.

I also changed the queued renderer to consume the event's `headerTarget` and added regression coverage for a second bash call queued behind active execution, a write path, malformed arguments, the client-tool fallback, and renderer target selection.

## details

I kept this change independent of #318 because that work has not landed yet. It only changes the completed-call queued state, so #318 can still introduce the earlier streaming state and transition the same tool card to this complete target after `toolcall_end`. No protocol shape or additional status is required here.

fixes #316

### DIFF
--- a/src/core/session/runner.ts
+++ b/src/core/session/runner.ts
@@ -288,14 +288,18 @@ function prepareToolCall(
   return { type: "ready", toolCall, definition };
 }
 
-function createToolQueuedEvent(toolCall: ToolCall): CoreToolUiEvent {
+function createToolQueuedEvent(
+  prepared: Extract<PreparedToolCall, { type: "ready" }>,
+  context: ToolDispatchContext,
+): CoreToolUiEvent {
+  const { toolCall, definition } = prepared;
   return {
     type: "tool_ui",
     uiEvent: {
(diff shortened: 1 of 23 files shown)


markusylisiurunen:
@markustherobot deslop

PR #322, opened 2026-07-21
show tool calls while arguments stream

### LINKED ISSUE #318: at least write tool still does not show queue or pending status while the tool arguments are being emitted by the model

markustherobot:
## why

Tool calls currently appear only after the model finishes emitting their arguments. Large calls such as `write` therefore leave no visible tool state while they are being prepared, even though the provider has already supplied a stable tool identity.

## what

I added a distinct streaming tool-call lifecycle across the model runner, core events, session protocol, host projections, and TUI. Once both the provider tool-call id and name are available, observers receive a `streaming` tool run and a generic `preparing <tool>` card. The complete call still enters the existing queue and executor only after `toolcall_end`.

The streaming protocol shape carries only the tool identity and draft assistant origin. Partial argument JSON is never copied into messages, tools, facets, or deltas. Completed runs retain the canonical `call` reference to a matching assistant `toolCall` block.

## details

Streaming identity is tracked by model content index. Identity changes replace the old tool and facet atomically, duplicate identities do not create conflicting runs, and retries, failures, and interruptions remove unfinished state. Calls rejected before dispatch now emit a generic blocked terminal event so a preparing card cannot remain stuck.

Although streaming patches are non-persisting, another snapshot write can capture the in-memory state. Recovery therefore explicitly removes streaming tools and their facets while preserving interrupted assistant text. Snapshot hydration and reconciliation also rebuild current preparing cards and remove discarded ones for local and remote clients.

I assume provider partial content is the canonical source for early id and name availability. Providers that do not expose both values remain on the existing completed-call path, and Tau does not synthesize an identity before completion. There are no remaining implementation questions.

fixes #318


src/core/session/runner.ts
```diff
@@ -115,10 +213,21 @@ export async function* runModelSubturn(
...
+        }
+
         if (event.type === "toolcall_end") {
           if (!accumulator) {
             continue;
           }
+          for (const streamingEvent of updateStreamingToolCall(event)) {
+            yield streamingEvent;
+          }
+          completeStreamingToolCall(event.contentIndex);
```
markustherobot: **[P1] Keep buffered completions tracked until admission**

When tool calls complete out of model order, this removes the later call from `streamingToolCalls` before the `while` loop can move it into the accumulator. If the earlier call is then interrupted or the stream fails, `discardStreamingToolCalls()` only emits a discard for the earlier call; the later run remains `streaming` while its draft message is promoted to `interrupted`. The host then tries to persist a snapshot whose `origin` no longer references a draft assistant message, causing turn cleanup to fail with `session snapshot is invalid` (and leaving the preparing card stuck). Keep completed-but-buffered identities tracked, and call `completeStreamingToolCall` only when that index is actually drained into the accumulator/admitted.

markustherobot (commented):
I found one material correctness issue in cleanup of streamed tool state. The lifecycle needs to retain cleanup ownership until a completed call is actually admitted.

markusylisiurunen:
@markustherobot fix

src/core/session/runner.ts
markustherobot: I now retain each completed identity until its content index is drained into the accumulator, so stream failure or interruption still discards completed-but-buffered calls. I also added a regression covering a later-index completion buffered behind an unfinished earlier call.

markusylisiurunen:
@markustherobot deslop

markusylisiurunen:
@markustherobot fix conflicts by rebasing

PR #326, opened 2026-07-22
allow explicit skill dependencies

### LINKED ISSUE #325: allow skills to activate explicit skill dependencies
## why

Explicit skills currently activate only when the user names them. This prevents skills from composing through deliberate skill references.

For example, a user may activate `@@skill:ui-work`, whose instructions require all browser operations to go through `@@skill:agent-browser`. If `agent-browser` uses `Trigger: explicit`, Tau's current guidance prevents the active `ui-work` skill from activating it because the user did not name `agent-browser` directly.

That forces users to know and name every transitive skill dependency, or forces dependency skills to use broader trigger sensitivity than their standalone behavior warrants.

## expected behavior

An exact `@@skill:<name>` reference in an already-active skill should count as explicit activation of that skill. `Trigger: explicit` should continue to prevent activation from generic keyword overlap or unrelated task context.

In the example above, naming `@@skill:ui-work` should allow `ui-work` to activate `@@skill:agent-browser`, while an ordinary browser-related request that names neither skill should not activate an explicit `agent-browser` skill.

Tau should document this compositional behavior in its trigger-sensitivity guidance and ensure repeated or cyclic references do not repeatedly activate the same skill.

markustherobot:
## why

Explicit capabilities previously activated only when the user named them directly. That prevented active project instructions and skills from delegating to narrowly triggered skill dependencies or sub-agents, and forced users to understand and repeat the full transitive dependency chain.

## what

The built-in persona and generated skill and sub-agent guidance now treat exact `@@skill:<name>` and `@@agent:<name>` references in the user request, active `AGENTS.md` instructions, or instructions of an already-active skill as explicit activation. Skill references compose transitively, while each skill activates at most once per request so repeated references and cycles terminate. Generic language, keyword, and task overlap still do not activate explicit capabilities.

The user and repository guidance document the same contract, and focused prompt assertions protect both skill and sub-agent activation sources, skill transitivity and cycle handling, and overlap exclusion.

## details

This implements the prompt-contract approach established in the issue discussion rather than introducing runtime dependency metadata. Skill dependencies remain informal and local to instruction text, with cycle termination and source discrimination enforced through model-facing guidance.

fixes #325

### DIFF
--- a/.tau/skills/guided-review/SKILL.md
+++ b/.tau/skills/guided-review/SKILL.md
@@ -12,7 +12,7 @@ Turn an overwhelming review scope into a two-phase guided review:
 
 Each review chunk should be small enough to understand in one sitting and large enough to preserve the relevant context.
 
-Use this skill only when the user explicitly asks for guided review, to split a diff into reviewable pieces, or to walk through changes step by step.
+Use this skill only after an exact `@@skill:guided-review` reference activates it under the trigger-sensitivity contract.
 
 ## Determine the review scope
 
--- a/.tau/skills/guided-review/SKILL.md
+++ b/.tau/skills/guided-review/SKILL.md
@@ -12,7 +12,7 @@ Turn an overwhelming review scope into a two-phase guided review:
 
 Each review chunk should be small enough to understand in one sitting and large enough to preserve the relevant context.
 
-Use this skill only after an exact `@@skill:guided-review` reference activates it under the trigger-sensitivity contract.
+Use this skill only when the user explicitly asks for guided review, to split a diff into reviewable pieces, or to walk through changes step by step.
 
 ## Determine the review scope
 
--- a/README.md
+++ b/README.md
@@ -388,7 +388,7 @@ sub-agents and skills define when they should be activated via trigger sensitivi
 
 - **eager**: use proactively whenever the capability would help, even if not explicitly requested. example: a dedicated codebase investigation sub-agent.
 - **balanced**: use when the request clearly matches the capability. this is the default if not specified. good for skills that solve specific problems but shouldn't be assumed.
-- **explicit**: use only when the user specifically names or requests it. example: a web research sub-agent that should only run when asked.
+- **explicit**: use only when explicitly named. for skills and sub-agents, an exact `@@skill:<name>` reference or `@@agent:<name>` reference in the user request, active `AGENTS.md` instructions, or instructions of an already-active skill counts as explicit activation. skill references compose transitively, and each skill activates at most once per request so repeated or cyclic references terminate. generic language, keyword, or task overlap does not count.
 
 when you write custom skills, you can specify trigger sensitivity in the skill description. if not specified, the default is balanced. the model respects these levels and won't trigger a skill or sub-agent inappropriately.
 
@@ -410,7 +410,7 @@ reference files in your message with `@<path>` (for example, `@src/tui/session_c
 
 reference skills with `@@skill:<name>` (for example, `@@skill:skill-name`). autocomplete will suggest available skills. exact skill references in active `AGENTS.md` instructions and already-active skill instructions also activate the referenced skill, including skills with `Trigger: explicit`. these references compose transitively, and each skill activates at most once per request so repeated or cyclic references terminate.
 
-to explicitly target a sub-agent, use `@@agent:<name>` (for example, `@@agent:default`).
+reference sub-agents with `@@agent:<name>` (for example, `@@agent:default`). exact sub-agent references in active `AGENTS.md` instructions and already-active skill instructions also activate the referenced sub-agent, including sub-agents with `Trigger: explicit`.
 
 you can also pipe content directly:
 
--- a/README.md
+++ b/README.md
@@ -408,7 +408,7 @@ toggle visibility of the model's thinking with `ctrl+t`.
 
 reference files in your message with `@<path>` (for example, `@src/tui/session_chat_app.ts`). autocomplete helps you find the right path.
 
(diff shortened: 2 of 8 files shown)


markustherobot (commented):
The prompt-contract change is coherent and the full test suite passes, but the rollout leaves contradictory user-facing and bundled skill instructions that still enforce the old activation rule.

README.md
```diff
@@ -408,7 +408,7 @@ toggle visibility of the model's thinking with `ctrl+t`.
 
 reference files in your message with `@<path>` (for example, `@src/tui/session_chat_app.ts`). autocomplete helps you find the right path.
 
-reference skills with `@@skill:<name>` (for example, `@@skill:skill-name`). autocomplete will suggest available skills.
+reference skills with `@@skill:<name>` (for example, `@@skill:skill-name`). autocomplete will suggest available skills. exact skill references in active `AGENTS.md` instructions and already-active skill instructions also activate the referenced skill, including skills with `Trigger: explicit`. these references compose transitively, and each skill activates at most once per request so repeated or cyclic references terminate.
```
markustherobot, line 411: **[P2] Keep the trigger-sensitivity section consistent**

The main trigger-sensitivity definition at `README.md:391` still says that `explicit` means only when the user names or requests the capability. That directly contradicts this new contract for `AGENTS.md` and active-skill references, so readers using the dedicated section will conclude that dependency activation is disallowed. Update that definition to distinguish skill activation sources from the intentionally user-only sub-agent rule.

src/core/utils/context_builder.ts
```diff
@@ -43,7 +43,7 @@ export function buildSkillsIndexBlock(skills: Skill[]): string | undefined {
   lines.push(
     "",
     "Guidelines:",
-    "- Trigger: Follow the skill's trigger sensitivity if specified; default is balanced. Always activate if user names it with `@@skill:<name>`. For explicit triggers, only activate when the user explicitly names the skill (for example with `@@skill:<name>`), not based on keyword overlap. Use the minimal set of skills that covers the request.",
+    "- Trigger: Follow the skill's trigger sensitivity if specified; default is balanced. An exact `@@skill:<name>` reference in the user request, active AGENTS.md instructions, or instructions of an already-active skill explicitly activates that skill. Skill references compose transitively. Activate each skill at most once per request; repeated or cyclic references do not reopen it. Do not infer skill activation from generic language, keyword, or task overlap. Use the minimal set of skills that covers the request.",
```
markustherobot, line 46: **[P2] Update the bundled skill that restores the old guard**

`starter/skills/guided-review/SKILL.md:15` (and its identical `.tau/skills/guided-review/SKILL.md` copy) still instructs the model to use the skill only when the user explicitly asks. If an active `AGENTS.md` or skill references `@@skill:guided-review`, opening this shipped explicit skill therefore reimposes the old user-only rule and conflicts with this guidance. Rewrite that guard around the new exact-reference activation contract and keep both copies synchronized.

src/core/personas.ts
```diff
@@ -154,7 +154,7 @@ Skills and sub-agents specify when they should be activated:
 
 - **eager**: Use proactively whenever the capability would help, even if not explicitly requested.
 - **balanced**: Use when the request clearly matches. This is the default when not specified.
-- **explicit**: Use only when the user explicitly names the skill or sub-agent (use \`@@skill:<name>\` or \`@@agent:<name>\`). Do not infer from generic language or keyword overlap.
+- **explicit**: Use only when explicitly named. For skills, an exact \`@@skill:<name>\` reference in the user request, active AGENTS.md instructions, or instructions of an already-active skill counts as explicit activation, and skill references compose transitively. Activate each skill at most once per request; repeated or cyclic references do not reopen it. For sub-agents, the user must explicitly name \`@@agent:<name>\`. Do not infer from generic language, keyword, or task overlap.
```
markusylisiurunen: subagents should effectively work the same so e.g. an explicit mention in AGENTS.md should be enough imo

markusylisiurunen:
@markustherobot fix

README.md
markustherobot: I updated the dedicated trigger-sensitivity section and the mention guidance to describe the exact-tag sources for both skills and sub-agents, while preserving the generic-overlap exclusion.

src/core/utils/context_builder.ts
markustherobot: I replaced the user-only guard with an exact `@@skill:guided-review` activation guard and kept the starter and project copies identical.

src/core/personas.ts
markustherobot: I extended both the built-in trigger guidance and the generated sub-agent block so an exact `@@agent:<name>` reference in the user request, active `AGENTS.md`, or an already-active skill activates an explicit sub-agent. The docs and focused prompt assertions now cover the same contract.

.tau/skills/guided-review/SKILL.md
```diff
@@ -12,7 +12,7 @@ Turn an overwhelming review scope into a two-phase guided review:
 
 Each review chunk should be small enough to understand in one sitting and large enough to preserve the relevant context.
 
-Use this skill only when the user explicitly asks for guided review, to split a diff into reviewable pieces, or to walk through changes step by step.
+Use this skill only after an exact `@@skill:guided-review` reference activates it under the trigger-sensitivity contract.
```
markusylisiurunen: I don't like this change, prev version was better imo

markusylisiurunen:
@markustherobot fix

.tau/skills/guided-review/SKILL.md
markustherobot: I restored the previous guided-review sentence in both the project and starter copies, keeping them identical. The central activation-source guidance and its tests remain unchanged.

PR #328, opened 2026-07-23
allow concurrent session side channels

### LINKED ISSUE #327: model-launched diff tools do not work at all
see screenshot below:

Image

markustherobot:
## why

Model-launched diff review pauses its parent turn while the TUI captures a snapshot with `session.exec` and drives an ephemeral review context. The protocol treated exec and ephemeral lifecycle requests as main-session work, so exec returned `busy` and ephemeral create/close could interrupt or wait on the parent turn they needed to complete.

## what

Exec is now an execution-environment side channel that can overlap turns, mutations, samples, ephemeral work, and other exec requests without entering main-session scheduling. Hosted sessions use one active-work lifecycle for turns, execs, samples, and maintenance so explicit interrupt, connection closure, session disposal, and host shutdown cancel and settle work consistently.

Ephemeral context create and close now bypass the main-session mutation queue, matching ephemeral submit and keeping their independent, non-persisted lifecycle separate from recoverable session state.

Focused host and protocol regressions cover concurrent exec scheduling, cancellation when a backend resolves after abort, disconnect during delayed session recovery, and ephemeral lifecycle during an active turn. RPC and SDK concurrency documentation reflects the canonical contract.

## details

Workspace races between concurrent execs are intentional and remain the client's responsibility. `session.interrupt` cancels all active work for the hosted session, while closing an ephemeral context continues to own cancellation of that context's live threads.

fixes #327

### DIFF
--- a/src/host/local_session_host.ts
+++ b/src/host/local_session_host.ts
@@ -456,10 +456,7 @@ export class LocalSessionHost implements TauSessionHost {
     }
 
     for (const session of this.sessions) {
-      session.interruptTurn();
-      session.interruptExecs();
-      session.interruptMaintenance();
-      session.interruptSamples();
+      session.interruptActiveWork();
     }
 
     const settlementResults = await Promise.allSettled(
@@ -542,12 +539,8 @@ class LocalHostedSessionHandle implements LocalHostedSession {
   private runtimeEventQueue: Promise<void> = Promise.resolve();
   private snapshotQueue: Promise<unknown> = Promise.resolve();
   private snapshotGeneration = 0;
-  private readonly execAbortControllers = new Set<AbortController>();
-  private readonly execPromises = new Set<Promise<unknown>>();
-  private readonly maintenanceAbortControllers = new Set<AbortController>();
-  private readonly maintenancePromises = new Set<Promise<unknown>>();
-  private readonly sampleAbortControllers = new Set<AbortController>();
-  private readonly samplePromises = new Set<Promise<unknown>>();
+  private readonly activeWorkAbortControllers = new Set<AbortController>();
+  private readonly activeWorkPromises = new Set<Promise<unknown>>();
   private activeTurnPromise?: Promise<SessionProtocolTurnOutcome>;
   private disposePromise?: Promise<void>;
   private disposing = false;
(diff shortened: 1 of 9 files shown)


src/host/local_session_host.ts
```diff
@@ -725,12 +740,24 @@ class LocalHostedSessionHandle implements LocalHostedSession {
...
-    return await backend.runBash(options.command, {
+    const promise = backend.runBash(options.command, {
       ...(options.cwd !== undefined ? { cwd: options.cwd } : {}),
       ...(options.timeoutMs !== undefined ? { timeoutMs: options.timeoutMs } : {}),
-      ...(options.signal !== undefined ? { signal: options.signal } : {}),
+      signal,
     });
+    this.execPromises.add(promise);
+    try {
+      return await promise;
```
markustherobot: **[P1] Return `cancelled` when a hosted exec is aborted**

`runBash` does not consistently reject on abort: the local and Fly backends resolve with a normal result containing `(tau) aborted`. When `session.interrupt` aborts this handle's private controller, the handler's signal also remains un-aborted, so this return reaches `executeExec` and emits `ok: true` instead of the documented `cancelled` error. Check the combined `signal` after `await promise` (for example, `signal.throwIfAborted()`) before returning, or otherwise make hosted exec cancellation reject canonically.

src/host/session_protocol_handler.ts
```diff
@@ -1293,15 +1262,19 @@ export class SessionProtocolHandler {
...
-      this.sendMessage(
-        createSessionProtocolSuccessResponse(request.id, "session.ephemeral.create", result),
-      );
+    const state = await this.getSessionState(request.params.sessionId);
+    if (!state) {
+      this.sendSessionNotFound(request.id, request.params.sessionId);
+      return;
+    }
+
+    const result = await state.session.createEphemeralContext({
```
markustherobot, line 1232: **[P2] Avoid creating contexts after the connection closes**

`withSessionMutation` used to re-check `this.closed` after awaiting session lookup. Now, if a WebSocket disconnects while `getSessionState` is recovering an unobserved session, `close()` completes with no state to clean up; when recovery finishes, this creates a context, the response is dropped by `sendMessage`, and the context remains inaccessible until session disposal. Re-check `this.closed` after lookup (as exec and sample do) before creating the context.

markustherobot (commented):
Two correctness issues remain in cancellation reporting and disconnect cleanup. The side-channel scheduling model is otherwise coherent, and the focused host and RPC regression suites pass.

markusylisiurunen:
@markustherobot fix

src/host/local_session_host.ts
markustherobot: I now check the hosted exec’s combined signal after `runBash` settles, so backends that resolve normally on abort still reject and the protocol returns `cancelled`. Host and RPC regressions cover the resolve-on-abort path.

src/host/session_protocol_handler.ts
markustherobot: I added the post-lookup closed check before ephemeral context creation. A regression now closes the handler while session recovery is blocked and verifies that no inaccessible context is created afterward.

markusylisiurunen:
@markustherobot deslop

deslop in general but also feels like this change should have simplified the code instead of adding more/making it more complex? also, is there anything to trim in tests now that things should be simpler.

markusylisiurunen:
@markustherobot review

PR #333, opened 2026-07-25
run commands in login Bash

### LINKED ISSUE #332: is the bash tool using sh under the hood?

markustherobot:
## why

Command execution was not consistently using the session execution environment's login-configured Bash toolchain. Local, attached, and hosted paths could resolve different Node, Git, and rg versions, while machine-facing helpers had begun expanding the public protocol with separate process and file methods.

The TUI and remote attach need one execution contract regardless of physical co-location, and every agent-visible helper executable should resolve after the same login initialization as model Bash commands.

## what

This standardizes model `bash` calls, direct `!`/`!!` commands, SDK helpers, and `session.exec` on fresh non-interactive `bash -lc` processes inside the session execution environment. Node, Git, rg, diff capture, and file helpers use exact positional arguments through `exec "$0" "$@"`, preserving argv boundaries while resolving executables from the login-configured `PATH`.

The public protocol remains centered on `session.exec`. It now carries a required execution id plus optional positional arguments, environment overrides, bounded binary stdin, cwd, timeout, and capture limit. `session.cancelExec` retains targeted cancellation. The separate `session.execProcess`, `session.readFile`, and `session.writeFile` methods have been removed, and local TUI and remote attach use the same SDK/session-protocol path.

## details

Execution keeps bounded capture and explicit timeout, abort, signal, and truncation metadata. Binary file helpers transfer content through exec stdin/stdout without shell quoting. Cloudflare privately stages stdin because its bridge does not expose process stdin, and Fly carries the same command, arguments, environment, and stdin contract through its worker while preserving process-group cleanup.

Tau sets `HOME` from the execution environment snapshot and relies on normal Bash login initialization. Login startup files are therefore required to be automation-safe and silent: they must not write to stdout or stderr, read stdin, require a TTY, or terminate the shell unexpectedly. Tau intentionally adds no output framing or filtering workaround.

fixes #332


src/execution/cloudflare_sandbox_execution_environment.ts
```diff
@@ -427,6 +414,7 @@ export class CloudflareSandboxBridgeClient {
         body: JSON.stringify({
           argv: options.argv,
           ...(options.cwd ? { cwd: options.cwd } : {}),
+          ...(options.env ? { env: options.env } : {}),
```
markustherobot: **[P1] Apply HOME through the supported bridge contract**

The current Cloudflare Sandbox bridge's `ExecRequest` and `/exec` route only accept and apply `argv`, `timeout_ms`, and `cwd`; this extra `env` field is ignored. Consequently the scoped snapshot `HOME` never reaches `bash -lc`, so a configured Cloudflare home and its login profile are not used, defeating the main contract for this backend. Pass `HOME` through supported argv (for example via `env`) or extend the deployed bridge contract and test that integration rather than only asserting the request body.

src/core/tools/execution_backend.ts
```diff
@@ -124,8 +124,7 @@ export function createLocalToolExecutionBackend(
           ? timeoutMs
           : undefined;
 
-      const result = await spawnCapture(command, [], {
-        shell: true,
+      const result = await spawnCapture("bash", ["-lc", command], {
```
markustherobot: **[P2] Isolate diff snapshot commands from login output**

This makes profile output part of every `session.exec`, but `/diff` still runs its `git` commands and Node file reader through `createSessionDiffReviewSnapshotDeps` and parses the entire returned output as machine data. If `.bash_profile` prints anything, `git rev-parse --show-toplevel` is prefixed, the derived repository path becomes invalid, and diff review fails before opening. Route these snapshot operations through a framed/direct helper, or add boundaries around their stdout and stderr as done for the SDK Node helpers.

markustherobot (commented):
The login-Bash direction is sound, but two integration paths still violate the new contract. The current tests pass without exercising the real Cloudflare bridge or the diff snapshot path under noisy login startup.

markusylisiurunen:
@markustherobot fix and also resolve conflicts by rebasing on main

src/execution/cloudflare_sandbox_execution_environment.ts
markustherobot: Resolved by applying the scoped environment through the bridge-supported argv contract (`env HOME=… bash -lc …`) and removing the unsupported request-body `env` field. The regression now asserts the exact supported request shape.

src/core/tools/execution_backend.ts
markustherobot: Resolved by routing diff snapshot Git commands and file reads through the SDK backend’s framed output path. A real-repository regression now injects login stdout and stderr around every `session.exec` and verifies tracked and untracked diff capture remains parseable.

markusylisiurunen:
@markustherobot deslop

markusylisiurunen:
Pushed `da4663c` with the actionable shell-execution audit fixes.

- Fly now escalates command-group cancellation from TERM to KILL, drains active commands before worker shutdown, avoids the disposal race, and decodes split UTF-8 safely.
- Cloudflare command deadlines now include queue/session setup and cleanup is bounded; file operations are cancelable, and binary size limits are enforced while streaming rather than after full download.
- Noninteractive Git safeguards now apply consistently to local, Cloudflare, and Fly Bash.
- SDK diff execution now propagates aborts and a bounded per-command capture limit through `session.exec`, preserves framing under truncation, and parses successful Git output from stdout without mixing warnings into patches.
- Hosted binary reads reject oversized files before encoding, and direct `!` history records nonzero exit status.

I left environment filtering policy, Cloudflare serialization, sequential Telegram bootstrap, output-capture limits, and direct-binary provisioning unchanged because those are intentional contracts rather than fixes for this PR. Formatting, type checking, build, and all 760 tests pass.

markusylisiurunen:
@markustherobot review

src/sdk/session.ts
```diff
@@ -572,9 +575,24 @@ class TauSdkSessionImpl implements TauSdkSession {
...
+      timeoutMs?: number;
+      maxCaptureBytes?: number;
+      signal?: AbortSignal;
+    } = {},
   ): Promise<SessionProtocolResultByMethod["session.exec"]> {
-    return await this.client.sendExec(this.activeSessionId(), command, options);
+    const { signal, ...execOptions } = options;
+    signal?.throwIfAborted();
+    const onAbort = () => {
+      void this.interrupt();
```
markustherobot: **[P1] Consume failures from the abort-triggered interrupt**

If the transport closes or otherwise rejects `session.interrupt` while an exec signal is being aborted, this discarded promise becomes an unhandled rejection. Under Node's default unhandled-rejection behavior that can terminate the SDK consumer, even if the caller handles the original `exec` failure. Attach a rejection handler to this fire-and-forget interrupt (or await it as part of the cancellation flow).

src/execution/cloudflare_sandbox_execution_environment.ts
```diff
@@ -441,41 +458,80 @@ export class CloudflareSandboxBridgeClient {
...
+      { method: "GET", signal },
     );
     if (!response.ok) {
       throw await this.createHttpError(response);
     }
-    return Buffer.from(await response.arrayBuffer());
+
+    const contentLength = Number(response.headers.get("Content-Length"));
+    if (Number.isFinite(contentLength)) {
+      assertFileWithinMaxBytes(contentLength, maxBytes);
```
markustherobot: **[P2] Cancel oversized responses rejected from Content-Length**

When `Content-Length` already exceeds `maxBytes`, this assertion throws before the response body is acquired or canceled. The fetch can therefore keep downloading the oversized artifact and hold its connection despite the early size check, defeating the resource protection this path is meant to provide. Cancel `response.body` before propagating the size error, as the streaming overflow branch already does.

markustherobot (commented):
The Bash execution contract is coherent overall, but two asynchronous failure paths still need hardening so cancellation and size limits remain safe at their boundaries.

markusylisiurunen:
@markustherobot fix

src/sdk/session.ts
markustherobot: Resolved by attaching a rejection handler to the abort-triggered `session.interrupt`, keeping the best-effort cancellation detached without allowing transport failures to become unhandled rejections. The SDK regression aborts an exec while interrupt rejects and verifies the exec caller remains safe.

src/execution/cloudflare_sandbox_execution_environment.ts
markustherobot: Resolved by cancelling the response body before rethrowing a `Content-Length` size violation. The regression uses an oversized declared length and asserts the stream cancellation hook runs on the early-rejection path.

markusylisiurunen:
The login-Bash direction is right, but the remaining issues share one architectural cause: the TUI/SDK is trying to reconstruct machine-facing argv execution, environment propagation, cancellation, and file access on top of the user-facing login-Bash `session.exec` contract using shell quoting and random output markers. That adapter cannot faithfully implement `ToolExecutionBackend`: it loses environment/unlimited-capture semantics and stream ordering, maps a per-call abort to session-wide interruption, and silently turns truncated file reads into empty content.

I’m going to fix this design-first by giving the session protocol a canonical machine-facing process execution path with explicit argv/environment and targeted exec cancellation, while keeping `session.exec` as login Bash. Then I’ll remove the framing workaround and address the resulting low-level findings: bounded diff file reads, Cloudflare timeout/result symmetry and awaited cleanup, Fly lifecycle/file-limit races, macOS path-portable tests, and the remaining startup/docs inconsistencies. The intended end state is one explicit Bash contract plus one explicit process contract, implemented symmetrically across local, Cloudflare, Fly, SDK, and TUI paths.

markusylisiurunen:
Commit `40921c9` replaces the shell-framing workaround with two explicit execution contracts. `session.exec` remains the user-facing path and starts a fresh non-interactive login Bash, while the new `session.execProcess` runs Tau-controlled argv directly with structured environment overrides. The TUI, SDK, Git helpers, Node helpers, and diff snapshot code now use the appropriate contract instead of reconstructing process execution with shell quoting and random output markers.

The protocol is bumped to v3 and now gives every exec a required `execId`. `session.cancelExec` cancels only that operation, including when cancellation comes from another attached client, while duplicate active ids return `busy`; `session.interrupt` remains session-wide. Execution results are symmetric across local, Cloudflare, and Fly and include combined and split output, truncation, timeout, abort, and close-signal metadata, with a shared 16 MiB maximum capture bound.

Machine-facing file access now uses bounded `session.readFile` and `session.writeFile` operations rather than encoding file contents into command arguments. This fixes large diff patch reads and preserves binary payloads. Internal config and prompt-context collectors also fail explicitly if their bounded output is truncated instead of parsing partial JSON.

The backend work completes the symmetry: local Bash is a thin `bash -lc` specialization of direct process execution; Cloudflare normalizes timeout/abort results and awaits command-session cleanup during disposal; Fly propagates termination metadata, rechecks binary size after reading, has repeatable disposal, and bounds helper shutdown. The SDK-backed TUI preserves argv, environment, stream ordering, capture metadata, and targeted abort behavior without shell framing.

Tests and documentation were updated across protocol, SDK, RPC, local, Cloudflare, Fly, TUI, and diff-review paths. Verification passed with `npm run check`, 767 tests across 70 files, `git diff --check`, and a real in-process SDK exercise covering direct argv execution, binary transfer, and isolated concurrent cancellation.

PR #334, opened 2026-07-25
recover Telegram sessions after failed turns

### LINKED ISSUE #329: telegram should not terminally fail a session if provider returns an error

markustherobot:
## why

A provider-returned failed turn currently moves the Telegram session into its terminal `failed` state, even though the underlying Tau session remains usable. That drops the persisted chat mapping and forces the user to start a new session after a transient provider or compaction failure.

## what

The Telegram session manager now emits a dedicated failed-turn event for resolved `failed` and `blocked` outcomes and lets the tracked submit lifecycle return the session to `waiting-input`. Failed steering batches emit one event per completed turn, keyed by their shared user history entry.

The adapter reports the concrete provider or blocking error, keeps typing active while aggregated steering work remains, and preserves chat routing to the same session for the next message.

Regression coverage verifies failed and blocked outcomes, batched steering deduplication, typing lifecycle, client retention, follow-up submission, persistence and recovery, user notification, and same-session routing.

## details

Rejected `submit()` or `steer()` operations still have terminal semantics and close the SDK client because they represent a session or transport boundary failure. Resolved turn outcomes reuse the canonical session protocol failure shape rather than introducing a duplicate Telegram error contract. No documentation changes are needed, and there are no remaining implementation questions.

fixes #329

### DIFF
--- a/src/core/telegram/adapter.ts
+++ b/src/core/telegram/adapter.ts
@@ -2609,7 +2609,6 @@ class TelegramAdapterImpl {
         return;
       }
 
-      this.stopTypingIndicators(event.sessionId);
       this.notifySession(
         event.sessionId,
         event.failure.status === "failed"
--- a/src/core/telegram/adapter.ts
+++ b/src/core/telegram/adapter.ts
@@ -2604,6 +2604,21 @@ class TelegramAdapterImpl {
       return;
     }
 
+    if (event.type === "session-turn-failed") {
+      if (!this.chatsBySession.has(event.sessionId)) {
+        return;
+      }
+
+      this.stopTypingIndicators(event.sessionId);
+      this.notifySession(
+        event.sessionId,
+        event.failure.status === "failed"
+          ? `turn failed: ${event.failure.errorMessage ?? "model provider returned an error"}`
+          : `turn blocked: ${event.failure.message}`,
+      );
+      return;
+    }
+
     if (event.type !== "session-state-changed") {
       return;
     }
--- a/src/core/telegram/session_manager.ts
+++ b/src/core/telegram/session_manager.ts
@@ -184,6 +184,13 @@ export type TelegramSessionManagerEvent =
       log: TelegramSessionLogEntry;
     }
   | {
+      type: "session-turn-failed";
+      sessionId: string;
+      projectId: string;
+      timestamp: string;
+      failure: Extract<SessionProtocolSubmitResult["turn"], { status: "failed" | "blocked" }>;
+    }
+  | {
       type: "session-progress";
       sessionId: string;
       projectId: string;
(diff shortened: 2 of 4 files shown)


src/core/telegram/session_manager.ts
```diff
@@ -909,11 +916,13 @@ class TelegramSessionManagerImpl implements TelegramSessionManager {
 
         if (!entry.cancelRequested) {
           if (failure) {
-            entry.record.error =
-              failure.status === "failed"
-                ? (failure.errorMessage ?? "model provider returned an error")
-                : failure.message;
-            this.setState(entry, "failed");
+            this.emit({
```
markustherobot, line 923: **[P2] Deduplicate failures for batched steering turns**

`session.steer` batches pending steering requests into one turn and returns the same `userHistoryEntryId` and turn result to every request. Each corresponding `submitText` reaches this unconditional emit, so one failed or blocked turn produces one Telegram error notification per batched message. Emit once per completed turn, for example by deduplicating on `userHistoryEntryId`, rather than once per protocol response.

src/core/telegram/adapter.ts
```diff
@@ -2604,6 +2604,21 @@ class TelegramAdapterImpl {
       return;
     }
 
+    if (event.type === "session-turn-failed") {
+      if (!this.chatsBySession.has(event.sessionId)) {
+        return;
+      }
+
+      this.stopTypingIndicators(event.sessionId);
```
markustherobot: **[P2] Keep typing active while later steering work remains**

A failed submit can resolve while `entry.activeSubmit` still aggregates pending steering submissions, and those submissions can start another turn without a new `running` state transition because the manager never left `running`. Stopping all indicators here therefore leaves the chat without typing feedback for the rest of that active work. Let the eventual transition out of `running` stop typing, or only stop here once the aggregate submit has settled.

markustherobot (commented):
Single-turn recovery works as intended, but concurrent Telegram steering still has failure-notification and typing-indicator regressions that should be addressed.

markusylisiurunen:
@markustherobot fix

src/core/telegram/session_manager.ts
markustherobot: I now deduplicate resolved turn failures by the shared `userHistoryEntryId`, so every batched steering turn emits exactly one failure event. A regression test covers two steering responses for the same failed turn.

src/core/telegram/adapter.ts
markustherobot: I removed typing teardown from the intermediate failed-turn event. Typing now remains active for aggregated steering work and stops on the eventual transition out of `running`, with timer-based regression coverage.

Issue #58, opened 2026-02-02
feels like the @file: options are not always updated


markusylisiurunen:
should refresh the list of options basically whenever it's used.

markusylisiurunen:
Closing as not actionable: @file suggestions refresh asynchronously on entry and update on subsequent input while autocomplete is open, which matches expected behavior.

PR #335, opened 2026-07-25
provision Telegram workspaces through Tau sessions

### LINKED ISSUE #330: align cowork and telegram workspace provision scripts

markustherobot:
## why

Telegram repository setup had two configuration-driven command paths with different readiness, failure, environment, and recovery behavior. That prevented repositories from exposing one setup entry point shared with Cowork and bypassed the shell environment available through normal Tau commands.

## what

I replaced `bootstrapCommands` and `backgroundBootstrapCommands` with one optional executable `.tau/scripts/provision` repository hook. Workspace preparation now returns explicit repository provision targets, and the connected Tau SDK session runs each hook through `session.exec` after the session becomes available.

Provision failures emit a dedicated session-manager event, retain the usable session and workspace, and send a bounded diagnostic to every linked Telegram chat. New and reconstructed workspaces run provisioning, while preserved workspaces continue to skip it on runner restart.

## details

Composite sessions run each member repository hook from that member's configured working directory and do not invent a hook for the generated root. Member failures do not prevent the remaining member hooks from running.

The hook must be a regular executable file, symlinks are rejected, and its shebang selects the script language. Provisioning is treated as a finite task, runs without an additional Telegram timeout, and is cancelled when the session closes.

This uses the existing `session.exec` contract and is independently mergeable. PR #333 supplies the agreed target-side login Bash behavior so provisioning receives the same initialized toolchain environment as Tau's normal command paths. The coordinated Cowork implementation is markusylisiurunen/cowork#58. There are no remaining implementation questions.

fixes #330


src/core/telegram/session_manager.ts
```diff
@@ -814,57 +847,80 @@ class TelegramSessionManagerImpl implements TelegramSessionManager {
...
+        this.log(entry, "info", "workspace provision started", {
+          targetProjectId: target.projectId,
+          cwd: target.cwd,
         });
-      })
-      .finally(() => {
-        if (entry.backgroundBootstrapPromise === backgroundBootstrapPromise) {
-          entry.backgroundBootstrapPromise = undefined;
+        try {
+          const result = await tauSession.exec(buildProvisionCommand(target.repositoryRoot), {
```
markustherobot: **[P1] Keep turn interrupts from cancelling provisioning**

When a user submits a message while a slow provision hook is still running and then invokes `/interrupt`, `interruptSession()` calls `tauSession.interrupt()`. The protocol handler implements that with `interruptActiveWork()`, which aborts every active-work controller, including this concurrent `session.exec`; because `cancelRequested` is still false, this is reported as a provision failure and the hook is never retried. An ordinary assistant-turn cancellation can therefore permanently stop workspace setup. Use a turn-only interruption path here, or otherwise isolate/retry provisioning so only session closure cancels it.

src/core/telegram/session_manager.ts
```diff
@@ -681,7 +719,7 @@ class TelegramSessionManagerImpl implements TelegramSessionManager {
     this.setState(entry, "waiting-input");
     this.log(entry, "info", "session recovered", { tauSessionId, workspacePath });
     if (shouldPrepareWorkspace) {
-      this.startBackgroundBootstrap(entry, sessionCwd, memberBackgroundBootstrapCommands);
+      this.startProvision(entry, tauSession, provisionTargets);
```
markustherobot, line 750: **[P2] Do not emit recovery failures before adapters subscribe**

For a reconstructed workspace, this starts provisioning during `sessionManager.initialize()`, while `startTelegramRuntime()` creates and subscribes the Telegram adapters only after that initialization returns. A fast invalid or non-zero hook can emit `session-provision-failed` before any adapter has restored the chat mapping, and the event is neither persisted nor replayed, so the linked chat receives no diagnostic. Delay recovery provisioning until adapters are ready, or retain the failure state for adapters to deliver after startup.

src/core/telegram/session_manager.ts
```diff
@@ -115,6 +117,34 @@ function elapsedMs(startTime: bigint): number {
...
+  return `'${value.replace(/'/g, `'\\''`)}'`;
+}
+
+function buildProvisionCommand(repositoryRoot: string): string {
+  return [
+    `provision_path=${shellQuote(resolve(repositoryRoot, PROVISION_SCRIPT_PATH))}`,
+    'if [ ! -e "$provision_path" ] && [ ! -L "$provision_path" ]; then exit 0; fi',
+    'if [ -L "$provision_path" ] || [ ! -f "$provision_path" ]; then echo ".tau/scripts/provision must be a regular file" >&2; exit 1; fi',
+    'if [ ! -x "$provision_path" ]; then echo ".tau/scripts/provision must be executable" >&2; exit 1; fi',
+    'exec "$provision_path"',
```
markustherobot, line 133: **[P2] Enforce the documented shebang requirement**

The preceding checks only require a regular executable file. If an executable text file has no shebang, `execve` returns `ENOEXEC` and the invoking shell interprets it itself, so this command can succeed under Bash/sh despite the documented requirement that the hook have a shebang (and an executable native binary is accepted too). Validate the `#!` prefix inside this session command before executing the hook so malformed hooks fail with the promised diagnostic instead of running under an implicit interpreter.

markustherobot (commented):
The repository-hook direction fits the intended session boundary, but provisioning lifecycle and validation still have material gaps. In particular, normal session operations can cancel or lose provision outcomes, so the user-visible nonterminal failure contract is not yet reliable.

markusylisiurunen:
@markustherobot fix

src/core/telegram/session_manager.ts
markustherobot: I now retry a cancelled provisioning exec for the same target unless session closure has set `cancelRequested`. The regression test interrupts an active turn while provisioning is running and verifies the hook restarts without emitting a failure.

src/core/telegram/session_manager.ts
markustherobot: I now retain provision failures in the session manager, then replay them after an adapter restores its chat mappings and subscribes. The recovery-path test verifies a failure produced before adapter startup still reaches the restored chat.

src/core/telegram/session_manager.ts
markustherobot: I added a preflight that reads the first two bytes and requires `#!` before direct execution. Executable text without a shebang and native binaries now fail with the documented diagnostic instead of falling through to an implicit shell.

PR #341, opened 2026-07-29
fix context compaction recovery

### LINKED ISSUE #338: auto-compact does not always work
Telegram session ended in:

```
turn failed: Codex error: Your input exceeds the context window of this model. Please adjust your input and try again.
```

markustherobot:
## why

Long tool-driven turns could auto-compact only once, then grow past the model context window later in the same turn. New user messages and tool results added after the last provider usage measurement were also invisible to the threshold check, while manual compaction could send unnecessarily large tool results to the summarization model. Telegram exposed raw provider failures directly to chats.

## what

Auto-compaction now runs as a preflight before every model subturn. The latest successful provider-reported usage remains the authoritative checkpoint, with Tau's existing token heuristic applied only to model-visible message content appended after that checkpoint. This allows repeated compactions in one turn and catches reserve consumed by new user messages or completed tool results without replaying model or tool work.

Compaction prompts now middle-truncate textual results from every tool to roughly 2,048 estimated tokens while preserving the persisted live result exactly. Telegram reports stable short turn and compaction failures while retaining detailed causes in adapter logs.

Focused regressions cover repeated same-turn compaction, post-checkpoint tool and user deltas, zero-usage error responses, retained live tool results, generic compaction truncation, and Telegram error sanitization.

## details

Provider errors are not classified and context-overflow requests are not retried. Error and aborted assistant messages do not replace the latest successful usage checkpoint. If a retained assistant/tool batch cannot fit after older context is compacted, Tau preserves that batch unchanged and leaves any exceptional provider rejection as a normal turn failure, matching the decision in the issue discussion. There are no remaining implementation questions.

fixes #338


src/core/telegram/adapter.ts
```diff
@@ -2109,7 +2109,11 @@ class TelegramAdapterImpl {
       await sessionManager.compactSession(session.id);
       await this.reply(chatId, "session compacted. previous context has been summarized.");
     } catch (error) {
-      await this.reply(chatId, this.formatManagerError(error));
+      this.log("error", "telegram session compaction failed", {
+        sessionId: session.id,
+        cause: this.formatManagerError(error),
```
markustherobot: **[P2] Preserve protocol causes in compaction logs**

When `compactSession()` fails through the session protocol, `TauSessionProtocolResponseError.message` is only `session protocol request failed`; the actual provider/compaction diagnostic is in `error.data.cause` (as produced by `session_protocol_handler.ts`). `formatManagerError()` ignores that data, so this new log still discards the detail it is supposed to retain, leaving operators with only the generic message. Extract `data.cause` for logging (or use a shared diagnostic formatter) while keeping the Telegram reply sanitized.

src/core/utils/compact.ts
```diff
@@ -108,13 +108,10 @@ function serializeToolResultMessage(message: ToolResultMessage): string {
...
   const status = message.isError ? "error" : "ok";
 
-  let content = outputText || "(no text output)";
-  if (message.toolName === TOOL_NAME_BASH) {
-    content = truncateForTokens(content, {
-      maxTokens: COMPACTION_BASH_TOOL_RESULT_MAX_TOKENS,
-      strategy: "middle",
-    }).content;
-  }
+  const content = truncateForTokens(outputText || "(no text output)", {
```
markustherobot, line 161: **[P2] Apply the cap to recovery tool results too**

This cap only runs for messages whose role is `toolResult`. If a model stream fails after a tool started, `buildToolRecoveryUserMessage()` embeds each completed tool result in a hidden user `<result-text>` block; `formatHistoryForCompaction()` takes the user branch first and emits that text without truncation. A manual or repeated auto-compaction after tool recovery can therefore still send a large custom/web-search/agent result unchanged and overflow the summary request, contrary to the new every-tool-result bound. Preserve the recovery structure in the compaction representation and cap those embedded result fields as well.

markustherobot (commented):
Found two normal-priority gaps in the compaction bounds and operational diagnostics. The repeated-compaction checkpoint behavior otherwise aligns with the intended design, and the full test suite passes.

markusylisiurunen:
@markustherobot fix

src/core/telegram/adapter.ts
markustherobot: Added a log-only diagnostic formatter that extracts `TauSessionProtocolResponseError.data.cause` and includes it with the generic protocol message. Telegram still receives the stable sanitized reply, and the regression now asserts the detailed protocol cause is retained in logs.

src/core/utils/compact.ts
markustherobot: The compaction formatter now caps every `<result-text>` inside canonical tool-recovery records at the same 2,048-token limit, decoding and re-escaping XML so the record structure remains valid. The regression also confirms the persisted recovery message is unchanged.

PR #344, opened 2026-07-29
add Exa code-mode web tool

### LINKED ISSUE #342: proper code-mode web tools
already have had some web tools but not good enough. need to improve them and would like to make them accessible as code-mode instead of normal individual model tools. could be a good opportunity to figure out a scalable way to introduce code-mode support to other things too or at least build it in a generic way.

a good source: https://research.perplexity.ai/articles/rethinking-search-as-code-generation

markusylisiurunen:
## why

The existing `web_search` and `web_fetch` tools force web research through serial, fixed-shape model tool calls. They also expose large intermediate results directly to the model instead of letting an agent compose retrieval, filtering, batching, and rendering as ordinary code.

## what

Replace both tools with one Exa-backed `web` tool that runs one-shot JavaScript against a bounded `web.discover`, `web.search`, and `web.fetch` facade. Generated programs receive only that facade, bundled documentation, and console output; JavaScript return values are ignored.

Add a thin generic code-mode lifecycle for execution, cancellation, output handling, and UI events. Remove Parallel tools, credentials, protocol values, UI handling, and cost accounting without a compatibility layer.

## details

Generated JavaScript runs in a resource-limited host Worker and SES compartment without process, filesystem, import, or network authority. Exa credentials and search/fetch requests remain in the trusted parent process. Provider calls use abortable host fetch, and Tau tracks and settles all bridge work before a completed, interrupted, or timed-out tool call returns.

Documentation discovery is keyless and runs through the session execution environment. It returns metadata only for negotiated or deterministic Markdown representations and `llms.txt` candidates, never discovered bodies. It can reach any HTTP(S) target available to that environment, including private addresses and non-default ports; deterministic candidates that redirect away from the requested origin are omitted.

The tool is available to main agents, subagents, and ephemeral agents. Generated code and stdout or stderr remain visible through the shared code-mode UI, with explicit termination reasons, UTF-8-safe stream decoding, a 1 MB capture tail, and model-visible output capped at 8,192 estimated tokens.

Configuration moves from Parallel to `apiKeys.exa` or `EXA_API_KEY`.

fixes #342


src/core/static/code_mode/web/runner.mjs
```diff
@@ -0,0 +1,174 @@
...
+
+await initialized;
+
+const exa = Object.freeze({
+  search: (...args) => callExa("search", args),
+  getContents: (...args) => callExa("getContents", args),
+});
+const context = vm.createContext(
+  {
+    console,
```
markustherobot: **[P1] Prevent generated code from escaping the VM**

The injected `console`, timer, and `exa` functions retain the outer child realm. In Node, code such as `console.log.constructor("return process")()` therefore recovers the outer `process` even under `--permission`; from there `globalThis.fetch` is available because Node's permission model does not restrict network access. A web-only agent can consequently bypass the Exa bridge for arbitrary HTTP/SSRF requests or signal processes. `node:vm` is not a security boundary, so run generated code behind an isolation mechanism that exposes no host-realm function references and actually restricts network/process capabilities.

src/core/tools/web.ts
```diff
@@ -0,0 +1,255 @@
...
+
+const input = JSON.parse(require("node:fs").readFileSync(0, "utf8"));
+const home = process.env.HOME;
+if (!home) throw new Error("execution environment HOME is not set");
+
+const parent = join(home, ".cache", "tau", "code-mode", input.name);
+const runtime = join(parent, input.hash);
+const ready = join(runtime, ".ready");
+const runnerPath = join(runtime, "runner.mjs");
+if (existsSync(ready)) {
```
markustherobot: **[P1] Do not trust a mutable cache with the Exa key**

This accepts any execution-environment-controlled directory containing `.ready`, then later launches its `runner.mjs` and passes the host-owned Exa key on stdin. A preseeded cache, or any bash/write-capable agent that replaces `runner.mjs` or `node_modules/exa-js` after the first call, can capture and exfiltrate the key on the next web call. Keep all credential-bearing runner code and dependencies outside agent-writable storage, or otherwise launch verified trusted bytes without a verification-to-execution race.

src/core/tools/web.ts
```diff
@@ -0,0 +1,255 @@
...
+): ToolDefinition {
+  let preparedRuntime: Promise<WebRuntime> | undefined;
+  const apiKey = getExaApiKey(config);
+
+  const implementation: CodeModeToolImplementation<WebArgs, WebRuntime> = {
+    schema: WEB_TOOL,
+    label: "web",
+    outputPolicy: { maxTokens: WEB_CODE_MODE_OUTPUT_TOKENS },
+    parseArguments: parseWebArguments,
+    prepare: async () => {
```
markustherobot: **[P2] Forward cancellation through runtime preparation**

The first call can spend up to five minutes in `npm ci`, but this hook discards its `signal` and `prepareWebRuntime` invokes `runNodeScript` without one. Terminating the subagent during installation therefore waits for npm to finish or hit the preparation timeout instead of following the normal abort path promised by code mode. Accept the signal here and pass it to the preparation backend call.

src/core/static/code_mode/web/documentation.md
```diff
@@ -0,0 +1,764 @@
...
+});
+```
+
+## Request Parameters
+
+| Parameter | Type | Default | Description |
+| --- | --- | --- | --- |
+| `query` | string | **(required)** | Natural language search query. Supports long, semantically rich descriptions. |
+| `type` | string | `"auto"` | Search method: `auto`, `fast`, `instant`, `deep-lite`, `deep`, `deep-reasoning`. |
+| `stream` | boolean | `false` | If `true`, returns `text/event-stream` with OpenAI-compatible chat completion chunks instead of a JSON body. |
```
markustherobot: **[P2] Do not advertise unsupported search streaming**

The pinned `exa-js@2.16.3` implementation throws whenever `exa.search(..., { stream: true })` is used and directs callers to `streamSearch()`, but the bridge exposes only `search` and `getContents`. Generated code following this bundled documentation will always fail. Remove the streaming option and section from the progressive docs, or expose and transport the SDK's streaming method explicitly.

markustherobot (commented):
The code-mode path needs hardening before merge. The main concerns are preserving the intended credential and tool isolation boundaries and honoring normal cancellation semantics; the progressive documentation also needs to match the SDK surface that the bridge actually exposes.

markusylisiurunen:
@markustherobot fix

some of the findings may be outdated already, just fyi.

also a comment to this:

> Discovery reports candidate Markdown representations and returns up to 20,000 characters of raw llms.txt

the discover method should not return actual content at all from any of the discovered pages. it should simply return metadata and the model may decide to use e.g. curl to fetch it.

what does this mean: "pin resolved public addresses for each request"?

was Parallel removed completely? if not, should be.

src/core/static/code_mode/web/runner.mjs
markustherobot: I replaced `node:vm` with a host-owned SES compartment. Generated code has only the documented `web`, `docs`, and console capabilities; process, fetch, imports, and the raw bridge capabilities are unavailable, and regression coverage exercises the original constructor escape.

src/core/tools/web.ts
markustherobot: I removed the execution-environment cache entirely. Tau now loads trusted sandbox bytes at startup, launches those bytes directly, and keeps both the Exa dependency and API key in the trusted parent process; the child receives neither the key nor provider client.

src/core/tools/web.ts
markustherobot: The preparation phase no longer exists because the mutable runtime was removed. Cancellation now terminates the running SES sandbox directly, with regression coverage for aborting non-terminating generated code.

src/core/static/code_mode/web/documentation.md
markustherobot: The bounded facade and bundled documentation now state that streaming is unsupported. Unknown options such as `stream` fail validation, and the regression test covers that rejection.

markusylisiurunen:
@markustherobot deslop

markusylisiurunen:
@markustherobot review

src/core/tools/web.ts
```diff
@@ -0,0 +1,533 @@
...
+      if (settled || terminating) return;
+      try {
+        worker.postMessage(message);
+      } catch {}
+    };
+
+    worker.stdout.on("data", (chunk: Buffer) => capture("stdout", chunk));
+    worker.stderr.on("data", (chunk: Buffer) => capture("stderr", chunk));
+    worker.on("message", (message: WebWorkerRequest) => {
+      void handleWebRequest(message, exa, deps, backend, signal).then(
```
markustherobot: **[P1] Cancel provider work when the sandbox stops**

Each Worker request starts an untracked host promise, while abort and timeout only terminate the Worker and suppress `postResponse`. For `web.search` and `web.fetch`, the signal is not passed into the provider client, so an interrupt or the 60-second timeout can return while HTTP calls continue, and a generated `Promise.all` batch can keep many paid requests alive. Tie bridge requests to the tool lifecycle and cancel or otherwise settle all in-flight provider work when terminating the sandbox.

src/core/tools/code_mode.ts
```diff
@@ -0,0 +1,135 @@
...
+  if (model.outputBytes === 0 && exitCode === 0) {
+    return "Program produced no output (exit 0)";
+  }
+
+  const output = model.content.trimEnd();
+  const truncationNote =
+    model.truncated || captureTruncated
+      ? `\n\n[Output truncated for context: ${model.outputLines} lines / ${formatBytes(model.outputBytes)} shown of ${model.totalLines} lines / ${formatBytes(model.totalBytes)} (full output estimate: ~${bytesToTokens(model.totalBytes)} tokens).${fullOutputPath ? ` Full output saved to ${fullOutputPath}.` : ""}]`
+      : "";
+  const exitNote = exitCode !== null && exitCode !== 0 ? `\n(exit ${exitCode})` : "";
```
markustherobot: **[P2] Preserve timeout and abort reasons in tool output**

The Worker maps both timeout and abort to `exitCode: null` and returns the corresponding flags, but this formatter receives only the exit code and emits no note for `null`. A 60-second timeout with no stdout therefore becomes an error result containing only `(no output)`, so neither the model nor the user can tell whether the program timed out, was interrupted, or failed another way. Thread the termination state through the code-mode result/UI and include the same explicit reason that Bash execution reports.

src/core/tools/web.ts
```diff
@@ -0,0 +1,533 @@
...
+    let truncated = false;
+    let timedOut = false;
+    let aborted = false;
+    let settled = false;
+    let terminating = false;
+
+    const capture = (target: "stdout" | "stderr", chunk: Buffer): void => {
+      capturedBytes += chunk.length;
+      if (capturedBytes > DEFAULT_COMMAND_CAPTURE_BYTES) truncated = true;
+      const text = chunk.toString("utf8");
```
markustherobot: **[P2] Decode Worker output across chunk boundaries**

Readable stream chunks are not guaranteed to end on UTF-8 character boundaries. Decoding each chunk independently replaces a multibyte character split between chunks with replacement characters, which can corrupt large or non-English web results. Keep a `StringDecoder` per stream and flush it at the end before appending captured text.

markustherobot (commented):
The architecture and contract migration are coherent, but the Worker lifecycle and output path still have correctness gaps that affect cancellation and result fidelity.

markusylisiurunen:
@markustherobot fix

src/core/tools/web.ts
markustherobot: I replaced the non-cancellable SDK calls with the same two Exa HTTP operations through abortable host `fetch`. Every bridge request now uses a lifecycle signal, is tracked, and is awaited after cancellation or Worker exit, so interrupt, timeout, and unawaited generated calls cannot leave provider work running.

src/core/tools/code_mode.ts
markustherobot: Code mode now derives Bash-style termination notes from the complete execution result before output preparation. Abort, timeout (including the configured duration), and close-signal reasons are therefore present in model text, expanded output, and compact UI previews.

src/core/tools/web.ts
markustherobot: Worker stdout and stderr now each use a `StringDecoder`, and both decoders are flushed after their streams finish before the result resolves. The regression test drives output beyond the capture boundary with three-byte characters and verifies that no replacement characters are introduced.

PR #347, opened 2026-07-30
convert Nook to code mode

### LINKED ISSUE #345: convert Nook to code-mode
relevant once issue #342 is merged.

markustherobot:
## why

Nook's operation-union tool could execute only one fixed action per call and had a separate UI and lifecycle from the web code-mode tool. Converting it to the same bounded code-mode shape allows supported Nook operations to compose without exposing credentials, raw network access, or direct filesystem authority to generated code.

## what

The `nook` tool now accepts one JavaScript `code` string and exposes a capability-limited `nook` facade for sites, templates, and per-site KV. Static `docs` describe this agent-facing SDK, while `nook.skill()` separately loads the configured deployment's version-matched app-authoring guide. List methods return arrays directly, KV get returns its value directly, site deployment requires an explicit `visibility: "private" | "public"` choice, and KV values can be transferred to and from JSON files through `getToFile` and `putFromFile`.

Web and Nook now share a narrow host-Worker executor for resource limits, UTF-8-safe output capture, timeout and interruption handling, bridge request tracking, and settlement. The executor caps each program at 64 bridge requests and four concurrent requests before privileged handlers are dispatched. Each tool retains its own SES sandbox and API contract. Nook HTTP, Cloudflare Access credentials, manifest validation, and execution-environment file operations remain in the trusted parent.

Nook HTTP requests and multi-file manifest, copy, download, and upload paths now observe cancellation. The old operation schema and Nook-specific tool rendering path were removed in favor of the generic code-mode lifecycle.

## details

The facade intentionally does not expose raw requests, URLs, general file access, manifests, deployment sessions, or setup/destroy operations. The dedicated KV file methods serialize or parse bounded JSON through the execution backend. Copy destinations must still exist and be empty, and existing hidden-file, symlink, reserved-path, size, hash, and manifest rules remain authoritative.

The implementation treats direct array results for all list methods, including `nook.kv.list`, and a direct JSON value from `nook.kv.get` as the canonical code API.

fixes #345


src/core/tools/code_mode_worker.ts
```diff
@@ -0,0 +1,185 @@
...
+    worker.stderr.on("data", (chunk: Buffer) => capture("stderr", stderrDecoder, chunk));
+    worker.on("message", (message: unknown) => {
+      if (settled || terminating) return;
+      if (!isWorkerRequest(message)) {
+        workerError = new Error("code-mode sandbox sent an invalid bridge request");
+        requestController.abort();
+        void worker.terminate().catch(() => {});
+        return;
+      }
+      const request = options.handleRequest(message, requestSignal).then(
```
markustherobot, line 157: **[P1] Bound bridge request fan-out**

Generated code can post requests without awaiting them (for example, a loop of `nook.sites.deploy(...)`), and every message here immediately invokes `handleRequest` and adds another promise. The Worker's 128 MiB and time limits do not constrain these parent-side operations: each Nook deploy may retain up to 100 MiB of file buffers, and mutating HTTP requests can be launched by the thousands before timeout or abort runs, so one tool call can exhaust the host or flood authenticated state changes. Enforce a small total and in-flight bridge budget, terminating or rejecting further requests when it is exceeded, before dispatching handlers.

markustherobot (commented):
The facade and authority split are coherent, and both CI and the local test suite pass. One host resource-boundary issue remains in bridge scheduling.

markusylisiurunen:
@markustherobot fix

in addition to the findings, I would like one behavioral change. I want the KV SDK API to provide ways to read and write KV values from disk. as an example, read a kv value to a file on disk for later processing. hope this makes sense, should still be kept lean.

src/core/tools/code_mode_worker.ts
markustherobot: The shared executor now stops a program before dispatch when it exceeds either four concurrent bridge requests or 64 total requests. Excess fan-out aborts and settles already active handlers, and focused tests cover both limits.

PR #355, opened 2026-07-31
unify agent execution runtime

### LINKED ISSUE #351: unify agent execution behind one canonical runtime
## problem

Tau currently has two independent model/tool execution loops:

- Main sessions execute through `ChatRuntime` / `ConversationTurnRuntime` / `CoreSession` / `SessionEngine.processTurn()`.
- Background subagents execute through the separate `runSubagent()` loop and `SubagentControlPlane`.
- Ephemeral diff-review threads already reuse `CoreSession` and `SessionEngine`, but wrap them with their own lifecycle, usage, and thread handling.

These agents have the same fundamental behavior: a system prompt, model, conversation history, tools, model subturns, tool execution, retries, context accounting, and events. Their actual differences are integration concerns such as persistence, protocol projection, parent/child supervision, thread forking, and progress presentation.

The duplicated subagent loop has already drifted from the main loop. It does not support automatic compaction, and it differs in tool-call streaming, recovery after tools have started, retry budgeting, event detail, and usage handling. Every new turn-level feature can be implemented for one loop and omitted from the other.

The current abstractions also mix unrelated concerns:

- `SessionEngine` contains the canonical execution behavior but is shaped around main-session concepts.
- `ConversationTurnRuntime` primarily owns interruption and event consumption rather than the actual turn loop.
- Tools receive a large `ToolDispatchContext` containing configuration, registries, model resolution, subagent control-plane state, paths, and a `main` / `subagent` scope union.
- Agent identity is hardcoded in places. For example, `SessionEngine` records usage as `main` even when an ephemeral thread uses it, while the ephemeral wrapper records another usage entry.
- Persistence, protocol projection, execution, and presentation boundaries are not consistently explicit.

Issue #349 identifies the immediate auto-compaction problem. This issue supersedes that narrow scope with the canonical architecture needed to solve it without adding another compatibility layer.

Pruning is intentionally out of scope for preservation. The `/prune:*` feature has not proven useful alongside automatic and manual compaction and should be removed completely during this migration.

## goal

Introduce one opinionated, stateful `AgentRuntime` used by main sessions, background subagents, and ephemeral threads.

The runtime must own all complex and behavior-sensitive agent execution semantics. Context-specific wrappers must supply resolved configuration and dependencies, then project the runtime's semantic events into persistence, session protocol state, subagent progress, or ephemeral thread updates.

This is a canonical breaking refactor. Update all call sites to the new shape. Do not preserve the old loops, dispatch context, pruning APIs, or compatibility adapters.

### one runtime per conversation

One `AgentRuntime` represents one serial conversation with mutable history and at most one active logical turn. Multiple agents run concurrently through external supervisors or context wrappers.

The runtime owns:

- Canonical conversation history and history-entry identities
- System prompt and resolved model settings
- Context construction and model notices
- Model streaming and draft accumulation
- Provisional and completed tool-call admission
- Sequential tool execution while later tool calls continue streaming
- Assistant/tool-result commit ordering
- Provider retries and retry budgets
- Recovery when a provider stream fails after tools have been admitted
- Context usage checkpoints
- Automatic and manual compaction
- Model-subturn limits
- Steering at safe model/tool boundaries
- Interruption
- Semantic agent events

The runtime does not own:

- Session snapshot persistence
- Session protocol requests, deltas, or request correlation
- TUI rendering
- Usage-log storage or agent-kind attribution
- Normal pending-message queue policy
- Parent/child agent records and concurrency limits
- Ephemeral context/thread maps or thread forking
- Execution-environment or Tau configuration discovery
- Persona, skill, or AGENTS.md resolution

### spec, state, and active execution

Keep three concepts explicit.

`AgentSpec` is the complete resolved behavior used to run the agent. It should contain the system prompt, model and streaming settings, bound tool set, retry policy, compaction policy, and execution limits. Runtime/config/bootstrap code resolves Tau configuration into this spec before constructing or updating the agent.

`AgentState` is the small durable and restorable semantic state. It should contain the agent id, state revision, canonical history entries, and the context/prompt epoch metadata required to validate provider-reported usage checkpoints. Do not require hidden invalidation flags whose meaning cannot survive restoration.

Active execution state is private and non-serializable. It includes abort controllers, provider streams, draft messages, provisional tool calls, active tool runs, retry budgets, pending steering, and boundary-stop state. Recovery creates an idle agent from committed state; it does not attempt to resurrect provider streams or processes. Canonical handling of dangling committed tool calls/results belongs in the runtime's restoration/recovery semantics.

A representative API is:

```ts
interface AgentRuntime {
  readonly status: "idle" | "running";
  readonly state: Readonly<AgentState>;
  readonly spec: Readonly<AgentSpec>;

  submit(input: AgentInput): AgentTurn;
  steer(input: AgentInput): SteeringSubmission;
  cancelSteering(): AgentInput[];
  interrupt(): boolean;

  updateSpec(spec: AgentSpec): void;

  compact(options: CompactOptions): Promise<CompactionResult>;
  rewind(target: HistoryEntryId): RewindResult;
  retry(): AgentTurn;

  snapshot(): AgentState;
  dispose(): void;
}
```

Exact names may follow local conventions, but the ownership and command boundaries above are required. There is no prune operation.

A logical turn captures the effective model/settings/tool set when it begins. Configuration changes made during a turn apply to the next submitted or steering turn, not later tool subturns of the active turn.

### one awaited semantic output port

The runtime must have one required asynchronous event sink rather than an internal pub-sub system or a collection of lifecycle hooks:

```ts
type AgentEventSink = (event: AgentEvent) => Promise<void>;
```

The runtime awaits every event. This preserves strict ordering and backpressure, lets a session projection commit before later model events overtake it, and gives event-delivery failure one explicit policy. If the sink fails during active execution, abort that execution and surface the sink failure.

Outer adapters may fan events out to persistence, protocol projection, logging, and UI concerns. Fan-out is not part of `AgentRuntime`.

The runtime owns canonical state. Events describe already-applied semantic transitions; they are not an event-sourced persistence log and consumers do not determine execution semantics by reducing them.

For durable transitions:

1. Mutate canonical agent state.
2. Increment the durable state revision.
3. Emit the semantic event.
4. Await the sink.
5. Continue execution.

Transient events such as assistant drafts, provisional tool calls, and tool activity do not increment the durable state revision.

### event fidelity

The event contract must preserve all existing protocol and TUI behavior without exposing raw provider noise or partial executable arguments.

Required semantic event categories include:

- Logical turn started/completed/stopped/interrupted/blocked
- User message committed
- Assistant draft started and cumulatively updated
- Provisional tool call preparing, replaced, or discarded
- Complete tool call admitted
- Tool run queued, started, activity, succeeded, failed, blocked, or cancelled
- Final assistant message committed
- Tool result committed
- Model retry scheduled/started
- Tool recovery committed
- Compaction started/completed/skipped/failed/aborted
- Usage checkpoint updated
- Typed notices for remaining human-readable conditions

Assistant draft updates may remain coalesced cumulative snapshots. Session protocol adapters can derive efficient append deltas from successive snapshots.

A provisional tool-call event must expose only stable identity available during generation, such as assistant message id, content index, tool call id, and tool name. Never expose partial arguments. Once the call is complete, emit the canonical full call and queue it immediately even if the model continues generating later calls.

Keep model-side tool-call state separate from execution-side tool-run state. Tool execution may physically finish before the final assistant message is committed. The runtime must distinguish execution completion from committing the corresponding tool result into conversation history.

### canonical turn loop

Keep the internal implementation as a readable imperative loop with typed phase results. Do not introduce a generic effect interpreter or configurable state-machine framework.

Before every model subturn:

1. Validate whether automatic compaction is needed using a fresh usage checkpoint and the active model's context window.
2. Run compaction when required.
3. Start and stream an assistant draft.
4. Emit provisional tool-call identities as they become stable.
5. Admit complete calls and queue sequential execution immediately.
6. Continue model streaming while admitted tools execute.
7. Reconcile admitted calls with the provider's final assistant message.
8. Commit the final assistant message.
9. Commit buffered tool results in canonical call order.
10. Apply tool-stream recovery when needed.
11. At the safe boundary, honor interruption or pending steering before deciding whether another model subturn is needed.

A submitted user message is committed and its event is acknowledged before model work starts. If model execution or automatic compaction later fails, the accepted user message remains recoverable.

Preserve the existing model-subturn limit and explicit warning/outcome when it is reached.

### retry and recovery

Encode two retry classes explicitly.

A provider failure before any complete tool call has been admitted may retry the same model subturn against the same context. Emit a typed retry event and preserve the configured delay and bounded budget.

After a tool has been admitted, a blind provider retry can duplicate side effects. Instead, reconcile the partial/final assistant message, finish or account for admitted tools, create explicit interrupted/unknown results where completion cannot be proven, commit the canonical recovery message, and continue through a new model subturn only when the recovery budget permits.

Main, subagent, and ephemeral agents must use exactly the same retry and recovery implementation.

### compaction

Automatic compaction must run before every model subturn for every agent type using the effective `autoCompact` settings and that agent's active model context window.

The runtime owns threshold detection, usage-checkpoint freshness, retained-tail selection, summary/preserved-user-message handling, history replacement, continuation messages, and compaction events.

A failed automatic compaction blocks the turn rather than proceeding toward likely context overflow. The already-committed user input remains in state.

Manual compaction uses the same canonical transformation machinery and is an explicit idle-only operation.

### steering

Steering is the runtime's only internal priority input lane.

When steering arrives during an active turn:

1. Accept and retain the steering input.
2. Request a stop at the next safe model/tool boundary.
3. Finish the current logical turn at that boundary.
4. Combine all pending steering inputs using the canonical hidden steering guidance.
5. Start a new logical turn using the latest `AgentSpec`.
6. Associate each steering submission with that resulting turn.

Cancelling all unapplied steering also cancels the pending boundary stop. Normal queued submissions remain an outer host/supervisor policy because main sessions queue them while subagent and ephemeral contexts currently reject concurrent normal input.

### fully bound tools

Replace the large `ToolDispatchContext` and `main` / `subagent` scope branching with fully dependency-bound tools assembled outside the runtime.

Representative construction:

```ts
const tools = [
  createBashTool({ backend, cwd }),
  createEditTool({ backend, cwd }),
  createWebTool({ worker, credentials }),
  createSpawnAgentTool({ supervisor, resolveChildSpec }),
  createClientTool({ broker, definition }),
];
```

Representative interface:

```ts
interface AgentTool {
  readonly schema: ToolSchema;

  describe(call: ToolCall): ToolCallDescription;

  execute(
    call: ToolCall,
    context: ToolExecutionContext,
  ): Promise<ToolExecutionOutcome>;
}
```

The execution context contains only execution-local facts:

```ts
type ToolExecutionContext = {
  agentId: string;
  turnId: string;
  assistantMessageId: string;
  signal: AbortSignal;
  emitActivity(activity: ToolActivity): Promise<void>;
};
```

`describe()` is synchronous and side-effect free. It supplies tool-owned header target, code preview, and related descriptive metadata.

The runtime emits normalized queue/start/finish/block/cancel lifecycle events. Tools emit typed tool-owned activity through `emitActivity`; the runtime passes this through without understanding bash, edit, web, Nook, or subagent presentation details.

Tools return content/outcome rather than constructing canonical model messages. The runtime assigns tool-call identity, name, timestamp, error state, and canonical `ToolResultMessage` shape. It also catches thrown tool errors and handles cancellation.

Constructing an `AgentSpec` must fail on duplicate tool names. A missing tool or invalid call is rejected canonically by the runtime.

Dynamic client tools and configuration-derived tools are bound before the logical turn and frozen in that turn's captured spec.

### external supervision and wrappers

Background child agents are ordinary `AgentRuntime` instances managed by an external `AgentSupervisor`. The supervisor owns active-agent limits, parent/origin relationships, child spec resolution, records, follow-up submissions, wait/terminate behavior, progress aggregation, cost aggregation, and cleanup after parent rewind. The spawn tool is a normal bound tool with the supervisor as a dependency.

A child without a bound spawn tool cannot create nested children. Do not enforce this through an agent-kind or dispatch-scope check.

Ephemeral contexts own their thread map, busy checks, state cloning for forks, request/response mapping, and disposal. Each thread uses the same runtime.

Main hosted sessions own persistence, recoverable session snapshots, protocol deltas, request correlation, pending normal submissions, client attachment, and projection of semantic agent and supervisor events.

There must be no subclasses and no `main` / `subagent` / `ephemeral` behavioral branches inside `AgentRuntime`. Agent-kind attribution belongs to usage/progress adapters. This must also remove duplicate or misclassified ephemeral usage logging.

## remove pruning

Delete pruning end-to-end rather than porting it:

- Core pruning implementation and types
- `/prune:earliest`, `/prune:largest`, and `/prune:smart`
- Session protocol `session.prune` request/result schemas and handlers
- SDK prune methods and types
- TUI command handling and autocomplete
- Pruned tool-result projection/UI handling that exists only for this feature
- Pruning tests and fixtures
- README, AGENTS.md, and docs references

Do not retain aliases, deprecated methods, migration branches, or compatibility shims.

## implementation sequence

Use an incremental migration, but keep only the canonical architecture in the final tree.

1. Add characterization tests for the existing main-loop behaviors that must remain invariant: provisional tool calls, immediate sequential admission during model streaming, assistant/tool-result ordering, safe retries, post-tool recovery, boundary stopping, compaction ordering, and sink backpressure.
2. Introduce canonical agent spec/state/event/turn types and the single awaited event sink.
3. Move the existing `SessionEngine` model/tool/compaction/retry logic into the new opinionated runtime without redesigning individual edge-case behavior.
4. Replace `ToolDispatchContext` with externally bound tools and a narrow execution-local context. Update every built-in, configured, client-provided, and subagent tool.
5. Adapt the main hosted session to project runtime events into snapshots, protocol deltas, persistence, tool facets, and usage logs.
6. Adapt ephemeral contexts/threads to construct the same runtime and project events into thread updates. Preserve fork, busy, close, and side-channel behavior.
7. Replace `runSubagent()` with the same runtime. Keep `SubagentControlPlane` responsibilities only where they belong to supervision, or replace it with the canonical supervisor abstraction.
8. Move steering application into the runtime while leaving normal queued-submit policy in the session host.
9. Remove old loops, wrappers made redundant by the new runtime, scope checks, hardcoded agent-kind logging, and compatibility code.
10. Delete pruning end-to-end.
11. Update architecture, protocol, SDK, commands, and configuration documentation to match the final code.

## acceptance criteria

- Main sessions, background subagents, and ephemeral threads execute through the same `AgentRuntime` implementation.
- `runSubagent()` and every other independent model/tool loop are deleted.
- There is one canonical implementation of model subturns, tool admission/execution, retries, tool recovery, context accounting, limits, steering boundaries, and compaction.
- Subagents auto-compact during a multi-tool-call run using their effective auto-compaction config and model context window.
- Main, subagent, and ephemeral execution preserve provisional tool-call UX, immediate completed-call queueing, sequential execution, rich tool activity, and canonical assistant/result ordering.
- Partial tool arguments never enter runtime events, session protocol state, or executable tool calls.
- Safe provider retries and post-tool recovery have parity across all agent contexts.
- The runtime exposes one awaited semantic event sink. A sink failure aborts active execution and cannot be overtaken by later events.
- Durable agent state is restorable without hidden non-serialized context-validity flags.
- Steering is applied by the runtime at safe boundaries; normal queued submissions remain outside it.
- Tools are fully dependency-bound before entering the runtime. The large dispatch context and `scope: "main" | "subagent"` checks are removed.
- Child agents are supervised externally and are ordinary instances of the same runtime.
- Persistence and session protocol projection remain outside the runtime.
- Usage attribution is supplied by adapters; ephemeral work is neither double-logged nor recorded as main-agent usage.
- Pruning has been removed from core, protocol, SDK, TUI, tests, and documentation.
- No compatibility aliases or parallel legacy implementation remain.
- `npm run check`, `npm run build`, and `npm test` pass.

## required tests

At minimum, cover:

- Main, subagent, and ephemeral construction through the same runtime
- User input committed before model work and retained after model/compaction failure
- Assistant draft streaming and coalescing
- Provisional tool call prepare/replace/discard behavior without partial arguments
- Multiple streamed tool calls admitted as soon as complete and executed sequentially
- Tool completion before assistant final, with canonical history commit ordering
- Safe retry before tool admission
- Recovery after provider failure with admitted/completed tools, including no duplicate execution
- Auto-compaction before a later subturn in a long-running subagent multi-tool sequence
- Same threshold and retained-tail semantics for main and subagent models
- Failed automatic compaction blocking the turn
- Steering acceptance, batching, cancellation, boundary stop, and next-spec capture
- Interruption during model streaming and tool execution
- Awaited event-sink ordering and sink failure
- Spec updates applying only to later logical turns
- Ephemeral thread continuation and fork behavior
- Child follow-up submission, wait, terminate, active limit, and origin cleanup
- Usage attribution for all three contexts without duplication
- Absence of pruning commands and protocol/SDK surface

### LINKED ISSUE #356: treat aborted subagents as successful terminate calls
## problem

Terminating a running subagent succeeds at the control-plane level but is presented as a failed tool call:

```text
✗ terminate failed <agent-id>
  Status: aborted
  (cost $1.24 · duration 3m 15s · Status aborted)
```

`SubagentControlPlane.terminate()` aborts the running agent and returns its expected final status, `aborted`. `createTerminateAgentToolDefinition()` then defines tool-call success as `result.status === "success"`, so the expected result of terminating a running agent becomes an error UI event and an error tool result.

This also causes `Status: aborted` to appear twice, once in the output body and once in the status line.

Relevant locations:

- `src/core/subagents/control_plane.ts:278-296`
- `src/core/subagents/control_plane.ts:362-367`
- `src/core/tools/terminate_agent.ts:146-171`
- `src/tui/ui/tool_ui_registry.ts:553-571`

The existing registry test at `test/tool_ui_registry.test.js:192-202` already represents `finalStatus: "aborted"` with tool status `"success"`, but there is no dispatch-level test covering this mapping.

## expected behavior

A terminate request that successfully aborts a running subagent should render as `terminated` and produce a successful tool result whose payload reports the agent's final status as `aborted`.

Malformed or unknown IDs and actual termination exceptions should remain tool failures. Add a dispatch-level regression test so the final agent status and terminate operation status cannot be conflated again.

### LINKED ISSUE #357: preserve the live subagent panel across main-session updates
## problem

The subagent panel is cleared and reconstructed when unrelated main-session state is reconciled. This is visible when submitting another user message while subagents are still running, and can also happen during slash actions such as compact, auto-compact, rewind, reload, and other snapshot synchronization paths.

The panel should behave as an independently living, real-time window into active subagents. Main-session message and history operations should not erase the activity lines currently showing what those agents are doing.

The current path is:

1. `SessionChatController.syncRenderedHistory()` calls `syncSnapshotToolAndAgentUi()`.
2. `syncSnapshotToolAndAgentUi()` calls `view.reconcileToolUiSession()`.
3. `TuiChatView.reconcileToolUiSession()` unconditionally calls `subagentPanel.reset()`.
4. The controller reconstructs agents from snapshot state, but each protocol agent stores only one `progress` string rather than the panel's accumulated activity window.

Relevant locations:

- `src/tui/session_chat_controller.ts:1336-1384`
- `src/tui/session_chat_controller.ts:1426-1436`
- `src/tui/chat_view.ts:292-295`
- `src/tui/session_chat_controller.ts:2578-2592`
- `src/host/local_session_host.ts:2318-2334`

This can leave the reconstructed panel completely blank. Some tool UI events are converted to an empty progress string in `src/core/subagents/control_plane.ts:423-448`; the live panel ignores that blank event while retaining earlier lines, but reconstruction has already discarded those earlier lines.

## expected behavior

The panel should continuously reflect realistic current subagent state independently of main-session rendering and history changes:

- Submitting, queueing, or steering user messages must not clear active subagent activity.
- Compact, auto-compact, rewind, reload, pruning, and similar snapshot operations must not indiscriminately reset the panel.
- Activity history for a surviving subagent should remain intact while new events continue to update it in real time.
- Reconciliation should update canonical status, usage, cost, selection, and agent membership without replacing unchanged panel entries or clearing their activity lines.
- An entry should disappear or transition only when canonical agent state actually requires it, such as the agent finishing, being terminated, being removed by rewind, or the client switching to a different session.

Add regression coverage using a stateful subagent-panel view rather than only asserting emitted controller events. The test should accumulate multiple activity lines, apply representative main-session snapshot updates, and verify that the live panel history and selected running agent survive.

markusylisiurunen:
## why

Main sessions, background subagents, and ephemeral review threads had separate model/tool loops with different retry, recovery, streaming, usage, and compaction behavior. That duplication made turn-level features drift between agent contexts and left persistence, supervision, tool binding, and execution semantics mixed together.

## what

Replace those loops with one stateful `AgentRuntime` shared by all three contexts. The runtime owns canonical history and revisions, frozen logical-turn specs, model streaming, provisional and completed tool admission, awaited lifecycle/activity events, sequential execution, retries, post-tool recovery, usage checkpoints, compaction, steering boundaries, interruption, rewind, and restoration.

Tools are now fully dependency-bound before entering a turn and receive only execution-local context. Background children are ordinary runtimes managed by an external `AgentSupervisor`, while ephemeral contexts own only thread lifecycle and state cloning. Hosted sessions project semantic runtime events into persistence and protocol state, and each adapter supplies its own usage attribution.

Pruning is removed from core, protocol, SDK, TUI, tests, and documentation. Tool lifecycle presentation now derives from canonical protocol status, while activities only enrich tool-specific presentation. Subagent reconciliation preserves live panel history and selection across unrelated main-session updates, and successful termination treats the expected final `aborted` agent status as a successful terminate operation.

## details

- Preserve provisional tool identity without exposing partial arguments.
- Acknowledge admission and queued events before tool side effects begin.
- Emit physical tool completion separately from canonical result commits.
- Apply one retry, recovery, context accounting, and auto-compaction implementation everywhere.
- Move active steering into the runtime while keeping normal queued submissions in the host.
- Serialize submission ownership across input persistence and steering startup boundaries.
- Keep interrupted persistence, observer deltas, and lifecycle state consistent.
- Preserve independently accumulated subagent activity during tool and snapshot reconciliation.
- Remove the legacy session engine, core session, turn wrapper, subagent loop/control plane, dispatch context, and duplicate batch tool runner.
- Add focused coverage for sink backpressure, retries, recovery, compaction, steering, interruption, restoration, supervision, ephemeral forks, usage attribution, termination semantics, and reconciliation.

fixes #351
fixes #356
fixes #357


src/host/local_session_host.ts
```diff
@@ -201,8 +205,11 @@ export class LocalSessionHost implements TauSessionHost {
         undefined,
         recovered.changed,
       );
-      hostedSession.session.restoreState({
-        sessionId: recovered.snapshot.sessionId,
+      const currentState = hostedSession.runtime.snapshot();
+      hostedSession.runtime.restoreState({
+        agentId: recovered.snapshot.sessionId,
+        revision: recovered.snapshot.revision,
+        contextEpoch: currentState.contextEpoch,
```
markustherobot: **[P1] Restore the usage checkpoint after recovery**

`AgentRuntime.shouldRunAutoCompaction()` now returns false unless `state.usageCheckpoint` exists and matches the context epoch, but this recovery state rebuild omits it even though recovered assistant messages retain provider usage. Every recovered session therefore skips threshold compaction on its first post-restart subturn; a near-limit session is sent directly back to the provider and may context-overflow. Reconstruct the latest valid checkpoint from persisted history (or persist it canonically) and include it in the restored state.

src/host/local_session_host.ts
```diff
@@ -708,6 +714,27 @@ class LocalHostedSessionHandle implements LocalHostedSession {
...
+  }> {
+    this.assertActive();
+    const activeTurn = this.activeTurnPromise;
+    if (!activeTurn) {
+      throw new Error("cannot steer without an active turn");
+    }
+    const association = await this.runtime.steer(text);
+    return {
+      userHistoryEntryId: association.historyEntryId,
+      turn: await activeTurn,
```
markustherobot: **[P1] Persist the outcome for the steering turn**

`runtime.steer()` associates the steering entry with a new logical turn, but `activeTurn` represents the entire outer run and `runTurnNow()` captured the original user entry before any steering. The final result is therefore stored only under that original entry, while `association.historyEntryId` never receives a persisted `turn` outcome; with multiple steering turns, earlier steer responses also await the last logical turn. Track completion by `association.turnId` and store/return the resulting outcome for the steering history entry.

markustherobot (commented):
Found two host-adapter regressions in recovery and steering outcome projection. Both can leave persisted/session-protocol state inconsistent with AgentRuntime's canonical semantics.

markusylisiurunen:
did some brief testing and noticed at least these regressions/bugs:

- nook tool was not available to the model even though I had nook config in place in config.json.
- using tui, steering apparently was sent to the model but the message stayed in the tui steering queue visually, was not shown in the chat history list as a user message, but once the model finished, the steering message appeared visually after the agent's final message. that's clearly wrong. it should appear in the correct place.

src/host/local_session_host.ts
markusylisiurunen: Fixed in 709cfeb. The session snapshot now persists a required, independent `agentState` with the canonical agent revision, context epoch, and usage checkpoint. Recovery restores that state instead of substituting the protocol revision, and a regression test proves a near-limit recovered session auto-compacts before its first post-recovery model call.

src/host/local_session_host.ts
markusylisiurunen: Fixed in 709cfeb. Steering submissions now expose separate `applied` and per-logical-turn `result` promises. The host persists each steering history entry as soon as it is applied, then stores and returns that steering turn’s own outcome rather than the outer run’s initial/final result. Batched steering and distinct initial/steering outcomes are covered by regression tests.

markusylisiurunen:
Follow-up fixes are in `709cfeb`.

This addresses the testing regressions and the architecture review together:

- Configured Nook is bound independently of persona tool filtering, while persona-disabled host tools remain excluded.
- Steering registration is atomic. Applied steering is projected and removed from pending state immediately in canonical history order, while the request waits for that logical turn's own persisted outcome. Every unapplied terminal path rejects both steering promises.
- Child write, edit, view-image, web, and bash tools all use the child-scoped backend.
- Main snapshots persist independent canonical agent revision/context/checkpoint state, including first-turn post-recovery auto-compaction.
- Tool outcomes are semantic (`succeeded`, `failed`, `blocked`, `cancelled`); presentation activity only adds facets and cannot overwrite status.
- Active child context is injected into compaction continuation state.
- `AgentRuntime` now receives an execution-ready model executor, resolved notice/settings/policies, and bound tools. It no longer owns Tau config, personas, auth/model construction, or TUI event types.
- The breaking session wire contract is version 4.

Verification passes: `npm run check`, `npm run build`, `npm test` (72 files, 764 tests), and `git diff --check`.

markusylisiurunen:
Pushed follow-up `0a6f978` addressing the accepted correctness findings.

- compaction now inherits captured stream options, cleans provider sessions, and keeps model notices out of maintenance calls and compacted history
- post-admission iterator failures enter canonical tool recovery, queued tools settle as cancelled, and stopped consumers release pending lifecycle/activity acknowledgements
- committed user events persist before acknowledgement
- steering validates before pending publication, cancellation targets only unapplied runtime submissions, and `session.steer` no longer accepts `historyEntryId`
- agent-state patch application and usage checkpoints now enforce complete snapshot invariants
- successful compaction mutation is no longer followed by a contradictory failure event when success delivery fails
- rewind is an awaited, idle-only runtime transition projected through the semantic event sink

Verification passed: `npm run check`, `npm run build`, `npm test` (72 files, 772 tests), and `git diff --check`.

markusylisiurunen:
Simplified model-system-notice handling in `7e536f7`.

Main and subagent runtimes now prepend the active model's notice once when committing user/steering input. The persisted block is ordinary history from then on, including for compaction. Ephemeral runtimes still never resolve notices, and maintenance calls do not add fresh ones. Removed the dynamic model-context mapper, turn-level notice capture, and notice-specific context epoch behavior.

Verification remains green: `npm run check`, `npm run build`, `npm test` (72 files, 772 tests), and `git diff --check`.

markusylisiurunen:
Addressed the accepted findings from the latest review in `7da8b4a`:

- recover and persist committed tool calls that lack durable results, including explicit unknown-completion results, a canonical recovery message, cancelled tool projection, usage-checkpoint invalidation, and idempotent restoration
- reject empty and whitespace-only record, submit, queue, and steering input at the protocol boundary
- validate full reconstructed snapshots after every applied delta, including keyed-record fast paths
- serialize reasoning changes with runtime event projection so emitted revisions remain contiguous
- reject steering while interrupted work is still unwinding
- honor subagent termination while its startup event is awaiting delivery
- publish inherited usage immediately for forked ephemeral threads
- remove the unused ephemeral model resolver dependency and the remaining obsolete pruning reference

The accepted edge cases remain unchanged: an indivisible assistant/tool suffix may exceed the compaction retention target, nested subagents remain unsupported, and tool argument validation remains owned by bound tools.

Verification is clean:

- `npm run check`
- `npm run build`
- `npm test` (72 files, 776 tests)
- `git diff --check`

markusylisiurunen:
Follow-up `3cde41b` removes whole-snapshot validation from reference-neutral delta paths.

Snapshot parsing now runs only when a change can invalidate identities or cross-record references. Scalar updates and ordinary tool, facet, agent, and timeline activity retain structural sharing without traversing message history. Assistant content appends continue using their existing specialized path. Reference changes still receive full snapshot validation, including usage-checkpoint invalidation and tool/message relationship checks.

This reduces both latency and energy use by avoiding repeated full-history traversal, Zod allocations, and resulting GC work during high-frequency activity. In the same synthetic keyed-facet benchmark:

- 1,000 messages: about 1.05 ms to 0.007 ms per delta
- 5,000 messages: about 5.0 ms to 0.007 ms per delta

The optimized path is effectively independent of history size. Full verification remains clean: 72 test files, 776 tests, `npm run check`, `npm run build`, and `git diff --check`.

markusylisiurunen:
Merged current `main` (`0b2c0a3`) into this branch in `0ac0c75` and resolved the conflicts against the unified runtime architecture.

Main's intent, doubling the model-subturn limit, is preserved by changing `AgentRuntime`'s canonical default from 512 to 1024. The obsolete runner-level constant was removed rather than restoring duplicate ownership, the deleted legacy subagent test remains deleted, and the old core-session boundary test was adapted to `AgentRuntime`. The adapted test uses a two-subturn configured limit to verify the same final-subturn/off-by-one behavior without performing 1,024 model/tool cycles, plus a separate assertion covers the new 1024 default.

Verification is clean: `npm run check`, `npm run build`, `npm test` (72 files, 778 tests), and `git diff --check`.

markusylisiurunen:
Pushed test and legacy-surface cleanup in `b455f43` after a two-pass Sol review plus a manual coverage review.

The cleanup removes 601 net physical lines from the branch working tree (177 additions, 778 deletions). Branch-added test code fell from 2,763 to 2,428 lines, reducing its share of additions from 38.1% to 35.1%.

Removed or consolidated:

- duplicate runner-level coverage already exercised through `AgentRuntime`
- repetitive provider-failure recovery cases into one parameterized regression
- low-value property, exact revision, setter, argument-permutation, and transport-fixture assertions
- duplicated steering cancellation and transport-specific steering fixture machinery
- repeated async/model-stream setup through focused local helpers

I restored compact checks where the automated pass cut too far: all three wrappers constructing `AgentRuntime`, the new 1,024-subturn default, whitespace rejection for the problematic queue path and independent steer schema, and supervisor admission mapping to semantic `blocked`.

The production sweep also removed the unused legacy `ModeAdapter` and dead `SessionChatApp` hooks, moved runner-only event shapes out of the semantic agent event surface, closed unused type exports, and typed tool activity end-to-end instead of casting from `unknown`.

Verification is clean: `npm run check`, `npm run build`, `npm test` (72 files, 769 tests), and `git diff --check`.

markusylisiurunen:
bug observed while testing this branch:

- a subagent was launched with a long-horizon task
- it fills its first context window, auto-compact most likely triggered
- waiting for that subagent returned `session snapshot is invalid`, I suspect this happened after auto-compact

the `wait_for_agents` showed this:

```
**931b4717-36c6-4729-846f-fa3e64f15693**
Error: session snapshot is invalid
```

PR #363, opened 2026-08-01
expose canonical subagent state

### LINKED ISSUE #348: subagent errors are hard to interpret
they return something like `Error: Sub-agent produced an empty response (stopReason: error)`.

### LINKED ISSUE #350: expose more info about subagents to the main agent
for example, the cost or context window use could be important when deciding whether to spawn new subagents or sending further input

markustherobot:
## why

The main agent could start and wait for subagents, but it could not rediscover completed threads or inspect the runtime, usage, cost, context pressure, progress, and terminal cause needed to make good orchestration decisions. The existing state also mixed reusable thread identity with latest-run fields and flattened provider failures into misleading empty-response errors.

## what

This makes `AgentSupervisor` the owner of one canonical thread snapshot with a nested latest run, actual model and reasoning, working directory, cumulative accounting, current context occupancy, retained latest response, structured failure cause, and supervisor capacity.

It adds a dense text `list_agents` tool and uses the same formatter for spawn, follow-up, wait, interrupt, and compaction continuation state. `wait_for_agents` is explicitly non-consuming and returns current metadata for both completed and still-running requested agents. The available-subagents prompt now advertises each type's effective default runtime and context window.

The change also replaces `terminate_agent` with `interrupt_agent`, including the SDK and `session.interruptSubagent` protocol operation, because interruption preserves the reusable thread. Protocol version 5 carries the canonical agent state directly through host and TUI projections. Structured run failures now distinguish interruption, auto-compaction failure, model-subturn limits, provider errors with stop reason, and runtime errors, while a genuinely empty successful model response remains successful.

## details

Response retention is latest-run only. Starting a follow-up replaces the prior response instead of introducing run history or run selectors. `list_agents` reports response availability without embedding response bodies; repeated `wait_for_agents` calls read the retained body.

The storage document advances to version 2 and drops legacy agent presentation during migration. Normal recovery also removes persisted agents and agent-owned facets, then rewrites the normalized snapshot, because subagent runtimes are not recoverable across process restart and any retained thread ids would be stale and unusable.

This intentionally does not add automatic parent notifications, response injection, run-history APIs, or budget controls. There are no remaining implementation questions.

fixes #350
fixes #348


src/store/session_snapshot_migrations.ts
```diff
@@ -99,6 +100,16 @@ function migrateStoredSessionV0ToV1(value: unknown): unknown {
   return snapshot;
 }
 
+function migrateStoredSessionV1ToV2(value: unknown): unknown {
+  if (!isRecord(value)) {
+    throw new Error("stored session snapshot must be an object");
+  }
+
+  const snapshot = structuredClone(value);
+  snapshot.agents = {};
```
markustherobot: **[P1] Clear agent state on every session recovery**

This migration only removes agents from version-1 documents. Once Tau writes a version-2 snapshot containing an agent and restarts again, no migration runs, so `restoreProtocolState()` restores that record while the new `AgentSupervisor` has no corresponding runtime. The TUI can then show an agent whose id fails `interruptSubagent`, while `list_agents` sees nothing. Clear `recovered.agents` (and any agent-owned facets) during every recovery and persist that normalized snapshot, rather than doing this only in the one-time format migration.

markustherobot (commented):
The canonical supervisor, tool, protocol, and TUI shapes align, but the restart boundary still leaves persisted agent state diverging from the live supervisor. The recovery path needs to discard unrecoverable current-format agent records as well as legacy ones.

markusylisiurunen:
@markustherobot review

src/store/session_snapshot_migrations.ts
markustherobot: **[P1] Clear agent state on every recovery**

This only clears agents while migrating a version-1 document. Once this version writes a version-2 snapshot containing a child and Tau restarts again, no migration runs: `restoreProtocolState()` restores the record while the fresh `AgentSupervisor` has no matching runtime. The TUI then shows a phantom agent whose id fails `list_agents`, follow-up, wait, and interruption operations. Clear agents (and agent-owned facets) in the normal recovery normalization path and persist that normalized snapshot, rather than limiting cleanup to this one-time migration.

src/core/subagents/format.ts
```diff
@@ -0,0 +1,68 @@
...
+  const lines = [
+    `**${state.id}** · ${state.title}`,
+    `${state.name} · ${state.availability} · ${state.model.provider}/${state.model.id} · reasoning ${state.model.reasoning}`,
+    `${state.workingDirectory} · run ${state.run.revision} ${state.run.status}`,
+    `cost ${formatCost(state.costTotal)} · turns ${state.turns} · tools ${state.toolCalls}`,
+    formatUsage(state),
+  ];
+
+  if (state.run.progress) {
+    lines.push(`progress: ${state.run.progress}`);
```
markustherobot: **[P2] Keep retained responses out of progress**

`assistant_final` stores the first line of the child response in `run.progress`, and the succeeded snapshot preserves it. As a result, `list_agents` calls this formatter with `includeResponses: false` but still embeds the response's first line here (the entire body for a one-line response), defeating the advertised response-availability-only contract and potentially producing very large list/compaction output. Clear response-derived progress on completion, or omit and bound progress for terminal runs.

src/core/subagents/agent_supervisor.ts
```diff
@@ -266,50 +274,118 @@ export class AgentSupervisor {
...
       })
       .catch((error) => {
-        const aborted = record.abortRequested;
-        record.status = aborted ? "aborted" : "error";
-        record.error = aborted ? undefined : error instanceof Error ? error.message : String(error);
+        if (record.run.interruptRequested) {
+          this.finishInterrupted(record);
+        } else {
+          this.finishFailed(record, {
+            kind: "runtime-error",
```
markustherobot, line 313: **[P2] Preserve thrown provider failure classification**

After retries are exhausted, a provider stream or iterator rejection bubbles out of `AgentRuntime` through this catch; the runtime even records it via `noteProviderError`, but this branch labels every thrown failure as `runtime-error`. Common transport/provider outages therefore lose the new `provider-error` classification, while only terminal assistant messages with `stopReason: "error"` receive it. Propagate a typed provider failure from the runtime or otherwise classify exhausted model-stream errors separately.

docs/rpc.md
```diff
@@ -183,7 +183,7 @@ params (required):
 
 ```json
 {
-  "version": 4,
+  "version": 5,
```
markustherobot, line 186: **[P2] Update both protocol version fields**

The response envelope is now version 5, but this same initialize example still returns `"protocolVersion": 4` below, while the implementation validates and emits 5. Update the nested field too so the documented handshake is internally consistent.

markustherobot (commented):
The thread-state design is coherent, but the restart, response-boundary, failure-taxonomy, and protocol-documentation paths still have correctness gaps. I launched four Luna subreviews covering supervisor lifecycle, protocol/recovery, agent tools, and TUI/SDK projection; all four completed successfully.

markusylisiurunen:
@markustherobot fix

src/store/session_snapshot_migrations.ts
markustherobot: Resolved by normalizing every recovered snapshot: persisted agents and agent-owned facets are removed, and any changed recovery is committed before observe returns. Current-format restart coverage verifies the normalized snapshot is persisted.

markustherobot: Resolved in the normal recovery path, not only migration. Recovery now clears all persisted agent records and agent-owned facets, rewrites the normalized snapshot immediately, and has a regression test for version-2 state.

src/core/subagents/format.ts
markustherobot: Resolved by clearing progress whenever a run enters a terminal state. The retained response now exists only in the response field, so response-free listing and compaction output cannot leak its first line.

src/core/subagents/agent_supervisor.ts
markustherobot: Resolved by marking model-stream failures at the AgentRuntime boundary while preserving the original thrown error identity. AgentSupervisor recognizes that provenance and records provider-error with stop reason error; regression coverage exercises iterator rejection.

docs/rpc.md
markustherobot: Updated the nested initialize result protocolVersion to 5, matching the response envelope and implementation.

PR #364, opened 2026-08-01
archive automatic compaction context

### LINKED ISSUE #361: should auto-compact store the previous session in a temp file so it can be referred back to later?

markustherobot:
## why

automatic compaction is intentionally lossy, so an omitted detail could previously disappear from the agent's usable context with no fallback. preserving a temporary pre-compaction transcript gives the agent a bounded escape hatch without changing the summary-first behavior or making archives durable session state.

## what

Tau now best-effort archives the pre-compaction conversation after automatic summarization succeeds and before live history is replaced. assistant thinking is excluded. each agent gets an isolated directory under the execution environment's OS temp directory, with ordered `.txt` and `.json` snapshot pairs that continue numbering across runtime recovery while the directory survives.

The JSON record is a versioned normalized conversation DTO with untruncated archived content. The text projection is optimized for lookup and middle-truncates tool results. Successful archive paths are added to the hidden continuation message, which keeps the summary and retained context as the default and guides targeted lookup across current or older snapshot pairs. It also suggests delegating precise archive searches to a low-effort subagent when that would preserve context. Archive failures remain non-fatal and inject no stale hint.

The archive callback is bound for main sessions, background subagents, and ephemeral threads through their execution backend. Manual compaction remains unchanged.

## details

The temp directory name is derived from a hash of the agent id, directories use mode `0700`, and files use mode `0600`. Complete pairs are immutable, partial writes are ignored and cleaned before choosing the next sequence, and a missing directory is recreated from sequence one. This intentionally relies on best-effort temp-directory lifetime rather than persisting archive paths in agent state.

No remaining implementation questions.

fixes #361


src/core/session/auto_compaction_archive.ts
```diff
@@ -0,0 +1,341 @@
...
+
+function formatArchiveContent(content: readonly ArchiveContent[]): string {
+  return (
+    content
+      .map((block) => {
+        if (block.type === "text") return block.text;
+        if (block.type === "image") {
+          return `[Image mimeType=${JSON.stringify(block.mimeType)} data omitted from text transcript]`;
+        }
+        if (block.type === "thinking") {
```
markusylisiurunen: no need to include thinking in either text or json

src/core/session/compaction.ts
```diff
@@ -526,6 +527,7 @@ export function buildAutoCompactionPrompt(preparation: AutoCompactionPreparation
 export function buildAutoCompactionContinuationMessage(args: {
   cutType: AutoCompactionCutType;
   now: number;
+  archive?: AutoCompactionArchivePaths;
```
markusylisiurunen: why optional? isn't this once again against the deslop skill? same applies everywhere

src/core/session/compaction.ts
```diff
@@ -539,6 +541,17 @@ export function buildAutoCompactionContinuationMessage(args: {
...
   }
 
+  if (args.archive) {
+    lines.push(
+      "A temporary snapshot of the complete model-visible context immediately before this automatic compaction, including context retained above, is available in these files:",
+      `- text transcript: ${args.archive.textPath}`,
+      `- full JSON: ${args.archive.jsonPath}`,
+      "Prefer the text transcript for normal lookup; its tool results are middle-truncated.",
+      "The JSON contains full captured content and may be very large. Query it narrowly with jq, Node, rg, or bounded reads instead of dumping the whole file.",
+      "Earlier numbered pairs in the same directory are older pre-compaction snapshots. These files are temporary and may no longer exist.",
```
markusylisiurunen: I feel like this could be written a but more clear. it's completely intended flow for the agent to inspect this previous but also older compactions if needed. them being in the same dir in intentional and by design.

src/core/session/compaction.ts
```diff
@@ -539,6 +541,17 @@ export function buildAutoCompactionContinuationMessage(args: {
     );
   }
 
+  if (args.archive) {
+    lines.push(
+      "A temporary snapshot of the complete model-visible context immediately before this automatic compaction, including context retained above, is available in these files:",
```
markusylisiurunen: could explain a bit that the compaction summary itself should be enough usually but these transcripts are provided as a backup if needed. they should not be inspected just for the sake of it. also, sometimes it's much more token efficient to launch a low effort subagent to answer a specific question or to find exavt information rather than reading these very large files directly (it may take a lot of context again and auto-compaction is meant to prevent that).

phrase this well yourself, do not use my words direvtly

markusylisiurunen:
@markustherobot fix

src/core/session/auto_compaction_archive.ts
markustherobot: Removed assistant thinking during archive normalization, so it reaches neither the JSON record nor the text projection. The archive test now asserts the reasoning text is absent from both files.

src/core/session/compaction.ts
markustherobot: Made `archive` a required field with the explicit `AutoCompactionArchivePaths | undefined` domain state. Runtime and test callers now always provide it instead of relying on an optional object shape.

src/core/session/compaction.ts
markustherobot: Reworded the continuation to make the shared numbered directory intentional: the current pair is identified directly, and targeted searches should include older pairs when the missing detail may predate the latest compaction.

src/core/session/compaction.ts
markustherobot: Reframed the archives as a fallback for a specific missing detail, with the summary and retained context remaining the normal path. The guidance now favors narrow reads and suggests a low-effort subagent for precise lookup when available.

PR #366, opened 2026-08-01
correct GPT-5.6 Terra and Luna pricing

markusylisiurunen:
## why

Tau currently uses `@earendil-works/pi-ai@0.83.0`, whose bundled GPT-5.6 Terra and Luna metadata predates OpenAI's latest price reductions. Until Pi publishes and Tau adopts a release containing the correction, Tau reports stale estimated costs for the first-party OpenAI API and Codex endpoints.

## what

Add guarded Tau-side pricing overrides for GPT-5.6 Terra and Luna on the `openai` and `openai-codex` providers. The override applies only when Pi exposes the exact stale standard and long-context prices, so a future dependency update with corrected or otherwise changed metadata automatically takes precedence.

Keep passthrough catalogs untouched and preserve the existing Codex-only GPT-5.6 context-window override. Catalog coverage verifies both first-party providers, long-context pricing, passthrough exclusion, and the exact-metadata guard.

### DIFF
--- a/src/core/models/tau_model_overrides.ts
+++ b/src/core/models/tau_model_overrides.ts
@@ -1,15 +1,110 @@
 import type { Api, Model } from "@earendil-works/pi-ai";
 
 const GPT_5_6_CODEX_MODEL_IDS = new Set(["gpt-5.6-sol", "gpt-5.6-terra", "gpt-5.6-luna"]);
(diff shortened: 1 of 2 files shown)


markusylisiurunen:
this is unnecessary, can just wait for Pi to update

PR #368, opened 2026-08-02
recover provider stream failures

### LINKED ISSUE #365: keep Telegram sessions recoverable after steering provider failures
## problem

A provider stream exception during a steering continuation is surfaced as a `session.steer` protocol `internalError`. The Telegram session manager treats that rejected SDK call as a terminal session failure and closes the client, even though the committed conversation state remains recoverable.

This happened in a real Telegram session immediately after a successful tool call. The next steering user message was durably committed, the provider failed before emitting an assistant partial, and the persisted snapshot ended with that unanswered user message while the Telegram session became failed.

## current behavior

- `AgentRuntime` rethrows a model-stream exception before tool admission.
- The steering submission result rejects.
- `SessionProtocolHandler` returns `internalError: steering turn failed`.
- `TelegramSessionManager` handles the rejected SDK submission as fatal, marks the session failed, and closes the client.
- The underlying provider error is not retained in the durable transcript.

## expected behavior

Exhausted provider-stream failures should produce a canonical failed turn that remains recoverable, including during steering continuations. The failure should be durable and visible, and the user should be able to submit another message. Event-sink, persistence, transport, and invariant failures should remain terminal.

## relevant code

- `src/core/agent/agent_runtime.ts`
- `src/core/telegram/session_manager.ts`
- `test/agent_runtime.test.js`
- `test/telegram_session_manager.test.js`

The regression coverage should exercise a provider exception after a steering message has been applied, not only successful steering and already-materialized failed turn results.

markustherobot:
## why

A provider stream exception could escape `AgentRuntime` after a steering message had already been committed. The rejected steering request caused Telegram to mark the whole session failed and close its SDK client, even though the persisted conversation remained usable.

## what

Provider failures from stream creation, iteration, and final result collection now become canonical failed assistant turns after retries are exhausted. The runtime persists the failure, emits a failed turn outcome, and resolves steering with `status: "failed"`, allowing Telegram and other clients to keep the session open and accept the next message.

The regression coverage exercises a retry-exhausted provider exception after steering has been applied, verifies the durable failed transcript and subsequent successful submission, and confirms failed steering results do not close the Telegram client. Host coverage verifies that provider failures publish and persist the same committed failed assistant state, while background subagents and ephemeral review threads reject failed partial output instead of reporting it as success.

## details

The model runner tags only errors originating at the provider stream boundary. Stream-processing invariants and event-sink failures remain rejecting terminal failures. This distinction is intentional and is the material implementation decision for the ambiguous exception boundary.

fixes #365


src/core/agent/agent_runtime.ts
```diff
@@ -1436,14 +1441,15 @@ export class AgentRuntime {
...
+          }
+          if (signal.aborted && admittedToolCalls.length === 0) {
             throw next.error;
           }
 
-          const errorMessage =
-            next.error instanceof Error ? next.error.message : String(next.error);
-          finalMessage = createToolRecoveryAssistantMessage({
+          const errorMessage = next.error.message;
+          finalMessage = createFailedAssistantMessage({
```
markustherobot, line 1452: **[P1] Do not report failed child runs as successful**

When a background subagent or ephemeral diff-review thread hits a provider exception after emitting partial text, this conversion now resolves `submit()` with an error `finalMessage` instead of rejecting. `AgentSupervisor.startRun()` and `HostedEphemeralAgentSession.submitMessage()` only check `aborted`, `blocked`, and `limitReached` before accepting non-empty assistant text, so they expose that incomplete failed partial as a successful child result. Update both one-shot consumers to handle `result.finalMessage?.stopReason === "error"` (preserving `errorMessage`) before extracting text, and cover a partial-output failure.

markustherobot (commented):
I found one downstream lifecycle regression. I launched three Luna subreviews covering provider-stream classification, AgentRuntime lifecycle and tool recovery, and host/protocol/Telegram behavior; I verified their evidence against the current head.

markusylisiurunen:
@markustherobot fix

src/core/agent/agent_runtime.ts
markustherobot: I updated both one-shot child consumers to reject assistant results with an error stop reason before extracting transcript text, preserving the provider error instead of treating partial output as success. Added partial-stream failure coverage for background subagents and ephemeral review threads.

markusylisiurunen:
@markustherobot fix conflicts with main

PR #369, opened 2026-08-02
improve automatic compaction

markusylisiurunen:
## why

Automatic compaction should leave a concise but useful continuation context without letting large tool results dominate the retained tail. Compaction summaries also benefit from more freedom to organize the handoff, while transient summary failures should not block a turn immediately.

The existing pre-compaction transcript archives provide a useful escape hatch for bulky exact details, but the summarizer did not know how the continuing assistant could recover those records.

## what

Automatic compaction now bounds retained textual tool and recovery results, retries summary generation once, and uses the active reasoning effort. The summary prompt keeps its continuity checklist while allowing project-specific organization, with dedicated guidance for split turns and an accurate description of bounded retained output.

Every archivable summarized record is labeled with its history entry id. The compaction model may cite those ids naturally for exact logs, payloads, errors, or code that can be recovered from the transcript archive, and the hidden continuation guidance explains how to resolve summary references and truncated retained output through the text and JSON archive pairs.


src/core/session/compaction.ts
```diff
@@ -465,6 +462,44 @@ function estimateTextTokens(text: string): number {
   return Math.max(1, bytesToTokens(Buffer.byteLength(text, "utf8")));
 }
 
+function boundRetainedEntry(entry: CompactionHistoryEntry): CompactionHistoryEntry {
+  const message = structuredClone(entry.message);
+  if (message.role !== "toolResult") {
```
markustherobot, line 496: **[P1] Bound retained tool-recovery payloads**

Failed model/tool execution stores completed results in a hidden `tool-recovery` user message, so this early return leaves those `<result-text>` payloads untouched. Because that hidden user message is also treated as a turn boundary, a large recovery result can become the retained suffix and bypass the new 8,192-token cap, leaving the next model call over its context limit. Apply equivalent result-text truncation to retained recovery messages while preserving the original in the archive.

src/core/session/compaction.ts
```diff
@@ -482,22 +517,22 @@ export function prepareAutoCompaction(
     .slice(latestCompaction.index + 1, cut.startIndex)
     .filter((entry) => !hasAutoCompactionContinuationMetadata(entry.message));
   const messagesToSummarize = entriesToSummarize.map((entry) => entry.message);
-  const userMessageIds = buildUserMessageIdMap(entriesToSummarize);
+  const historyEntryIds = buildHistoryEntryIdMap(entriesToSummarize);
```
markustherobot: **[P2] Label only records the archive contains**

This map labels assistant entries with `stopReason: "error"` or `"aborted"`, but `tryArchiveAutoCompaction()` writes `modelHistoryEntries`, which filters those entries out. During tool-recovery flows the summarizer can therefore cite a shown assistant history id that appears in neither the text nor JSON archive. Use the same source set for prompt labels and archival, or withhold citation ids for records that will not be archived.

src/core/session/compaction.ts
```diff
@@ -514,13 +549,28 @@ export function prepareAutoCompaction(
   };
 }
 
+const AUTO_COMPACTION_ARCHIVE_REFERENCE_GUIDANCE = `After compaction, the continuing assistant will receive paths to temporary text and JSON transcripts of the pre-compaction context. Every conversation record above is labeled with its history entry id. The continuing assistant can search the numbered text transcripts for that id, then inspect the corresponding JSON record when it needs the full archived content.
```
markustherobot: **[P2] Only permit citations after archiving succeeds**

This unconditionally promises archive paths, but `runAutoCompaction()` attempts the best-effort archive only after summary generation and swallows archive failures. If the summarizer offloads exact logs to a history id and that write fails, compaction commits a dead reference, potentially after the retained copy was truncated too. Archive before generating the summary and include this guidance only on success, or otherwise keep the summary contract independent of the archive.

markustherobot (commented):
I found one retained-context overflow path and several gaps in the new archive-recoverability contract. I launched 3 subreviews covering summary retry/model lifecycle, retained tool-result fidelity, and prompt/archive protocol coherence; all completed successfully.

src/core/session/compaction.ts
```diff
@@ -514,13 +549,28 @@ export function prepareAutoCompaction(
...
+  const retainedContextGuidance =
+    preparation.cutType === "split-turn"
+      ? `The retained context will begin in the middle of the latest assistant/tool turn. Add a "## Current Turn Handoff" section that clearly captures:
+- the original user request for this turn
+- work completed before the retained suffix
+- tool state and findings at the cut boundary
+- what the first retained message is continuing and what remains unresolved
+
+Place the section wherever it makes the handoff clearest. Preserve earlier session context elsewhere in the summary without duplicating the retained suffix.`
+      : "The retained context will include recent messages verbatim. Ensure the summary complements that retained context without duplicating unnecessary detail.";
```
markustherobot: **[P2] Stop calling the bounded suffix verbatim**

A turn-boundary suffix can now contain a tool result truncated by `boundRetainedEntry`, yet this tells the summarizer it will be retained verbatim and the continuation message repeats that claim. The summarizer may omit details it believes remain exact, while the continuing assistant is not told that recent output was bounded. Describe the truncation in both prompts and direct recovery of omitted details through the archive when available.

src/core/session/compaction.ts
markusylisiurunen: over-engineering. archive basically always succeeds. no action needed.

src/core/session/compaction.ts
markusylisiurunen: not sure what this means. if a real bug, should be fixed.

markusylisiurunen:
@markustherobot fix

src/core/session/compaction.ts
markustherobot: I now apply the 8,192-token per-result bound to `<result-text>` payloads in retained tool-recovery messages on a clone, so the archived source history remains complete. Regression coverage verifies both truncation and non-mutation.

src/core/session/compaction.ts
markustherobot: I now withhold citation ids from error and aborted assistant records, matching the `modelHistoryEntries` archive filter. Regression coverage exercises both stop reasons.

src/core/session/compaction.ts
markustherobot: I removed the verbatim claim from both handoff prompts. They now state that retained tool and recovery output may be bounded, and the continuation guidance directs targeted text-transcript and JSON recovery for omitted output.

PR #370, opened 2026-08-02
add persistent session goals

### LINKED ISSUE #367: a `/goal` implementation, similar to Codex

markustherobot:
## why

Tau needs a durable goal primitive that survives compaction and can recover when a model returns before explicitly completing or blocking long-running work. A slash command alone would not provide those guarantees because the objective, continuation lifecycle, and model controls must remain host-owned and visible through the session protocol.

## what

This adds a nullable two-state goal to the canonical session snapshot, host-owned goal lifecycle and continuation probing, and `get_goal`, `create_goal`, and `update_goal` model tools. `/goal <objective>`, bare `/goal`, `/goal resume`, and `/goal clear` expose the user controls through new SDK and session protocol methods.

Active goal policy is injected into the initial turn, fallback continuation turns, and compaction continuation context. Completion clears the goal, while model blocking, interruption, terminal failure, and recovery leave it blocked. Goal creation and its visible objective message are committed atomically, and stored sessions migrate to the required nullable goal field.

Logical-run cancellation begins before start/resume persistence, and steering received between physical goal turns is applied before autonomous continuation. Goal-controlled messages carry persisted metadata so stale active policy cannot be replayed through retry, and an invalid clear never interrupts unrelated work.

The protocol, SDK, persistence, TUI, and user documentation are updated together, with regression coverage for continuation, tool completion, objective revision, interruption, recovery, compaction context, protocol deltas, SDK routing, and migration.

## details

The issue left the compact status treatment open. I used a two-column editor-header indicator instead of adding footer text: `◆` means active and `◇` means blocked. Bare `/goal` remains the place to inspect the full objective and status.

Queued user work waits behind the continuous logical goal run, as agreed in the issue discussion. This first version intentionally has no budget, wall-clock accounting, or consecutive-continuation circuit breaker.

The session protocol version advances to 6 and the stored session document version advances to 3.

fixes #367


markustherobot (commented):
Reviewed the full goal lifecycle across host scheduling, model tools, protocol/storage/SDK, and TUI behavior. I found several lifecycle-control and durability inconsistencies that should be addressed before merge, plus smaller boundary-validation and protocol-documentation gaps. I launched four Luna xhigh subreviews covering host lifecycle, protocol/storage/SDK, tools/prompts/runtime, and TUI/docs/tests; all four completed successfully.

src/host/local_session_host.ts
```diff
@@ -648,36 +743,42 @@ class LocalHostedSessionHandle implements LocalHostedSession {
...
+        const outcome = turnOutcomeFromResult(result, lastAssistantMessage);
+        this.turnOutcomes.set(userMessage.id, outcome);
+        if (outcome.status !== "completed") {
+          await this.blockActiveGoal();
+        }
+        await this.emitSnapshotResetIfChanged("assistant-message");
+        if (outcome.status !== "completed" || this.goal?.status !== "active") {
+          return outcome;
+        }
+        await this.session.commitUserText(buildGoalContinuationText(this.goal));
```
markustherobot, line 907: **[P1] Cancel the logical run between continuations**

After `runtime.runTurn()` resolves, the runtime is idle while this code awaits snapshot persistence and commits the next hidden turn. If `session.interrupt`, `session.clearGoal`, or shutdown lands in that window, `interruptTurn()` returns false; clear/shutdown then wait on the active submit while this loop starts another model turn. With no circuit breaker, that can hang indefinitely. Add host-owned logical-run cancellation state, check it before committing or starting each continuation, and block the active goal when it is set.

src/host/local_session_host.ts
```diff
@@ -648,36 +743,42 @@ class LocalHostedSessionHandle implements LocalHostedSession {
...
-      await this.emitSnapshotResetIfChanged("assistant-message");
-      return outcome;
-    } catch (error) {
+
+      let lastAssistantMessage: AssistantMessage | undefined;
       try {
-        await this.cleanupFailedTurn();
-      } catch {
-        // Preserve the original turn failure for the protocol response.
+        const result = await this.runtime.runTurn();
```
markustherobot, line 876: **[P1] Honor the final steering subturn outcome**

`AgentRuntime.runTurn()` deliberately returns its initial result after processing steering continuations. If the initial subturn completed but a steering continuation aborts, fails, or blocks, this code still sees `completed` with an active goal and schedules another continuation. Escape during a steered goal can therefore appear to interrupt while autonomous work resumes. Have the goal loop consume the final physical-subturn outcome or an explicit terminal signal and block or stop accordingly.

src/host/local_session_host.ts
```diff
@@ -648,36 +743,42 @@ class LocalHostedSessionHandle implements LocalHostedSession {
...
+        lastAssistantMessage = result.finalMessage;
+        if (result.aborted && this.draftAssistantMessage) {
+          await this.interruptDraftAssistantMessage();
+        }
+        const outcome = turnOutcomeFromResult(result, lastAssistantMessage);
+        this.turnOutcomes.set(userMessage.id, outcome);
```
markustherobot: **[P1] Persist the logical goal run's terminal outcome**

Each iteration attaches its physical outcome to that iteration's user message. If the visible goal turn completes and a later hidden continuation aborts or fails, `startGoal()` returns the terminal outcome but its `userHistoryEntryId` still points to a message persisted as `completed`; after reconnect, `getTauSdkSessionTurnOutcome()` reports the wrong result. Track the root goal entry and update it with the final logical outcome, or expose a stable logical-run outcome identifier.

src/host/local_session_host.ts
```diff
@@ -1214,6 +1315,29 @@ class LocalHostedSessionHandle implements LocalHostedSession {
...
+    }
+    const normalized = objective.trim();
+    if (!normalized) {
+      throw new Error("goal objective must not be empty");
+    }
+    return { objective: normalized, status: "active" };
+  }
+
+  private async setGoal(goal: SessionProtocolGoal | null): Promise<void> {
+    this.goal = structuredClone(goal);
```
markustherobot, line 1483: **[P2] Roll back goal state when persistence fails**

This changes the authoritative in-memory goal before the awaited patch is durable. If the store write fails, create/update/clear rejects (and model tools report a blocked tool result), but `this.goal` retains the rejected mutation; a transient create failure can still drive continuations, while a completion failure can stop work although disk still says active. Restore the previous value on failure, or otherwise make the in-memory transition commit atomically with the patch.

src/core/tools/goal.ts
```diff
@@ -0,0 +1,115 @@
...
+      return {
+        goal,
+        instruction: buildGoalPolicy(goal),
+      };
+    }),
+    createGoalTool(UPDATE_GOAL_TOOL, updateArgsSchema, async (update) => {
+      const goal = await manager.updateGoal(update);
+      return goal === null
+        ? { goal: null, completed: true }
+        : { goal, instruction: buildGoalPolicy(goal) };
```
markustherobot: **[P2] Do not return active policy for blocked goals**

For `update_goal({ status: "blocked" })`, the manager returns a blocked goal here, but `buildGoalPolicy()` says an active goal is in effect and instructs the model to continue. Because the runtime performs a follow-up model subturn after the tool result, this directly contradicts the persisted state and can keep work going after an explicit block. Return blocked-specific guidance or omit the active policy, and ensure blocked transitions supersede prior active guidance.

src/core/tools/catalog.ts
```diff
@@ -76,7 +87,10 @@ export const ToolCatalog = {
       tools.push(createNookToolDefinition(options.backend, options.config));
     }
     const enabledToolNames = new Set<string>(options.persona.tools);
-    return new ToolRegistry(tools.filter((tool) => enabledToolNames.has(tool.schema.name)));
+    return new ToolRegistry([
+      ...tools.filter((tool) => enabledToolNames.has(tool.schema.name)),
+      ...createGoalToolDefinitions(options.goalManager),
```
markustherobot, line 92: **[P2] Reserve the new host tool names**

`ClientToolBroker.HOST_TOOL_NAMES` does not include `get_goal`, `create_goal`, or `update_goal`. A client advertising one is accepted during initialization, but the first turn merges it with these definitions and `ToolRegistry` throws a duplicate-tool error after the user message has already been recorded. Reserve these names at the client registration boundary and cover the collision path.

src/core/commands/registry.ts
```diff
@@ -249,6 +254,29 @@ export function createCommandRegistry(): CommandRegistry<CommandDispatchContext>
     run: (ctx, command) => ctx.diff(command.argsText),
   });
 
+  registry.register({
+    id: "goal",
+    usage: "/goal [objective|resume|clear]",
+    description: "show, start, resume, or clear a session goal",
+    autocompleteDescription: "manage the persistent session goal",
+    argument: "none",
```
markustherobot, line 262: **[P2] Keep goal inspection and clearing available while running**

This command keeps the default `allowDuringStreaming: false`, so the controller rejects both bare `/goal` and `/goal clear` while an autonomous goal is active. An intentionally unbounded goal therefore cannot be inspected or cleared directly from the TUI, even though `session.clearGoal` is designed to interrupt active work. Add action-specific streaming permission so show/clear dispatch while start/resume remain gated.

docs/rpc.md
```diff
@@ -183,7 +186,7 @@ params (required):
 
 ```json
 {
-  "version": 5,
+  "version": 6,
```
markustherobot, line 189: **[P2] Finish updating the v6 protocol contract**

The envelope is updated to v6 here, but the same initialize example still reports `"protocolVersion": 5`. The current-reason lists in this file and `docs/session-event-protocol-redesign.md` also omit `goal`, and the redesign snapshot shape omits the new required nullable `goal` field. Update the nested version and the documented reason/snapshot unions so clients following the canonical contract do not reject valid responses or construct invalid snapshots.

markusylisiurunen:
@markustherobot fix the findings but don't just blindly trust them, analyze whether they are real and realistic, and whether it's clear how to resolve them

src/host/local_session_host.ts
markustherobot: I added host-owned logical-turn cancellation that is checked before every continuation commit/start. Interrupt, clear, and shutdown now stop the run even while the runtime is idle, and the active goal is durably blocked.

src/host/local_session_host.ts
markustherobot: I kept the initial subturn result semantics and added an explicit terminal subturn result from AgentRuntime. The goal loop now uses that terminal result for continuation and blocking decisions, including aborted or failed steering continuations.

src/host/local_session_host.ts
markustherobot: I now track the root goal history entry across continuations and overwrite its persisted turn outcome with the logical run’s terminal outcome. Recovered SDK lookups therefore agree with the startGoal response.

src/host/local_session_host.ts
markustherobot: I made setGoal restore the previous in-memory value when the awaited patch commit fails. Regression coverage exercises failed create, update, and completion/clear transitions.

src/core/tools/goal.ts
markustherobot: I added blocked-specific tool-result guidance that explicitly stops autonomous work and supersedes the earlier active policy. Active goals still receive the full continuation policy.

src/core/tools/catalog.ts
markustherobot: I reserved get_goal, create_goal, and update_goal at client registration and added all three to the collision test matrix.

src/core/commands/registry.ts
markustherobot: I made streaming permission action-specific: bare /goal and /goal clear dispatch while active, while start and resume remain gated until idle. The TUI test now covers both sides.

docs/rpc.md
markustherobot: I completed the v6 documentation updates: the initialize result now reports 6, the snapshot examples/types include required nullable goal state, goal.set is documented, and goal is present in the reason unions.

markusylisiurunen:
@markustherobot review

src/host/local_session_host.ts
```diff
@@ -625,6 +638,89 @@ class LocalHostedSessionHandle implements LocalHostedSession {
...
+    await this.setGoal(goal);
+    return structuredClone(goal);
+  }
+
+  async startGoal(objective: string): Promise<SessionProtocolStartGoalResult> {
+    const goal = this.buildNewGoal(objective);
+    this.goal = goal;
+    this.pendingGoalCommit = goal;
+    try {
+      const { userHistoryEntryId } = await this.record({
```
markustherobot: **[P1] Make goal startup cancellable before recording**

`activeLogicalTurn` is not installed until `runTurn()`, after both this record and the analogous resume setup finish. While either persistence step is pending, `session.interrupt` reports no interruption, and clear/shutdown wait on the active submit without having cancelled it; a slow or blocked store can therefore make the goal uninterruptible and let it start after Escape. Create the logical cancellation state before the first start/resume await and check it before entering model work.

src/host/local_session_host.ts
```diff
@@ -641,53 +737,93 @@ class LocalHostedSessionHandle implements LocalHostedSession {
...
+          await this.blockActiveGoal();
+        }
+        await this.emitSnapshotResetIfChanged("assistant-message");
+        if (logicalTurn.cancellationRequested) {
+          return await this.cancelLogicalTurn(rootUserMessage.id);
+        }
+        if (terminalOutcome.status !== "completed" || this.goal?.status !== "active") {
+          return goalRootHistoryEntryId ? terminalOutcome : initialOutcome;
+        }
+        await this.session.commitUserText(buildGoalContinuationText(this.goal));
```
markustherobot, line 907: **[P1] Preserve steering across continuation boundaries**

Between a physical turn ending and this continuation being committed, the runtime is idle although the logical goal submit is still active. `handleSteer()` only calls `session.steer()` when `isTurnRunning` is true, so steering received in this window is put in `pendingQueuedSubmits` and runs only after the potentially unbounded goal settles, rather than changing the next continuation. Add a logical-run steering buffer/path so Ctrl+Enter at this boundary is applied before autonomous work continues.

src/host/session_protocol_handler.ts
```diff
@@ -1184,6 +1194,103 @@ export class SessionProtocolHandler {
...
+          { cause: error instanceof Error ? error.message : String(error) },
+        ),
+      );
+    }
+  }
+
+  private async handleClearGoal(
+    request: Extract<SessionProtocolRequestMessage, { method: "session.clearGoal" }>,
+  ): Promise<void> {
+    await this.withSessionMutation(request, "session goal cleared", async (state) => {
```
markustherobot: **[P1] Validate a goal before interrupting for clear**

`withSessionMutation()` interrupts the active submit and rejects queued work before this callback calls `clearGoal()`. Because `/goal clear` is allowed during any streaming turn, issuing it while an ordinary turn has no goal aborts unrelated work and then returns `no goal exists`. Check goal existence before the interrupting mutation (and revalidate inside it), so an invalid clear cannot destroy active or queued work.

src/core/session/goal.ts
```diff
@@ -0,0 +1,45 @@
...
+
+Explain the blocker in the current response. Do not continue work on this goal unless the user explicitly resumes it.`;
+}
+
+export function prependGoalPolicy(text: string, goal: SessionProtocolGoal): string {
+  return prependTauHiddenSystemMessages(text, [buildGoalPolicy(goal)]);
+}
+
+export function buildGoalContinuationText(goal: SessionProtocolGoal): string {
+  return prependTauHiddenSystemMessages("", [buildGoalPolicy(goal), GOAL_CONTINUATION_MESSAGE]);
```
markustherobot: **[P2] Reconcile goal state before retrying a goal turn**

Goal start and continuation messages persist this active policy, but interruption/failure changes the snapshot goal to `blocked` and completion/clear changes it to `null`. `session.retry` runs the latest user history entry again without adding a new turn, so retrying one of these messages gives the model an active-goal instruction while the authoritative goal is no longer active, bypassing `/goal resume` and potentially re-executing old work or updating a nonexistent goal. Mark goal-controlled turns and reject or rewrite retry according to the current goal state.

markustherobot (commented):
Reviewed the current head across host scheduling, runtime/tools, protocol/storage/SDK, and TUI behavior. I found remaining lifecycle-control holes around startup and continuation boundaries, plus goal-state validation and retry inconsistencies. I launched four Luna xhigh subreviews covering host lifecycle/persistence, runtime/tools/prompts, protocol/storage/SDK, and TUI/docs/tests; all four completed successfully.

markusylisiurunen:
@markustherobot do another round of fixes, still staying critical like before

src/host/local_session_host.ts
markustherobot: I moved logical-turn ownership to cover the entire start/resume operation, before the first persistence await. Interrupt and shutdown now mark cancellation during setup, the setup path checks it before model work, and regression tests cover both objective-message persistence and resume activation.

src/host/local_session_host.ts
markustherobot: I added a host-owned steering buffer for active goal boundaries. The protocol now recognizes that boundary as steerable, persists buffered steering with the current authoritative goal policy, and applies it before another autonomous continuation starts.

src/host/session_protocol_handler.ts
markustherobot: I now validate goal existence before entering the interrupting mutation and revalidate inside the serialized mutation. A clear with no goal returns invalid_request without interrupting active work or cancelling queued messages.

src/core/session/goal.ts
markustherobot: I marked goal starts, continuations, and goal steering with persisted goal-turn metadata and added a dedicated retry path that rejects those turns with invalid_request. Blocked goals must resume through session.resumeGoal, so stale active policy is never replayed.

PR #372, opened 2026-08-02
serialize repeated steering projections

### LINKED ISSUE #371: serialize repeated steering outcome projection
## problem

Repeated steering across successive turn boundaries can corrupt the hosted session snapshot and fail both the steering request and the original active submit.

This is deterministic when a second steering message arrives while the first steering continuation is running. When the first continuation finishes, its result callback performs a full snapshot rebuild outside the runtime event queue. The runtime may already have appended the second steering user message, even though that message's awaited `user_message` event has not yet been projected. The rebuilt snapshot therefore contains the second message early, and the later event attempts to append it again.

Snapshot validation then fails with an error shaped like:

```text
session delta produced an invalid snapshot: messages: duplicate message id '...'; timeline: duplicate timeline item id '...'
```

The shared runtime rejects the pending steering submission and rethrows the event-sink failure. The protocol surfaces two generic errors:

```text
steering failed: steering turn failed
session turn failed: failed to run session turn
```

The protocol responses contain the underlying exception in `error.data.cause`, but the TUI currently renders only `error.message`, hiding the actionable diagnostic.

## expected behavior

- Successive steering continuations must remain ordered through the awaited runtime projection boundary.
- A steering result must not rebuild protocol state from runtime history that is ahead of semantic event projection.
- Each steering user message and timeline item must be projected exactly once.
- A failed queued-message commit must not leave later queued requests pending forever.
- Failure of one pending-message observer must not abort shared session work.
- TUI submit, queue, and steering errors should include the protocol `data.cause` when present.

## relevant code

- `src/host/local_session_host.ts`
  - `steer()` steering application/result mapping
  - runtime event and snapshot serialization
  - pending queue drains
  - pending-message observer fanout
- `src/tui/session_chat_controller.ts`
  - submit, queue, and steering error rendering
- `test/local_session_host.test.js`
- `test/rpc_server.test.js`
- `test/session_chat_controller.test.js`

## regression coverage

Add a real hosted-session test with three controlled model subturns:

1. Start the original turn.
2. Apply the first steering message and start its continuation.
3. Submit a second steering message during that continuation.
4. Finish the first continuation and verify the second steering message is projected once.
5. Finish the second continuation and verify both steering requests and the original submit settle successfully with valid persisted outcomes.

Also cover failing and clearing the remaining pending queue after one queued message cannot be committed, without starting another agent turn, plus isolation from a throwing pending-state observer.

markusylisiurunen:
## why

Successive steering messages could race protocol snapshot projection. Finishing one steering continuation rebuilt a full snapshot from runtime history after the next steering message had been appended but before its semantic event was projected. The later event then attempted to append the same message again, invalidating the snapshot and failing both the steering request and the original active submit.

Related pending-message failures were also opaque or could disrupt later work: the TUI hid the protocol cause, one failed queued commit stopped the remaining queue from draining, and a throwing observer could abort shared pending-state publication.

## what

Serialize steering outcome projection through the runtime event queue and persist it as an exact user-message replacement instead of rebuilding the full mutable runtime snapshot. Steering application now relies on the already-awaited durable user event rather than issuing a redundant snapshot write.

A queued commit failure now fails and clears every remaining queued request without starting another agent turn. Pending-message observer delivery is isolated from shared session state, and TUI submit, queue, and steering errors include the protocol-provided cause.

## details

Regression coverage reproduces the original failure with three controlled hosted-session subturns and two successive steering boundaries. Additional coverage verifies that a failed queued commit clears the remaining batch without running it, plus observer isolation and actionable TUI diagnostics.

fixes #371

### DIFF
--- a/src/host/local_session_host.ts
+++ b/src/host/local_session_host.ts
@@ -730,14 +730,12 @@ class LocalHostedSessionHandle implements LocalHostedSession {
     const submission = this.runtime.steer(text);
     return {
       id: submission.id,
-      applied: submission.applied.then(async (association) => {
-        await this.commitSnapshot();
-        return { userHistoryEntryId: association.historyEntryId };
-      }),
+      applied: submission.applied.then((association) => ({
+        userHistoryEntryId: association.historyEntryId,
+      })),
       result: submission.result.then(async (association) => {
         const turn = turnOutcomeFromResult(association.result, association.result.finalMessage);
-        this.turnOutcomes.set(association.historyEntryId, turn);
-        await this.emitSnapshotResetIfChanged("assistant-message");
+        await this.commitSteeringTurnOutcome(association.historyEntryId, turn);
         return {
           userHistoryEntryId: association.historyEntryId,
           turn,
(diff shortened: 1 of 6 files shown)


markusylisiurunen:
@markustherobot review

PR #374, opened 2026-08-02
add TUI input prefill tool

### LINKED ISSUE #373: a tool for the model to prefill the TUI input
should probably be a client tool and tied to TUI only. does this break multiple TUIs attached to the same session? the point is to allow the model to prefill the input e.g. when it needs answers to multiple questions, to speed up user's work.

markustherobot:
## why

The model needs a TUI-local way to prepare structured draft input for the user without submitting it or overwriting text the user is already composing. Multiple TUIs can observe one hosted session, so client-tool advertisement also needs an explicit opt-out for secondary clients.

## what

Added the `prefill_input` TUI client tool with one required, non-blank `text` argument. It fills an empty editor with a draft the user can review, edit, and submit, and returns a tool error without changing the editor when any draft text already exists. Its model guidance reserves the tool for structured drafts that meaningfully save user effort and asks it to use simple placeholders for details the user needs to supply.

Added `--no-client-tools` to local `tau` and `tau attach`. The flag disables both `diff_review` and `prefill_input` through the shared TUI client-tool factory, and applies to future tools added there as well. CLI help, user documentation, and regression coverage were updated for the new behavior.

## details

The existing duplicate-name rejection remains the ownership rule. If two client-tool-enabled TUIs observe the same session, the second attachment still fails; additional TUIs should use `--no-client-tools`. No implicit routing or ownership selection was added.

The prefill operation checks and updates the editor synchronously in the TUI controller, so a failed attempt preserves the existing draft unchanged. There are no remaining implementation questions.

fixes #373


src/tui/session_chat_app.ts
```diff
@@ -47,9 +52,20 @@ export type SessionChatTransportOptions = Omit<SessionChatAppOptions, "client" |
       }
   );
 
+const prefillInputArgsSchema = z
+  .object({
+    text: z.string().min(1),
```
markustherobot: **[P2] Reject blank prefill drafts**

`min(1)` accepts whitespace-only values such as `"   "` or `"\n"`. The tool then reports success and fills the editor, but `handleSubmit()` trims that value and treats it as an empty submission (or an empty-submit retry gesture), so the required non-empty draft contract is violated. Use a trim-aware refinement that preserves the original text, mirror it in the TypeBox schema with a non-whitespace constraint, and cover blank input.

markustherobot (commented):
The overall wiring and client-tool ownership look sound, but one argument-validation edge remains. I launched three subreviews covering client-tool protocol and ownership, editor behavior, and CLI and transport wiring; all completed.

markusylisiurunen:
@markustherobot fix the finding. I also think some small refinement to the tool description would be good to help the model use the tool well. for example, it should not be _too_ eager to use it and prefill everything. it should be reserved for cases where it truly saves effort from the user. phrase this somehow well, I'm bad at writing. it should also use some simple placeholders the user can easily replace with the actual content or edit

src/tui/session_chat_app.ts
markustherobot: Fixed in 1b87a50. The runtime parser now rejects whitespace-only drafts without trimming valid text, the advertised TypeBox schema includes the matching non-whitespace pattern, and the regression test covers blank input while verifying the editor remains unchanged.

PR #378, opened 2026-08-02
speed up tui startup

markusylisiurunen:
## why

Starting the regular Tau TUI eagerly loaded unrelated functionality and performed several serial initialization steps before the first screen could render. The largest user-visible delay on this VPS came from nested `AGENTS.md` discovery when Tau was started in the home directory: the bounded walk still visited 10,000 directories and performed redundant filesystem lookups in each one.

## what

The CLI now imports command-specific implementations only when those paths are selected, while the normal TUI path imports its dependencies directly instead of traversing the broad core barrel. Native image processing and hosted execution adapters are deferred until they are actually needed.

Nested `AGENTS.md` discovery now walks breadth-first so a deep early subtree cannot starve later shallow paths when the budget is reached. It uses each directory listing to detect an instructions file instead of probing a nonexistent path in every directory, avoids resolving the real path of ordinary child directories, and skips standard tool-managed cache and installation directories. Tool-managed home exclusions apply only to direct children of the execution home, so project directories such as `.cargo`, `.config`, or `.npm` remain visible at deeper levels. Platform-specific home data directories are selected inside the execution environment, while universally safe exclusions such as VCS metadata, dependencies, and build outputs remain global. Symlink traversal retains canonical-path scope and cycle checks.

The regular in-process TUI also reuses the host configuration already loaded by startup instead of reading it again inside the SDK, and terminal appearance detection runs concurrently with session bootstrap rather than adding up to 100ms serially.

## details

A five-run alternating-process benchmark on this VPS modeled the regular startup path through runtime config loading, execution-environment session bootstrap, and terminal detection. Median startup improved from 2.068s to 1.164s with AGENTS context enabled, and from 1.152s to 0.963s with `--no-agent-context-files`.

Isolating the host bootstrap showed a median improvement from 1.274s to 0.484s. Within that stage, execution-environment prompt discovery fell from 1.102s to 0.307s. The static startup graph also retains the earlier reduction in memory use from roughly 146 MB to 128 MB.

The original 10,000-directory bound was reached from this VPS home because it contains 184 real Cowork workspace roots, including 179 root and 477 total nested `AGENTS.md` files. Those project trees are intentionally not excluded; discovery now has a deterministic 8,192-directory budget and a separate 16-level depth guard.


src/sdk/client.ts
```diff
@@ -35,30 +34,33 @@ async function createInProcessSdkHost(options: TauSdkClientOptions): Promise<Loc
   const configDeps = createDefaultConfigDeps();
   const cwd = options.cwd ?? deps.env.cwd();
   const home = deps.env.home() || process.env.HOME || homedir();
-  const runtime = await loadRuntimeConfig(cwd, configDeps);
+  const config = loadConfig(cwd, configDeps);
```
markustherobot: **[P2] Keep the SDK startup contract in sync**

This switches client construction from `loadRuntimeConfig` to config-only loading, but `docs/sdk.md:81,97` still promises that `options.cwd` resolves config/content before sessions are created. Callers now get content resolution only when a session is created in its execution environment, so update those SDK docs to distinguish host config loading from per-session runtime content resolution.

markustherobot (commented):
One documentation mismatch remains in the SDK startup contract. I launched three read-only subreviews covering CLI lazy imports, SDK host initialization, and image-tool lazy loading; no other material issues survived verification.

PR #379, opened 2026-08-02
make Telegram messages read naturally

### LINKED ISSUE #375: the telegram messages should read naturally, not "project: tau"

markustherobot:
## why

Telegram status messages exposed project context as metadata such as `project: tau` and leaked internal state names such as `waiting-input`. These bot-authored messages should read like natural conversation.

## what

Project and session identifiers are now folded into prose across status, preference, provisioning, interruption, readiness, and lifecycle messages. Status output also translates internal session states into user-facing language. The Telegram adapter tests cover the revised status and provisioning messages, and the presentation rule is documented in AGENTS.md, README.md, and the Telegram guide.

## details

I treated the request as applying to every programmatic Telegram message that combines project and session context, including currently quiet lifecycle paths, rather than only the `/status` literal. There are no remaining questions.

fixes #375

### DIFF
--- a/src/core/telegram/adapter.ts
+++ b/src/core/telegram/adapter.ts
@@ -2675,7 +2675,7 @@ class TelegramAdapterImpl {
       this.notifySession(
         event.sessionId,
         [
-          `provisioning failed for your ${formatTelegramSessionName(event.sessionId, event.targetProjectId)}.`,
+          `provisioning ${event.targetProjectId} failed in your ${formatTelegramSessionName(event.sessionId, event.projectId)}.`,
           truncateText(event.diagnostic, MAX_PROVISION_DIAGNOSTIC_CHARS),
           "the session remains available.",
         ].join("\n"),
--- a/src/core/telegram/adapter.ts
+++ b/src/core/telegram/adapter.ts
@@ -664,10 +664,8 @@ function formatTelegramSender(user: TelegramUser | undefined): string {
   return `id ${user.id}`;
 }
 
-function formatTelegramSessionName(
-  session: Pick<TelegramSessionRecord, "id" | "projectId">,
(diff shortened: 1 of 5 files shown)


src/core/telegram/adapter.ts
```diff
@@ -2664,8 +2675,7 @@ class TelegramAdapterImpl {
       this.notifySession(
         event.sessionId,
         [
-          formatSessionHeadline(event.sessionId, "provision failed"),
-          `project: ${event.targetProjectId}`,
+          `provisioning failed for your ${formatTelegramSessionName(event.sessionId, event.targetProjectId)}.`,
```
markustherobot: **[P2] Keep the provision target separate from the session project**

For a composite project, `event.projectId` is the owning session project (for example, `platform`), while `event.targetProjectId` is only the member repository whose provision hook failed (`alpha`). This now tells the user that `s1` is “your alpha session” even though the actual session is `platform`. Build the session name from `event.projectId` and mention the target separately, such as “provisioning alpha failed in your platform session s1”; a composite provisioning-failure test would pin the distinction.

markustherobot (commented):
The wording refactor is otherwise coherent, but one composite-project notification becomes inaccurate. I launched 3 subreviews covering status/state formatting, non-status notifications, and tests/documentation; all completed successfully.

markusylisiurunen:
@markustherobot fix the findings

src/core/telegram/adapter.ts
markustherobot: Fixed in 6cf3670. Provision failure messages now use `projectId` for the owning session and name `targetProjectId` separately; the composite-project test pins `alpha` failing inside the `platform` session.

PR #380, opened 2026-08-02
add Telegram effort selectors

### LINKED ISSUE #376: ability to change reasoning effort from telegram

markustherobot:
## why

Telegram sessions exposed their current reasoning effort through `/status`, but there was no simple way to change it without leaving Telegram.

## what

Added `/effort_low`, `/effort_medium`, `/effort_high`, and `/effort_xhigh` as advertised Telegram commands. Each selector requires an active session, checks that the active model and persona support the requested level, updates reasoning through the existing SDK session contract, and confirms the selected effort.

The Telegram session manager now delegates reasoning updates through both its root and chat-scoped views, preserving project and owner isolation. Command registration, selector handling, active-session and argument validation, model and persona compatibility, running-session behavior, scoped ownership, and the Telegram command budget are covered by tests. Telegram documentation now describes the selectors and their next-turn semantics.

## details

The selector surface is intentionally limited to the four agreed effort levels even though the underlying session protocol supports additional values. A selector is rejected when the active model does not support reasoning or the active persona excludes that level. Changes remain valid while a turn is running and follow the existing session behavior: they apply to the next independently submitted or queued turn. Reserving the four new built-ins reduces the maximum advertised `/use_*` commands from 96 to 92.

No unresolved questions remain.

fixes #376


src/core/telegram/adapter.ts
```diff
@@ -2099,6 +2110,30 @@ class TelegramAdapterImpl {
...
+    }
+
+    const session = await this.requireActiveSession(chatId);
+    if (!session) {
+      return;
+    }
+
+    try {
+      const sessionManager = this.getSessionManagerForChat(chatId);
+      await sessionManager.setReasoning(session.id, reasoning);
```
markustherobot, line 2158: **[P2] Reject efforts the active persona cannot use**

`/effort_xhigh` reaches this call for every session, including Opus 4.6 and the Gemini personas whose `allowedReasoningLevels` exclude `xhigh`, as well as custom non-reasoning models. The host persists the requested value and this handler confirms it, but the pinned model runtime clamps unsupported levels (for example, `xhigh` to `high`), so `/status` can report `xhigh` while turns actually run at another effort. Check the active snapshot's model capability and current persona's allowed levels before setting it, and return an unsupported-level response instead of confirming success.

markustherobot (commented):
Found one normal-priority correctness gap around persona and model effort compatibility. I launched three focused Luna subreviews covering adapter command handling, the session/protocol reasoning boundary, and config/tests/docs; all completed successfully. CI and the focused Telegram suites pass.

markusylisiurunen:
@markustherobot fix findings

src/core/telegram/adapter.ts
markustherobot: Fixed in cfdcd6c. The adapter now reads the active snapshot before mutation and rejects levels excluded by either the model capability or the current persona. Regression coverage includes both cases.

PR #382, opened 2026-08-03
show goal state in Telegram status

### LINKED ISSUE #377: pursuing a goal should be made visible in a telegram /status

markustherobot:
## why

Telegram `/status` reports whether a session is running or waiting, but it does not reveal when the session is autonomously pursuing a goal. That makes active goal work invisible outside the TUI.

The issue discussion established that the goal objective must not be included because it can be arbitrarily long.

## what

Telegram status replies now add a short goal-state sentence when the session snapshot has a goal. Active goals report that the session is pursuing a goal, while blocked goals report that its goal is blocked. Replies with no goal remain unchanged.

The Telegram adapter coverage exercises both goal states with objective text that is intentionally absent from the replies.

## details

This is derived directly from the canonical snapshot goal state, so no additional Telegram session state or protocol fields are introduced. There are no remaining implementation questions.

fixes #377

### DIFF
--- a/src/core/telegram/adapter.ts
+++ b/src/core/telegram/adapter.ts
@@ -710,11 +710,16 @@ function formatSessionStatus(
     return `${status}${preferenceStatus}`;
   }
 
+  const goalStatus = snapshot.goal
+    ? snapshot.goal.status === "active"
+      ? " it is pursuing a goal."
+      : " its goal is blocked."
+    : "";
   const model = snapshot.bootstrap.model.name || snapshot.bootstrap.model.id;
   const reasoning = snapshot.settings.reasoning ?? "none";
   const context = formatTelegramContextUsage(snapshot);
   const cost = formatTelegramSessionCost(getTelegramSessionCostTotal(snapshot));
-  return `${status} it is using ${model} with ${reasoning} reasoning. context usage is ${context}. cumulative cost is ${cost}.${preferenceStatus}`;
+  return `${status}${goalStatus} it is using ${model} with ${reasoning} reasoning. context usage is ${context}. cumulative cost is ${cost}.${preferenceStatus}`;
 }
 
 function ensureTerminalPunctuation(text: string): string {
--- a/test/telegram_adapter.test.js
+++ b/test/telegram_adapter.test.js
@@ -425,6 +425,60 @@ describe("telegram adapter", () => {
     }
   });
 
+  it("preserves detailed status when there is no goal", async () => {
+    const chatId = 13;
+    const apiHarness = createApiHarness([
+      [
+        {
+          update_id: 1,
+          message: {
+            chat: { id: chatId, type: "private" },
+            from: { id: 7 },
+            text: "/status",
+          },
+        },
+      ],
+    ]);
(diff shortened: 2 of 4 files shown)


src/core/telegram/adapter.ts
```diff
@@ -710,11 +710,16 @@ function formatSessionStatus(
...
+    ? snapshot.goal.status === "active"
+      ? " it is pursuing a goal."
+      : " its goal is blocked."
+    : "";
   const model = snapshot.bootstrap.model.name || snapshot.bootstrap.model.id;
   const reasoning = snapshot.settings.reasoning ?? "none";
   const context = formatTelegramContextUsage(snapshot);
   const cost = formatTelegramSessionCost(getTelegramSessionCostTotal(snapshot));
-  return `${status} it is using ${model} with ${reasoning} reasoning. context usage is ${context}. cumulative cost is ${cost}.${preferenceStatus}`;
+  return `${status}${goalStatus} it is using ${model} with ${reasoning} reasoning. context usage is ${context}. cumulative cost is ${cost}.${preferenceStatus}`;
```
markustherobot, line 722: **[P2] Keep the Telegram command docs in sync**

`docs/telegram.md` explicitly describes the fields reported by `/status`, and the repository guidance requires documented command behavior to be updated with code changes. This adds active/blocked goal state without updating that contract, so the new output remains undiscoverable there. Update the `/status` documentation to mention the goal-state sentence and that the objective is intentionally omitted.

test/telegram_adapter.test.js
```diff
@@ -371,7 +372,13 @@ describe("telegram adapter", () => {
     ]);
 
     const managerHarness = createSessionManagerHarness([], {
-      createSnapshot: () => createStatusSnapshot(),
+      createSnapshot: () =>
```
markustherobot, line 375: **[P2] Preserve detailed no-goal coverage**

Repurposing this fixture removes the only test that sends `/status` with an available snapshot whose goal is `null`; the remaining no-goal status assertion has no snapshot and returns through the earlier short path. Since `null` is the normal goal state and the PR promises those detailed replies remain unchanged, keep the original assertion or add a separate no-goal case alongside the active and blocked cases.

markustherobot (commented):
The goal-state formatting follows the canonical snapshot contract and avoids exposing objective text. I found two follow-ups around keeping the user-facing contract documented and preserving regression coverage. I launched 2 focused subreviews covering goal/session lifecycle semantics and Telegram formatting/test coverage; both results were verified against the repository before publication.

markusylisiurunen:
@markustherobot fix

src/core/telegram/adapter.ts
markustherobot: Updated `docs/telegram.md` and the README command summary to document active/blocked goal status and the intentional omission of objective text.

test/telegram_adapter.test.js
markustherobot: Added a separate detailed `goal: null` status case that pins the pre-change reply alongside the active and blocked goal coverage.

PR #383, opened 2026-08-03
retry transient Telegram deliveries

### LINKED ISSUE #381: Final Telegram response is lost after transient send failure
## Bug

A completed Telegram session can silently lose its final assistant response when the Telegram API request fails transiently. The response remains in the Tau session snapshot, but it is never delivered or retried, so the chat appears to stop after the last progress update.

## Observed behavior

On 2026-08-02, a long-running `markustherobot` session completed normally and persisted its final assistant message. The Telegram adapter remained alive, but its output ended with:

```text
[telegram:markustherobot:warn] failed to send telegram typing action: telegram sendChatAction network request failed: fetch failed: read ECONNRESET (ECONNRESET)
[telegram:markustherobot:warn] failed to send telegram typing action: telegram sendChatAction network request failed: fetch failed: read ECONNRESET (ECONNRESET)
[telegram:markustherobot:warn] failed to send telegram notification: telegram sendRichMessage network request failed: fetch failed: read ECONNRESET (ECONNRESET)
```

The user received the preceding progress update, but not the final result. The adapter process and tmux session continued running.

## Expected behavior

Final assistant responses should not be permanently lost because of a transient Telegram network error. Tau should retry retryable delivery failures with bounded backoff. If retries are exhausted, the failed outbound response should remain recoverable and delivery failure should be surfaced clearly rather than silently ending the conversation.

Typing-action failures do not need the same durability guarantee.

## Reproduction outline

1. Start a Telegram-backed Tau session.
2. Let the agent send a progress update and complete with a final response.
3. Cause the final `sendRichMessage` request to fail with a transient network error such as `ECONNRESET`.
4. Observe that the final message is persisted in the Tau session but absent from Telegram, with no later retry.

markustherobot:
## why

A transient Telegram network or API failure could drop a committed assistant response permanently even though the response remained in the Tau session snapshot. Later notifications could also overtake a response once retry backoff was introduced.

## what

Telegram requests now distinguish retryable network, HTTP 408/429/5xx, and matching Telegram API failures from permanent request and formatting failures. Outbound message chunks use an abortable 30-second deadline per attempt and three total attempts with 1-second and 5-second retry delays, honoring longer Telegram `retry_after` guidance.

Session notifications are serialized per chat, tracked through adapter shutdown, and carry the assistant snapshot message ID into structured exhausted-delivery logs. Tests cover transient recovery, permanent failure, exhausted retries, preserved ordering, and message identity, with the Telegram behavior documentation updated to match.

## details

Retries are intentionally in-process. The persisted Tau snapshot and logged message ID make an exhausted delivery recoverable by an operator, but automatic redelivery after runner restart would require a durable Telegram outbox and remains outside this minimal fix. Telegram does not provide an idempotency key, so a lost successful HTTP response can still cause an at-least-once duplicate on retry.

fixes #381


src/core/telegram/adapter.ts
```diff
@@ -2886,22 +2979,71 @@ class TelegramAdapterImpl {
...
+    chatId: number,
+    text: string,
+    options: TelegramNotificationOptions,
+  ): void {
+    const previousTask = this.notificationQueueTailByChat.get(chatId) ?? Promise.resolve();
+    const queuedTask = previousTask
+      .then(async () => {
+        if (this.abortController.signal.aborted) {
+          return;
+        }
```
markustherobot, line 3030: **[P1] Put a deadline on each serialized delivery**

If `sendMessage` or `sendRichMessage` stalls, this await keeps the per-chat tail pending, so every later notification for that chat, including the final response, is blocked and no retry starts. The built-in outbound `fetch` has no application timeout or cancellation signal, so one hung network call can turn into a chat-wide delivery outage. Add an abortable per-attempt deadline and classify that timeout as retryable so the queue advances after the bounded policy is exhausted.

markustherobot (commented):
Found material gaps in delivery reliability across request classification, serialized queue liveness and shutdown, and operator-visible recovery metadata. I launched four subreviews covering request/retry semantics, notification serialization, message identity, and regression coverage; all completed, and I verified the findings against the branch.

src/core/telegram/adapter.ts
```diff
@@ -2886,22 +2979,71 @@ class TelegramAdapterImpl {
...
+            : new TelegramDeliveryError(error, 1, false);
+        this.log(
+          "error",
+          deliveryError.retryable
+            ? "telegram notification delivery retries exhausted"
+            : "failed to send telegram notification",
+          {
+            sessionId,
+            chatId,
+            ...(options.rich === true ? { messageId: options.messageId } : {}),
```
markustherobot, line 3051: **[P1] Surface the recovery identifiers in runtime logs**

These identifiers are present in the adapter entry, but `formatRuntimeLog()` in `src/core/telegram/runtime.ts` reads only `data.cause`; the CLI therefore prints an exhausted-delivery line without `sessionId`, `chatId`, `messageId`, or `attempts`. That makes the documented operator recovery path unusable. Preserve these fields in the formatted runtime line and cover the actual runtime sink in a test.

src/core/telegram/adapter.ts
```diff
@@ -973,41 +1037,67 @@ function createTelegramApi(botToken: string): TelegramApi {
...
+      throw new TelegramRequestError(
         `telegram ${method} network request failed: ${formatTelegramNetworkError(error)}`,
+        { retryable: true, cause: error },
       );
     }
 
     if (!response.ok) {
-      const detail = await readTelegramHttpErrorDetail(response);
-      throw new Error(
+      const { detail, errorCode, retryAfterMs } = await readTelegramHttpError(response);
```
markustherobot: **[P1] Preserve retryability when reading an HTTP error body fails**

If Telegram returns a retryable 429/5xx but the connection resets while `response.text()` reads its body, the raw transport error escapes before the known status is classified. `sendWithRetry()` then sees a non-`TelegramRequestError` and makes only one attempt, recreating the transient-loss case this PR is meant to fix. Catch body-read failures and wrap them using the response status; likewise distinguish `response.json()` transport failures from genuinely malformed JSON.

src/core/telegram/adapter.ts
```diff
@@ -3063,16 +3207,45 @@ class TelegramAdapterImpl {
...
       if (index < chunks.length - 1) {
         await this.wait(TELEGRAM_MESSAGE_SPLIT_DELAY_MS);
       }
     }
   }
 
+  private async sendWithRetry(send: () => Promise<void>): Promise<void> {
+    for (let attempt = 1; ; attempt += 1) {
+      try {
+        const result = await this.raceWithAbort(send());
```
markustherobot: **[P2] Do not start later chunks after shutdown**

`send()` is evaluated before `raceWithAbort()` checks the signal. If `close()` aborts during an in-flight chunk or the inter-chunk delay, the wrapper returns but the chunk loop continues, and each later iteration starts another API request after shutdown before immediately returning; those requests are no longer tracked. Check abort before invoking the thunk and propagate an aborted result so the chunk loop stops.

markusylisiurunen:
@markustherobot fix

src/core/telegram/adapter.ts
markustherobot: Added a 30-second abortable deadline to every delivery attempt. The send contract now receives an AbortSignal, timeouts remain retryable, and a regression test verifies a stalled send is aborted before the queue retries.

src/core/telegram/adapter.ts
markustherobot: The runtime formatter now includes sessionId, chatId, messageId, and attempts in the emitted CLI line. A runtime-sink test asserts the complete exhausted-delivery log.

src/core/telegram/adapter.ts
markustherobot: Response-body transport reads are now handled separately from JSON parsing. HTTP error read failures retain status-based retryability, successful-response read failures retry, and malformed JSON remains permanent, with coverage for each case.

src/core/telegram/adapter.ts
markustherobot: sendWithRetry now checks shutdown before invoking the send thunk and propagates an explicit aborted result to the chunk loop. The shutdown regression test verifies the in-flight signal is aborted and no later chunk starts.

Issue #384, opened 2026-08-03
allow starting a TUI through SDK


markusylisiurunen:
it would be very useful to be able to use the SDK to define completely custom agents (custom system prompt, tools, etc.) but still allow using the same Tau TUI we have built.

markusylisiurunen:
We discussed splitting this into two layers. The command-backed client-tool extension for the stock `tau` TUI will be implemented separately; this issue should own the fully programmatic, clean TUI API.

The intended API is a public TUI entry point, likely `@markusylisiurunen/tau/tui`, with a lifecycle-oriented function such as `launchTauTui()`. It should support local, WebSocket, and stdio transports and return a handle (`stop()` plus a completion signal) rather than calling `process.exit()`.

The important requirement is a genuinely custom runtime mode, not merely exposing the current `SessionChatApp` with more flags. Session creation should use an explicit union along these lines:

```ts
type SessionRuntime =
  | {
      kind: "configured";
      personaId?: string;
      reasoning?: ReasoningLevel;
    }
  | {
      kind: "custom";
      model: ModelTarget;
      systemPrompt: string;
      hostTools: HostToolName[];
    };
```

A custom launch could then look roughly like:

```ts
await launchTauTui({
  connection: { kind: "local", cwd: process.cwd() },
  runtime: {
    kind: "custom",
    model: {
      provider: "anthropic",
      model: "claude-opus-4-6",
      reasoning: "high",
    },
    systemPrompt: "You are an assistant for my application.",
    hostTools: [],
  },
  clientTools: [searchTool, saveTool],
  tui: { builtInClientTools: false },
});
```

In `custom` mode, the model-visible agent definition must be exact. Tau may still use config for infrastructure such as credentials, model metadata, themes, and transport settings, but it must not silently add behavioral content. Specifically, custom mode should add no persona prompt, prompt template, skills, AGENTS.md context, model-system notice, execution-environment context block, built-in client tools, implicit host tools, goal tools, or subagents unless the caller explicitly selected them. `hostTools: []` must mean no host tools. Programmatic `clientTools` execute on the TUI/client machine; explicitly selected host tools execute in the session execution environment.

[...]

PR #385, opened 2026-08-03
add command-backed TUI client tools

markusylisiurunen:
## why

Tau's client-tool protocol already allows SDK clients to expose tools that run on the client machine, but the stock TUI only advertises its built-in diff review and input prefill tools. Extending the normal `tau` and `tau attach` experience therefore requires modifying Tau or replacing the TUI.

Command-backed client tools provide a language-independent extension point while retaining the standard configured Tau experience. Global config grants trust to execute a local definition, while workspace config can select which trusted definitions are exposed to the model.

## what

Add global `clientTools` definitions with a model-facing name, required default-enabled state, description, object JSON Schema, command, and optional arguments and timeout. Without a workspace selection, the TUI advertises default-enabled definitions; the nearest project `enabledClientTools` value can select an exact subset. `--no-client-tools` disables configured and built-in client tools together.

Tau verifies that each parameter schema has root `type: "object"` and otherwise treats its contents as trusted user configuration. Each invocation validates model arguments against that schema, starts the command directly without a shell, and exchanges a versioned JSON request and strict JSON content result over stdin and stdout.

Commands run from the TUI's current working directory with the TUI process environment unchanged. Execution caps output, enforces timeouts, and terminates the complete process group with grace-period escalation even when its leader exits first. The SDK aborts active client-tool executions on client close or terminal transport failure and waits for cleanup even if transport shutdown fails before propagating that error. Stdio and WebSocket write failures enter the same terminal failure path.

## details

Relative command paths resolve from the user's home directory. Project config can select globally trusted names but cannot define executable behavior. Duplicate names are rejected, unknown selected names are ignored, and existing host-tool collision checks remain authoritative.

Command tools execute unsandboxed on the TUI machine. In an attached session, the remote host can dispatch an advertised tool, so enabling one grants that host authority to invoke the globally configured local command. The TUI startup intro lists the effective configured command tools for both local and attached sessions.

README, SDK, RPC, and architecture documentation describe the configuration, selection, process protocol, execution location, lifecycle, and trust boundary.

This is the configured stock-TUI extension layer discussed alongside #384. The fully programmatic clean TUI runtime remains separate work tracked by that issue.


src/tui/command_client_tools.ts
```diff
@@ -0,0 +1,109 @@
...
+          ...(config.env ?? {}),
+        },
+        detached: true,
+        signal: context.signal,
+        timeoutMs,
+        maxCaptureBytes: DEFAULT_COMMAND_CAPTURE_BYTES,
+        maxCaptureMode: "terminate",
+        maxCaptureStrategy: "head",
+        captureOutput: "split",
+        killGraceMs: COMMAND_KILL_GRACE_MS,
```
markustherobot, line 50: **[P1] Await process termination before exiting**

On TUI shutdown, `TauSdkClient.close()` aborts active executions but does not await `runClientTool()`, and both main shutdown paths call `process.exit()` immediately after `app.stop()`. If this detached command traps `SIGTERM`, Tau exits before the delayed `SIGKILL` can fire, leaving the process group alive to continue side effects. Track and await active command executions, or otherwise complete the process-group escalation before close returns.

src/core/config/client_tools.ts
```diff
@@ -0,0 +1,119 @@
...
+  executionTimeoutMs?: number;
+};
+
+const CommandClientToolSchema = z
+  .object({
+    name: z.string().trim().min(1),
+    description: z.string().trim().min(1),
+    parameters: z
+      .record(z.string(), z.unknown())
+      .refine((value) => value.type === "object", "must be an object JSON Schema"),
```
markustherobot: **[P2] Validate the complete parameter schema**

Checking only the root `type` accepts malformed nested schemas such as a property with `type: "bogus"`. TypeBox treats that nested type as unconstrained, so an out-of-contract value can pass `Check` and reach the command, while a provider may instead reject the advertised schema entirely. Validate the full supported JSON Schema at config load rather than casting an unchecked record to `TSchema` later.

src/tui/session_chat_app.ts
```diff
@@ -168,6 +184,8 @@ export class SessionChatApp {
       clientTools: createTuiClientTools({
         enabled: options.clientToolsEnabled,
         getController: () => controller,
+        commandTools: options.config?.clientTools,
```
markustherobot: **[P2] Close the stdio transport when tool setup fails**

For `tau attach -- <command>`, the stdio child is spawned before this `createTuiClientTools()` call. A configured name colliding with `diff_review` or `prefill_input` throws here before an SDK client exists to own cleanup, so the dedicated protocol process is never closed. Construct the tools before spawning the transport, or close the transport in this setup failure path.

markustherobot (commented):
Found material lifecycle and validation gaps in the new command-tool path. The required CI check passes. I launched four focused subreviews covering config trust/schema validation, subprocess execution, TUI/SDK/host integration, and protocol/test alignment; all completed successfully.

markusylisiurunen:
@markustherobot fix the findings

src/tui/command_client_tools.ts
markustherobot: Fixed in 59e66ad. The SDK now tracks active client-tool executions, aborts them on close, and waits for cleanup alongside transport shutdown, so process-group escalation completes before TUI exit.

src/core/config/client_tools.ts
markustherobot: Fixed in 59e66ad. Global config loading now recursively validates supported JSON Schema keywords and nested schemas before storing parameters as TSchema; malformed nested types are covered by regression tests.

src/tui/session_chat_app.ts
markustherobot: Fixed in 59e66ad. SessionChatApp.connect now constructs and validates the complete TUI tool set before creating either transport, with a regression test proving a colliding tool does not spawn the stdio child.

markusylisiurunen:
@markustherobot review

src/tui/command_client_tools.ts
```diff
@@ -0,0 +1,107 @@
...
+        killProcessGroup: true,
```
markustherobot, line 51: **[P1] Complete escalation for the whole process group**

When cancellation or timeout sends `SIGTERM`, the command leader can exit while a descendant in the same group ignores the signal. `spawnWithCapture` handles the leader's `close` event by clearing the grace-period kill timer, so this process-group mode never sends `SIGKILL` and SDK shutdown returns with the descendant still running; I reproduced this with a leader that spawned a signal-resistant child. Keep escalation alive after the leader closes, or explicitly verify and kill the group before resolving, so model-selected local processes cannot outlive TUI shutdown.

src/sdk/session.ts
```diff
@@ -136,11 +144,13 @@ class TauSdkClientImpl implements TauSdkClient {
 
     const abortController = new AbortController();
     this.clientToolAbortControllers.set(message.callId, abortController);
-    void this.runClientTool(tool, message, abortController)
+    const execution = this.runClientTool(tool, message, abortController)
```
markustherobot, line 158: **[P2] Abort executions when the transport dies**

If the WebSocket closes or the stdio RPC child exits while this execution is pending, the transport only rejects protocol requests; it does not call `client.close()` or abort these controllers. The local command therefore keeps running until its configured timeout even though its owning host is gone and can no longer send cancellation. Surface terminal transport failure to the SDK and abort active client tools when that happens.

src/core/config/client_tools.ts
```diff
@@ -0,0 +1,202 @@
...
+        writeOnly: z.boolean().optional(),
+        deprecated: z.boolean().optional(),
+        multipleOf: z.number().positive().optional(),
+        maximum: z.number().optional(),
+        exclusiveMaximum: z.number().optional(),
+        minimum: z.number().optional(),
+        exclusiveMinimum: z.number().optional(),
+        maxLength: nonNegativeIntegerSchema.optional(),
+        minLength: nonNegativeIntegerSchema.optional(),
+        pattern: z.string().optional(),
```
markustherobot: **[P2] Validate regex-bearing schema keywords**

A schema with `pattern: "["` (or an invalid `patternProperties` key) passes config loading, but TypeBox compiles it during `Check()` and throws `Invalid regular expression` on invocation, leaving the advertised tool unusable instead of producing a config diagnostic. Compile every recursively accepted regex with the same Unicode semantics during schema parsing, including `patternProperties` keys.

markustherobot (commented):
Found material gaps in command cancellation and schema acceptance. The required build check passes. I launched four focused subreviews covering config/schema trust, subprocess execution, SDK/TUI lifecycle, and end-to-end contract/tests; all completed successfully.

markusylisiurunen:
@markustherobot fix the findings

src/tui/command_client_tools.ts
markustherobot: Fixed in 4ad48d2. Group-mode termination now keeps a referenced escalation pending after the leader closes, checks whether the group remains, and sends SIGKILL after the grace period before resolving. A real subprocess regression covers a signal-resistant descendant.

src/sdk/session.ts
markustherobot: Fixed in 4ad48d2. The transport contract now exposes terminal failures; stdio and WebSocket transports publish them, and the SDK aborts all active client-tool controllers immediately. Tests cover both concrete transports and SDK cancellation.

src/core/config/client_tools.ts
markustherobot: Fixed in 4ad48d2. Config parsing now compiles every accepted pattern and patternProperties key with the same Unicode RegExp semantics TypeBox uses. Invalid values produce config diagnostics and are covered by regression tests.

markusylisiurunen:
@markustherobot review

src/core/config/client_tools.ts
```diff
@@ -0,0 +1,257 @@
...
+}, "must be a valid Unicode regular expression");
+const jsonSchemaThenKeyword = "then";
+const jsonSchemaSchema: z.ZodType<unknown> = z.lazy(() =>
+  z.union([
+    z.boolean(),
+    z
+      .object({
+        $id: z.string().optional(),
+        $schema: z.string().optional(),
+        $ref: z.string().optional(),
```
markustherobot: **[P2] Reject references the executor cannot resolve**

A schema containing an external or missing `$ref` passes this validator, but execution calls `Check(config.parameters, args)` without a reference set, so TypeBox evaluates the unresolved reference as `false` and rejects every invocation before starting the command. Validate that references resolve within the configured schema, or reject unsupported external and unresolved references during config loading.

src/sdk/session.ts
```diff
@@ -111,13 +115,24 @@ class TauSdkClientImpl implements TauSdkClient {
...
+    this.closePromise ??= this.closeClient();
+    return this.closePromise;
+  }
+
+  private async closeClient(): Promise<void> {
     this.unsubscribeClientTool();
+    this.unsubscribeTransportFailure();
+    const executions = [...this.clientToolExecutions.values()];
+    this.abortClientTools();
+    await Promise.all([this.transport.close(), ...executions]);
```
markustherobot: **[P2] Await handlers even when transport close fails**

`Promise.all` rejects as soon as `transport.close()` rejects, so `client.close()` can settle before an aborted handler finishes. In-process host shutdown can reject on persistence or disposal errors, and custom transports may reject too; a stock TUI exit can then call `process.exit()` while a detached command group is still in its grace-period escalation. Await all executions regardless of the transport outcome, then rethrow the close error.

src/transport/stdio_session_transport.ts
```diff
@@ -405,6 +416,7 @@ export class StdioSessionProtocolTransport implements SessionProtocolTransport {
     }
 
     this.fatalError = error;
+    notifySessionProtocolFailureListeners(this.failureListeners, error);
```
markustherobot, line 414: **[P2] Treat failed writes as terminal transport failures**

The new failure signal is emitted only through `failTransport`, but the stdio request write callback/catch merely rejects that one request; the analogous WebSocket `send()` catch does the same. On `EPIPE` or a send failure, active local client tools are therefore not aborted and later requests keep targeting a dead transport. Route these write errors through `failTransport` in both transports.

markustherobot (commented):
Found material gaps in schema validation and client-tool cleanup across transport failure paths. The required CI check passes. I launched four focused subreviews covering config/schema trust, subprocess execution, SDK/transport lifecycle, and TUI integration/tests; all completed successfully.

markusylisiurunen:
@markustherobot fix the findings

src/core/config/client_tools.ts
markustherobot: Fixed in f51762f. Config loading now accepts only resolvable document-local fragment references, including JSON pointers and local anchors, and rejects missing or external references before advertisement. Regression coverage includes valid local, missing, and external cases.

src/sdk/session.ts
markustherobot: Fixed in f51762f. SDK close now waits for transport shutdown and every active handler with all-settled semantics, then rethrows the transport close error only after handler cleanup completes. The regression keeps a handler pending while transport close rejects.

src/transport/stdio_session_transport.ts
markustherobot: Fixed in f51762f. Asynchronous stdio write errors and synchronous WebSocket send errors now call failTransport, which publishes terminal failure, rejects all pending requests, and prevents later requests. Both concrete paths have regression tests.

PR #389, opened 2026-08-03
render user messages through session state

### LINKED ISSUE #268: streamline user message rendering through session protocol
## why

The TUI currently has a local-only optimization for normal user submits: it renders the user message immediately with a client-generated history id, then sends the submit over the session protocol and reconciles with the host-owned snapshot later.

That makes submit handling more complicated and error-prone. It creates special cases around idle steering, queued messages, optimistic history ids, hidden/rendered message tracking, and snapshot reconciliation. I would rather keep this canonical and simpler: user-message rendering should be driven by host-owned protocol events, like the rest of session state.

## what

Remove the TUI-local optimistic user-message rendering path for submitted user messages. Submitting user text should simply send the relevant session protocol request to the host. The host should record/run the turn and emit the user-message/session deltas, and the TUI should render the message when those protocol events arrive.

While doing this, look for similar local-only presentation optimizations or duplicate state paths that can be streamlined into the same host-owned protocol flow.

Likely areas to inspect:

- `SessionChatController.submitUserText()` and client-generated `session-user-*` ids
- `clientRenderedUserMessages`
- `renderedMessageIds` reconciliation around locally inserted user messages
- queued-message draining and steering paths
- direct bash / hidden history handling, to distinguish necessary local presentation state from avoidable optimistic rendering

markusylisiurunen:
## why

The TUI rendered submitted user messages optimistically with client-generated history ids, then reconciled that local state with the host-owned session snapshot. This duplicated ownership of message identity and presentation, complicated failure handling, and let returned diff reviews appear as committed user messages before session recording succeeded.

## what

User submissions and queued messages now rely on host-generated history ids and render from session deltas and snapshots. The client-side user-message cache and its reconciliation branches are removed.

Returned diff reviews follow the same path. Their persisted Tau metadata preserves review styling across recovery and remote observation, while recording failures without an observed commit leave the review card failed instead of showing an uncommitted user message or a false success notification. A review already confirmed by a session delta remains committed if the record response is lost. The session protocol and stored-session document versions advance with the new persisted metadata contract, and the now-unused chat-view replacement adapter is removed.

fixes #268


src/tui/chat_controller/diff_review_user_message.ts
```diff
@@ -5,6 +5,7 @@ import type { DiffReviewReturnedReview } from "./diff_review_service.js";
 export function formatDiffReviewUserMessage(review: DiffReviewReturnedReview): string {
   return formatTauUserText({
     text: review.review,
+    metadata: [{ type: "diff-review", version: 1 }],
```
markustherobot, line 8: **[P1] Version the persisted diff-review metadata boundary**

This marker is generated by the TUI but parsed strictly by the host, so a new protocol-v6 TUI can send it to an older protocol-v6 host. The old host accepts and persists the record, then throws `invalid tau user metadata: unknown record type` on the next history projection/model turn; an older binary also cannot safely consume the persisted session. Bump the session protocol (and the stored format where downgrade rejection is required) so mixed versions fail at the boundary, or otherwise establish an explicitly compatible metadata contract.

markustherobot (commented):
Two material correctness issues remain at the failure and compatibility boundaries. I launched three read-only subreviews covering session submission/reconciliation, the diff-review lifecycle, and metadata/rendering/tests; all completed. The host-owned rendering direction is sound, but these boundaries need tightening before merge.

src/tui/chat_controller/diff_review_service.ts
```diff
@@ -327,17 +327,17 @@ export class DiffReviewService {
...
-      });
+      try {
+        await onReviewReturned({
+          diffCommand: state.diffCommand,
+          reviewedFiles: state.reviewedFiles,
+          review: result.review,
+          historyEntryId: state.messageId,
+        });
+      } catch (error) {
+        state.messageFinalized = false;
```
markustherobot, line 338: **[P2] Preserve a review already confirmed by a delta**

`session.record` persists and emits the user-message delta before its RPC success response. If the transport closes after that delta but before the response, this catch resets finalization even though the controller snapshot already contains `historyEntryId`; the outer failure path then overwrites the committed review with a failed card and can prompt a duplicate retry. On rejection, first honor an already observed matching snapshot entry, or show an unconfirmed state when commit status is genuinely unknown.

markusylisiurunen:
@markustherobot fix the findings, try to find clean fixes that don't over-engineer or be overly defensive

src/tui/chat_controller/diff_review_user_message.ts
markustherobot: Bumped the session protocol to v7 and the stored-session document to v4, with a v3-to-v4 pass-through migration. Mixed-version clients now fail at the wire boundary, and older Tau versions reject sessions written with the new persisted metadata contract.

src/tui/chat_controller/diff_review_service.ts
markustherobot: The controller now accepts a record failure only when the same observed session snapshot already contains the requested committed user entry. Added a regression test for a delta arriving before the response-side transport failure; the review remains committed and is not replaced by a failed card.

PR #390, opened 2026-08-03
report stale auth refreshes

markusylisiurunen:
## why

`tau auth list` silently fell back to cached Codex account and usage data when refreshes failed. Expired credentials could therefore appear alongside stale quota data without explaining that the account could no longer refresh.

## what

Track credential and usage refresh outcomes separately when listing Codex accounts. The CLI now identifies expired credentials that require re-authentication, warns about refresh failures for unexpired credentials, and labels cached usage as stale when its refresh fails.

Add provider and CLI regression coverage for successful refreshes and stale fallback behavior.

### DIFF
--- a/src/core/auth/providers/openai_codex.ts
+++ b/src/core/auth/providers/openai_codex.ts
@@ -38,6 +38,15 @@ class UnexpectedUsageWindowError extends Error {
 type CodexAccount = StoredOAuthAccount;
 type UnknownRecord = Record<string, unknown>;
 type AccountPriorityCandidate = { usage?: AuthAccountUsage; index: number };
+type RefreshStatus = "not-requested" | "succeeded" | "failed";
+type AccountRefreshResult = {
+  account: CodexAccount;
+  refreshStatus: Exclude<RefreshStatus, "not-requested">;
+};
+type UsageSnapshotResult = {
+  usage?: AuthAccountUsage;
+  refreshStatus: RefreshStatus;
+};
 
 export class OpenAICodexAdapter implements AuthProviderAdapter {
   readonly id = PROVIDER_ID;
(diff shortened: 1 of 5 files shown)


src/core/auth/providers/openai_codex.ts
```diff
@@ -110,12 +119,20 @@ export class OpenAICodexAdapter implements AuthProviderAdapter {
           return undefined;
         }
         const identity = decodeIdentity(currentAccount.access);
+        const credentialRefreshStatus =
+          accountRefresh.refreshStatus === "succeeded" ||
+          !hasSameCredentialGeneration(currentAccount, accountRefresh.account)
```
markustherobot, line 124: **[P2] Compare against the attempted credential generation**

If this refresh rejects after another process has already replaced the same account's credentials, the catch path reloads that replacement and returns it as `accountRefresh.account`. This comparison then compares the current replacement with itself, leaves the status as `failed`, and makes `tau auth list` warn even though valid newer credentials won the race. Preserve the original `account` generation (or an explicit superseded outcome) so a changed current generation is classified correctly.

src/core/auth/providers/openai_codex.ts
```diff
@@ -272,38 +294,41 @@ export class OpenAICodexAdapter implements AuthProviderAdapter {
...
         usage: refreshedUsage,
       }));
       if (updateResult.status === "missing") {
-        return undefined;
+        return { refreshStatus: "failed" };
       }
-      return updateResult.status === "changed" ? updateResult.account.usage : refreshedUsage;
+      return {
+        usage: updateResult.status === "changed" ? updateResult.account.usage : refreshedUsage,
+        refreshStatus: "succeeded",
```
markustherobot: **[P2] Do not mark a discarded usage refresh successful**

If credentials rotate between `fetchUsage()` and `updateStoredOAuthAccount()`, the update returns `changed`; line 324 then returns the newer account's cached usage, which may be stale or absent, while this line still marks the refresh successful. The CLI consequently suppresses the new stale/unavailable warning for the data it actually displays. Retry against the current generation or classify the superseded result as failed/stale.

markustherobot (commented):
The new status plumbing is coherent in the normal path, but two supported concurrent-auth races can still make the freshness reporting inaccurate. I launched 3 subreviews covering provider/storage concurrency, the CLI/type contract, and regression coverage; all completed, and I verified their candidates against the implementation and focused auth tests.

markusylisiurunen:
@markustherobot address the findings, avoid being overly defensive or over-engineering

src/core/auth/providers/openai_codex.ts
markustherobot: Fixed in e62b336. A failed refresh now compares the reloaded account with the originally attempted credential generation, so a concurrent replacement is treated as current rather than falsely reported stale. Added a regression test for the race.

src/core/auth/providers/openai_codex.ts
markustherobot: Fixed in e62b336. When credential rotation causes the fetched usage update to be discarded, the returned current usage is now marked failed/stale instead of successful. Added a regression test for the race.

PR #393, opened 2026-08-03
add durable searchable session history

### LINKED ISSUE #391: add durable, searchable session history
Tau sessions may be created by local TUIs, remote automation, Telegram bots, and other SDK clients across multiple machines. The current session snapshot is intentionally only the state needed to recover and continue a session, so it is not a complete durable transcript.

Add lightweight transcript history separate from snapshots, together with immutable string-to-string session attributes for provenance and discovery. History should be searchable and readable across Tau hosts, including through an optional self-hosted, single-owner service.

The exact transcript, storage, protocol, and query shapes will be worked out in follow-up discussion.

Supersedes #314 and #352.

markusylisiurunen:
## why

Session snapshots intentionally optimize recovery and continuation, so compaction, rewind, and distributed Tau hosts make them insufficient as a complete searchable record of prior work. Tau needs a separate lightweight transcript history that agents can query across local TUI, automation, and Telegram sessions without coupling history to recovery state.

## what

This adds automatic machine-local transcript capture for every session, with immutable creation attributes, flat user/assistant/tool entries, active-transcript truncation on rewind and retry, SQLite full-text search, bounded pagination, and a durable ordered replication outbox.

Built-in personas now receive a read-only `history` code-mode tool with `search()` and `read()` APIs. Session creation carries required immutable attributes, snapshots persist those attributes and creation time, and the session protocol advances to version 8.

An optional bundled Cloudflare service can be deployed with `tau history setup`. It stores replicated history in D1, authenticates with a regular API key, and adaptively regenerates complete title/summary digests from the active transcript through Cloudflare AI using GPT-5.6 Luna at medium reasoning effort. Without remote configuration, capture and queries remain entirely machine-local.

## details

The transcript retains raw user instruction blocks, committed assistant text including preambles, and completed tool calls with full results while excluding thinking, streaming drafts, and operational state. Compaction never mutates history. Rewind and explicit retry remove the superseded suffix so search and digest generation reflect the active conversation rather than an audit trail.

Remote delivery is asynchronous and idempotent. Tau always commits locally first, bounds replication requests, keeps failed operations in SQLite, and never silently falls back to a different collection when a configured service is unavailable.

A disposable `scripts/import-session-snapshots.js` bootstrap script migrates existing on-disk snapshots directly to the shared service. It applies current snapshot migrations, approximately projects retained messages and completed tools, preserves stable identities and immutable attributes, uploads in bounded idempotent batches with retries, skips malformed inputs, and never modifies source files. It is intentionally not a Tau command or supported migration mode.

fixes #391


src/host/local_session_host.ts
```diff
@@ -2292,6 +2385,7 @@ class LocalHostedSessionHandle implements LocalHostedSession {
             },
             { type: "tool.set", tool: structuredClone(nextTool) },
           ]);
+          this.appendToolHistory(existing, event.historyEntryId, event.message);
```
markustherobot: **[P1] Project terminal tool recovery results into history**

When the provider fails or execution is interrupted after tool admission, `AgentRuntime` emits `tool_recovery` with actual or synthesized terminal results. Only this ordinary `tool_result` path calls `appendToolHistory`; the recovery branch commits state to the snapshot without projecting those results, so `history.read()` permanently omits completed tool calls and full results. Project each recovered result through the same stable, idempotent history path.

src/host/local_session_host.ts
```diff
@@ -2177,6 +2259,11 @@ class LocalHostedSessionHandle implements LocalHostedSession {
           ],
           { persist: true },
         );
+        this.history.append(
```
markustherobot: **[P1] Make transcript projection recoverable**

The snapshot is durably persisted before the independent SQLite append; rewind similarly persists the reset before truncating history. A process exit or SQLite failure in either gap leaves recovery with committed messages missing from history, or a rewound suffix still searchable, because recovery only registers session metadata and never reconciles entries. Add an idempotent projection journal or recovery reconciliation so these two durable stores cannot remain divergent.

src/core/history/history_manager.ts
```diff
@@ -0,0 +1,86 @@
...
+  private scheduleReplication(target: HistoryRemoteTarget): void {
+    this.targets.set(target.endpoint, target);
+    if (this.replicationRuns.has(target.endpoint)) return;
+    const run = this.replicate(target.endpoint).finally(() => {
+      if (this.replicationRuns.get(target.endpoint) === run) {
+        this.replicationRuns.delete(target.endpoint);
+      }
+    });
+    this.replicationRuns.set(target.endpoint, run);
+    void run.catch(() => undefined);
```
markustherobot, line 154: **[P1] Retry a failed outbox drain**

A transient request failure rejects this run, removes it from `replicationRuns`, and is silently swallowed without scheduling another attempt. If the outage occurs on the final turn of a lightly used or long-running host, those durable outbox rows remain undelivered indefinitely until another write or process restart happens to call `scheduleReplication`. Keep a cancellable endpoint-level retry/backoff running while pending rows remain.

src/history/worker/index.ts
```diff
@@ -0,0 +1,715 @@
...
+  const filters: string[] = [];
+  for (const [key, value] of Object.entries(attributes)) {
+    filters.push(
+      "EXISTS (SELECT 1 FROM attributes a WHERE a.session_id = s.session_id AND a.key = ? AND a.value = ?)",
+    );
+    values.push(key, value);
+  }
+  if (query) {
+    filters.push(
+      "(s.digest_title LIKE ? ESCAPE '\\' OR s.digest_summary LIKE ? ESCAPE '\\' OR EXISTS (SELECT 1 FROM entries e WHERE e.session_id = s.session_id AND e.search_text LIKE ? ESCAPE '\\'))",
```
markustherobot: **[P1] Preserve full-text semantics for remote search**

With remote history enabled, a multiword query is treated as one contiguous substring: for example, `durable history` will not match `durable searchable session history`, although the local FTS implementation matches the terms and the tool exposes one query contract. The leading-wildcard `LIKE` also forces corpus-wide scans as history grows. Use a D1 full-text index with the same token semantics as local search, while retaining the digest-first ranking.

src/core/history/local_history_store.ts
```diff
@@ -0,0 +1,466 @@
...
+      }
+
+      if (remote) {
+        const targetInsert = this.database
+          .prepare(
+            "INSERT OR IGNORE INTO history_replication_targets (session_id, endpoint) VALUES (?, ?)",
+          )
+          .run(session.sessionId, remote.endpoint);
+        if (Number(targetInsert.changes) > 0) {
+          this.enqueue(remote.endpoint, {
```
markustherobot, line 87: **[P2] Backfill a newly configured remote target**

When remote history is enabled after local capture has begun, or the endpoint changes, inserting the new target enqueues only `create`. Existing `history_entries` are never emitted, so the new remote collection stays permanently incomplete while only future appends replicate. Queue the current ordered transcript in bounded append operations immediately after the create operation.

src/core/history/history_manager.ts
```diff
@@ -0,0 +1,86 @@
...
+    }
+  }
+}
+
+function boundedReplicationBatch<T extends { operation: unknown }>(pending: T[]): T[] {
+  const batch: T[] = [];
+  let bytes = 0;
+  for (const item of pending) {
+    const itemBytes = Buffer.byteLength(JSON.stringify(item.operation), "utf8");
+    if (batch.length > 0 && bytes + itemBytes > REPLICATION_BATCH_BYTES) break;
```
markustherobot, line 187: **[P2] Do not let one oversized operation poison the outbox**

The byte guard deliberately accepts the first operation even when it already exceeds 6 MiB. A large submitted message or retained full tool result can then exceed the Worker's 8 MiB body limit, fail forever at the head of this ordered queue, and block every later session sharing the endpoint. Split operations by serialized size before enqueueing, or handle an oversized row explicitly so it cannot permanently stall replication.

src/history/worker/index.ts
```diff
@@ -0,0 +1,715 @@
...
+    .prepare(
+      "SELECT position, payload_json FROM entries WHERE session_id = ? AND position > ? ORDER BY position LIMIT ?",
+    )
+    .bind(sessionId, position, limit + 1)
+    .all<{ position: number; payload_json: string }>();
+  const selected = rows.results.slice(0, limit);
+  const last = selected.at(-1);
+  return {
+    session: await descriptor(database, session),
+    entries: selected.map((row) => JSON.parse(row.payload_json)),
```
markustherobot: **[P2] Paginate reads by bytes as well as entry count**

This can return 50 to 100 untruncated full tool results, but `RemoteHistoryClient` rejects any response above 16 MiB. A tool-heavy transcript can therefore make the documented default page unreadable, and because the whole response is rejected the caller receives no cursor with which to advance. Stop a page before the response budget and return a cursor for the remaining entries.

src/core/config/schema.ts
```diff
@@ -1175,6 +1244,19 @@ export function getMistralApiKey(config: Config, env?: NodeJS.ProcessEnv): strin
   return configKey || undefined;
 }
 
+export function getHistoryApiKey(
+  config: HistoryConfig,
+  env?: NodeJS.ProcessEnv,
+): string | undefined {
+  const envName = config.apiKeyEnv?.trim();
```
markustherobot, line 1253: **[P2] Honor the documented history-key precedence**

When both variables are present, this returns the configured `apiKeyEnv` before checking `TAU_HISTORY_API_KEY`, while the README and repository configuration contract say the standard variable has precedence. An operator trying to override an old configured secret can silently authenticate with the wrong key. Check `TAU_HISTORY_API_KEY` first, then the custom environment variable, then the inline value.

markustherobot (commented):
I found material gaps in the new history durability contract and in the remote collection's query and pagination behavior. I launched four read-only subreviews covering local storage/replication, host transcript projection and retry, the Cloudflare service/setup path, and protocol/tool/call-site integration; all four completed, and I independently verified the findings below against the current head.

src/host/local_session_host.ts
markusylisiurunen: Fixed in the current working tree. Recovery tool results now use the normal history projection path after snapshot persistence, with an end-to-end regression test.

src/host/local_session_host.ts
markusylisiurunen: Keeping this best-effort by design. History is non-critical, and the documented non-atomic snapshot/transcript boundary is preferable to adding a cross-store projection journal.

src/core/history/history_manager.ts
markusylisiurunen: Keeping this best-effort by design. Pending rows remain durable and drain on later history activity or session registration; a continuous retry lifecycle is not warranted for the optional remote copy.

src/history/worker/index.ts
markusylisiurunen: Fixed by the subsequent Cloudflare integration changes. Remote search now uses D1 FTS indexes with ANDed token semantics and digest-first ranking instead of leading-wildcard substring scans.

src/core/history/local_history_store.ts
markusylisiurunen: Keeping replication prospective by design. Automatically backfilling a newly selected endpoint would add implicit migration work; retained snapshots can be seeded explicitly with the provided importer when desired.

src/core/history/history_manager.ts
markusylisiurunen: Fixed in the current working tree. Local entries remain complete; remote entries are capped at 1 MiB with explicit middle truncation, and append operations are split below the 6 MiB replication budget. The snapshot importer uses the same projection.

src/history/worker/index.ts
markusylisiurunen: Fixed in the current working tree. Remote reads now treat limit as an upper bound and select entries under a 12 MiB payload budget, returning nextCursor whenever more transcript remains. The code-mode documentation makes the cursor contract explicit.

src/core/config/schema.ts
markusylisiurunen: Fixed in the current working tree. Runtime resolution now checks TAU_HISTORY_API_KEY first, then history.apiKeyEnv, then the inline key, matching the documented contract and snapshot importer. Regression coverage includes both environment variables being present.

markusylisiurunen:
@markustherobot review

markusylisiurunen:
@markustherobot what attributes are added by TUI, Telegram, and other built-in components? SDK probably does not add anything unless user defines them?

src/history/worker/index.ts
```diff
@@ -0,0 +1,1182 @@
...
+    return {
+      sessions: rows.results.slice(0, limit).map((row) => descriptor(row)),
+      ...(rows.results.length > limit
+        ? { nextCursor: encodeCursor({ offset: offset + limit }) }
+        : {}),
+    };
+  }
+
+  const ftsQuery = buildFtsQuery(query);
+  const candidateLimit = Math.min(1_000, offset + limit + 200);
```
markustherobot: **[P1] Paginate matching sessions, not matching rows**

Both FTS queries cap raw matches before deduplicating by session. A single session with more than `candidateLimit` matching entries can fill `transcriptRows`, leaving `ordered` with only that session and no `nextCursor` even though later matching sessions exist; the hard 1,000-row cap also makes all matches beyond that window unreachable. Page distinct session candidates (preferably with a stable keyset/source cursor) and determine exhaustion from the underlying query rather than this truncated merged array.

src/history/worker/index.ts
```diff
@@ -0,0 +1,1182 @@
...
+    sessionId,
+    latest.position,
+    previous,
+  );
+  await env.DB.prepare("DELETE FROM sessions_fts WHERE session_id = ?").bind(sessionId).run();
+  await env.DB.prepare("INSERT INTO sessions_fts (session_id, title, summary) VALUES (?, ?, ?)")
+    .bind(sessionId, digest.title, digest.summary)
+    .run();
+  await env.DB.prepare(
+    "UPDATE sessions SET digest_title = ?, digest_summary = ?, digest_through_entry_id = ? WHERE session_id = ?",
```
markustherobot: **[P1] Reject stale digest writes after transcript changes**

Digest refreshes run concurrently in `waitUntil`: each captures `latest`, waits on an AI call, then unconditionally replaces the FTS row and session digest. If a pre-rewind refresh finishes after the forced post-rewind refresh, it restores a digest containing deleted transcript content and points `digest_through_entry_id` at a removed entry, making superseded text searchable until a later cron repair. Serialize refreshes per session or conditionally commit the FTS/session update only when the captured transcript generation is still current.

markustherobot (commented):
I found material correctness and operational gaps in the new history path, mainly around transcript recovery, remote query/digest consistency, host configuration ownership, and failure handling. I launched five read-only subreviews covering local storage/replication/import, host/runtime/client integration, the D1 worker, the model-facing tool boundary, and deployment/configuration; all five completed, and I independently verified the findings below against the current head.

src/host/local_session_host.ts
```diff
@@ -2335,8 +2464,12 @@ class LocalHostedSessionHandle implements LocalHostedSession {
           };
           this.tools.set(nextTool.id, nextTool);
           changes.push({ type: "tool.set", tool: structuredClone(nextTool) });
+          historyTools.push({ tool: existing, result: toolResult });
         }
         await this.emitPatch("recovery", changes);
+        for (const { tool, result } of historyTools) {
```
markustherobot, line 2453: **[P1] Project restart-synthesized tool results into history**

This covers the live `tool_recovery` event, but process restart uses a different path: `recoverAgentState()` returns `recoveredToolResults`, and `persistRecoveredAgentState()` commits those synthesized results to the snapshot without calling `appendToolHistory`. A crash after a tool call but before its result therefore leaves the recovered snapshot complete while local and remote history permanently omit that tool entry. Project the recovered results idempotently during startup recovery as well.

src/sdk/local_client.ts
```diff
@@ -58,6 +61,8 @@ async function createInProcessSdkHost(
 
   return new LocalSessionHost({
     store: new FileSessionStore({ directory: getDefaultSessionStoreDirectory(home) }),
+    history: HistoryManager.open(getDefaultHistoryDatabasePath(home)),
+    historyRemote: resolveHistoryRemoteTarget(config),
```
markustherobot, line 65: **[P1] Preserve host history config for Telegram workspaces**

`createTauSdkClient` loads this `config` from `options.cwd`, while Telegram sets that cwd to each generated workspace. For the documented `/var/lib/...` workspace roots, Tau intentionally excludes the home-level config; because `history` is global-only, `resolveHistoryRemoteTarget(config)` then returns undefined and Telegram sessions silently never use the configured shared history service. Pass the runner's already-loaded host config (or its resolved history target) separately while continuing to resolve agent-visible runtime config from the workspace.

src/core/history/setup.ts
```diff
@@ -0,0 +1,213 @@
...
+  }
+}
+
+export async function destroyHistoryService(options: HistoryDestroyOptions): Promise<void> {
+  if (!options.yes) throw new Error("tau history destroy requires --yes");
+  const env = options.env ?? process.env;
+  const stdout = options.stdout ?? console.log;
+  requireCloudflareAuth(env);
+
+  const worker = await runWrangler(["delete", "--name", WORKER_NAME], { env, allowFailure: true });
```
markustherobot: **[P1] Fail when history resources are not deleted**

Both teardown calls use `allowFailure: true` and their results are only printed, so authentication, permission, network, or Wrangler failures still make `tau history destroy --yes` exit successfully. Automation can report a completed teardown while the Worker or billable D1 database remains (and the second delete may leave a surviving Worker without its database). Continue attempting both resources, but aggregate non-not-found failures and return a nonzero exit.

src/core/telegram/session_manager.ts
```diff
@@ -764,6 +776,7 @@ class TelegramSessionManagerImpl implements TelegramSessionManager {
     entry.unsubscribeClientEvents = tauSession.onDelta((event) => {
       this.handleClientEvent(entry, event);
     });
+    this.handleSnapshotNotices(entry, await tauSession.snapshot());
```
markustherobot, line 838: **[P2] Replay recovered notices after adapters subscribe**

At runner startup, `startTelegramRuntime` awaits `sessionManager.initialize()` before constructing any Telegram adapter. This reconnect path therefore calls `handleSnapshotNotices()` with no listeners, marks each ID in `emittedNoticeIds`, and drops the event; after adapters subscribe, those persisted warnings can never be emitted again. Buffer/replay recovered notices after adapter registration, or move this replay to a point where chat routing and listeners already exist.

src/core/history/local_history_store.ts
```diff
@@ -0,0 +1,466 @@
...
+      .prepare("SELECT *, NULL AS snippets FROM history_sessions WHERE session_id = ?")
+      .get(input.sessionId) as SqlRow | undefined;
+    if (!sessionRow) {
+      throw new Error(`history session '${input.sessionId}' was not found`);
+    }
+    const rows = this.database
+      .prepare(
+        "SELECT position, payload_json FROM history_entries WHERE session_id = ? AND position > ? ORDER BY position LIMIT ?",
+      )
+      .all(input.sessionId, position, input.limit + 1) as SqlRow[];
```
markustherobot: **[P2] Bound local read pages before bridging them**

Local reads cap only entry count, so the default 50 (or allowed 100) complete tool results can deserialize and cross the code-mode bridge as tens or hundreds of megabytes before generated code projects them. That can exceed the SES worker's 128 MiB heap and make a valid local transcript unreadable, while the remote path already stops near 12 MiB and returns a cursor. Apply a cumulative payload-byte budget locally and preserve `nextCursor` when the requested count is not reached.

scripts/import-session-snapshots.js
```diff
@@ -0,0 +1,274 @@
...
+    }
+  }
+
+  const tools = new Map(Object.values(snapshot.tools).map((tool) => [tool.toolCallId, tool]));
+  const entries = [];
+  for (const wrapped of snapshot.messages) {
+    if (wrapped.state !== "committed" && wrapped.state !== "interrupted") continue;
+    const message = wrapped.message;
+    if (message.role === "user") {
+      entries.push(userHistoryEntry(wrapped.id, message));
```
markustherobot, line 131: **[P2] Exclude Tau-internal hidden user records from imports**

The live projection does not append synthetic tool-recovery or auto-compaction-continuation user messages, but the importer converts every retained user message. `userHistoryEntry()` strips their Tau metadata while preserving the leading `<system>` payload, so imported history exposes internal recovery/continuation instructions as if they were user transcript entries. Detect the existing `tool-recovery` and `auto-compaction-continuation` metadata and skip those records while retaining genuine user-authored system blocks.

src/core/history/setup.ts
```diff
@@ -0,0 +1,213 @@
...
+            apiKeyEnv: "TAU_HISTORY_API_KEY",
+          },
+        },
+        null,
+        2,
+      ),
+    );
+    stdout("");
+    stdout("Set TAU_HISTORY_API_KEY to:");
+    stdout(apiKey);
```
markustherobot: **[P2] Do not echo caller-supplied API keys**

This prints the bearer secret even when it came from `--api-key` or an existing `TAU_HISTORY_API_KEY`, which can leak access to all replicated transcripts through CI logs or captured setup output. Track whether the key was generated by this invocation and print it only in that case; for supplied credentials, print the configuration instruction without reproducing the secret.

src/history/worker/index.ts
markusylisiurunen: Fixed in the current working tree. Remote FTS search now groups digest and transcript matches by session before applying `LIMIT`/`OFFSET`, preserves digest-first and recency ordering, and loads bounded snippets only for selected sessions. A real SQLite FTS regression covers one session contributing 250 matching rows without hiding later matching sessions or losing `nextCursor`. `npm run check`, the build, and all 21 focused history tests pass.

src/history/worker/index.ts
markusylisiurunen: Fixed in the current working tree. The canonical v1 D1 schema now includes a persisted `transcript_revision`, effective appends and truncations increment it in their operation batch, and digest generation captures the revision with the latest entry. Digest metadata and `sessions_fts` now update in one conditional batch only when that revision is still current, so stale model results are discarded across Worker isolates. A real SQLite race regression completes the newer digest first and verifies the delayed older result cannot overwrite it. `npm run check`, the build, and all 22 focused history tests pass.

src/host/local_session_host.ts
markusylisiurunen: Fixed in the current working tree. Restart recovery now finalizes each recovered protocol tool, persists the repaired snapshot, then projects the synthesized result through the existing idempotent `appendToolHistory` path. The host recovery regression now asserts both dangling-running and already-finished tool calls appear in SQLite history with their repaired outcomes. `npm run check`, the build, and 95 focused history/host tests pass.

src/sdk/local_client.ts
markusylisiurunen: Fixed in the current working tree. Telegram now creates each in-process SDK client with the runner’s already-resolved Tau config via `createTauSdkClientWithHostConfig`, so host-owned history and infrastructure remain stable across generated workspaces. `resolveSessionBootstrap` still resolves personas, prompts, skills, models, and project settings from each execution-environment cwd. `npm run check`, the build, and 54 focused Telegram/SDK integration tests pass.

src/core/history/setup.ts
markusylisiurunen: Fixed in the current working tree. History teardown now attempts Worker and D1 deletion independently, treats only resource-specific not-found diagnostics as already absent, accumulates all other command/spawn failures, and throws one detailed `AggregateError` after both attempts so the CLI exits nonzero on partial cleanup. Per discussion, I did not add a brittle test coupled to Wrangler’s unstable human-readable diagnostics. `npm run check`, the build, and all 22 focused history tests pass.

src/core/telegram/session_manager.ts
markusylisiurunen: Fixed in the current working tree with an adapter-local replay rather than a new manager/runtime protocol. Each adapter subscribes to live events, then best-effort reads only its restored session mappings and queues persisted warning/error notices through the existing ordered notification path; snapshot-read failures only log a warning. A focused adapter regression covers replay after subscription. `npm run check`, the build, and 88 focused Telegram tests pass.

src/core/history/local_history_store.ts
markusylisiurunen: Fixed in the current working tree. Local reads now select positions and SQLite payload byte lengths first, apply the same 12 MiB cumulative page budget as remote reads, then deserialize only the selected complete entries and return `nextCursor` whenever more remain. Stored payloads remain unmodified, and one oversized entry is still allowed for forward progress. The code-mode contract now describes byte-bounded pages generally, with a regression covering continuation across complete 1 MiB local entries. `npm run check`, the build, and all 23 focused history tests pass.

scripts/import-session-snapshots.js
markusylisiurunen: Keeping this behavior by design. The snapshot importer intentionally projects every retained `role: "user"` record, including Tau-generated tool-recovery and auto-compaction-continuation records whose content consists only of hidden `<system>` blocks. More generally, system-only user messages must remain importable; imported historical content is already treated as untrusted evidence by the tool and digest boundaries. No code change.

src/core/history/setup.ts
markusylisiurunen: Fixed in the current working tree. Setup now distinguishes caller-supplied credentials from a newly generated key: generated keys are printed once for bootstrap, while `--api-key` and existing `TAU_HISTORY_API_KEY` values receive only a non-secret configuration reminder. Wrangler still receives the selected secret through stdin. `npm run check`, the build, and all 23 focused history tests pass.

markusylisiurunen:
Follow-up to the earlier creation-attribute clarification: the current working tree broadens the conventional `repository` attribute while keeping creation attributes fully client-owned and authoritative.

- Local `tau` TUI startup now supplies `source: "tui"` plus a conventionally normalized `repository` value from the current Git repository. When the workspace root is not itself a repository, it inspects direct child repository roots and supplies their normalized origins as one comma-delimited string.
- TUI `/new` keeps `source: "tui"` and copies a known `repository` value because it reuses the same execution environment.
- `tau attach --new` and prompted attach creation still supply only `source: "tui"`; attach does not provision or know the referenced execution environment's repositories.
- Telegram repository projects supply `source`, `project`, and normalized `repository`. Composite projects join member repositories with commas in configured `projectIds` order.
- SDK and raw RPC/WebSocket clients remain unchanged: their caller supplies the complete attribute map, and `{}` remains valid.
- Recovery and snapshot import preserve stored attributes unchanged. The host, history manager, stores, and replication do not infer, normalize, merge, sort, or parse attribute values.

History attribute filters now accept either an exact string or `{ contains: string }` for ordinary case-sensitive substring matching. The `history` tool documentation describes `source` and `repository` as optional conventional, untrusted attributes and shows repository substring filtering for both single-repository and composite sessions.

The full suite passes: 76 test files and 927 tests.

PR #394, opened 2026-08-03
serialize session snapshot mutations

### LINKED ISSUE #392: preserve and expose the initiating error when a Telegram turn fails mid-tool
## summary

A Telegram-driven session failed internally immediately after the model emitted a `create_goal` call followed by a `bash` call. The Telegram manager then stopped its SDK client, which interrupted the active Tau turn. The resulting snapshot records only the secondary abort, leaves both tools nonterminal, and does not retain the exception that initiated the failure.

The sole Telegram user did not send `/interrupt`, `/new`, or any other concurrent command. There was only one active Telegram session for that user. The adapter process remained alive throughout.

The original exception is no longer recoverable because the Telegram manager keeps its diagnostic only in an in-memory session log and excludes failed sessions from its persisted session registry. This makes the actual root cause impossible to debug after the fact.

## environment

- Tau package: `@markusylisiurunen/tau@0.3.36`
- Node.js: `v24.18.1`
- Platform: Linux 6.8.0-136-generic x86_64
- Telegram adapter process started: 2026-08-03 09:24:29 UTC
- Failure: 2026-08-03 19:40:01 UTC
- Tau session ID: `cc7fb736-0197-4524-820e-4a416f7542cf`
- Snapshot file: `~/.config/tau/sessions/Y2M3ZmI3MzYtMDE5Ny00NTI0LTgyMGUtNGE0MTZmNzU0MmNm.json`

## request that triggered it

The user asked the agent to:

1. Post a follow-up comment on issue #391 describing the design decisions reached in the conversation.
2. Create a persisted goal.
3. Implement the design through verification and opening a PR.

The assistant emitted a short acknowledgement and one model response containing these tool calls in order:

1. `create_goal`, with an objective covering durable searchable session history.
2. `bash`, running `gh issue comment 391 --body-file - ...`.

## observed persisted state

The final assistant message is persisted with:

- `state: "interrupted"`
- `stopReason: "aborted"`

Its tools are persisted as:

- `create_goal`: `status: "running"`, with `startedAt: 1785786001474`
- `bash`: `status: "queued"`

Neither tool has a result message. The `bash` command never ran, and issue #391 received no follow-up comment.

The goal mutation itself was persisted, but its status became `blocked`:

```json
{
  "objective": "Implement durable searchable Tau session history ... and open a pull request.",
  "status": "blocked"
}
```

The session lifecycle is `idle`, despite the semantic tool records still saying `running` and `queued`.

The final usage record at the failure timestamp contains zero input/output tokens, consistent with teardown after the model had already emitted the tool calls rather than a new model failure.

## established causal chain

Inspection of the current Telegram/session-host code and persisted artifacts establishes this sequence:

1. `TelegramSessionManagerImpl.submitText()` awaited `tauSession.submit(payload)`.
2. The submission rejected while the tool turn was in progress.
3. The manager's catch path marked the Telegram session `failed` and called `stopClient()`.
4. `stopClient()` called `tauSession.interrupt()` before unobserving and closing the SDK client.
5. That secondary interrupt produced the persisted assistant `state: "interrupted"` and `stopReason: "aborted"` and caused the newly active goal to become `blocked`.
6. Tool lifecycle reconciliation did not move the active `create_goal` or queued `bash` call to terminal cancellation states.

The abort is therefore a consequence of the manager's failure cleanup, not evidence of a user-initiated interrupt.

## missing root-cause evidence

The initiating exception cannot be determined from durable state:

- `submitText()` stores the exception on `entry.record.error` and in `entry.logs`.
- `entry.logs` are memory-only.
- `persistSessions()` serializes only entries whose Telegram state is in `ACTIVE_STATES`; once the entry becomes `failed`, it disappears from `<workspaceRoot>-sessions.json`.
- The Tau snapshot persists the secondary abort but not the submission exception or failed turn diagnostic.
- The tmux output did not contain the initiating exception.

Consequently, it is not possible to tell whether the original rejection came from goal mutation persistence, runtime event handling, snapshot emission, protocol handling, or another boundary.

## expected behavior

When a Telegram submission fails during a turn:

- Persist or otherwise durably log the initiating exception, including its cause chain and the session/public-session IDs, before stopping the client.
- Clearly distinguish an internally generated cleanup interrupt from a user-requested interrupt.
- Reconcile every started or queued tool to a terminal state such as `cancelled` when teardown interrupts the turn.
- Preserve a failed Telegram session record, or a bounded failure tombstone, long enough for postmortem debugging.
- Emit a useful adapter error to stderr/logging even if the failed session is removed from the active registry.

## investigation pointers

Relevant paths in the current tree:

- `src/core/telegram/session_manager.ts`
  - `submitText()` catch path
  - `stopClient()`
  - `persistSessions()` filtering to active states
  - in-memory `entry.logs`
  - `executeSubmit()` converts `runTurn()` exceptions into `failed to run session turn`
  - `snapshotAfterFailedSubmit()`
  - connection/session interruption paths
- `src/host/local_session_host.ts`
  - `runTurnNow()` failure cleanup and goal blocking
  - `interruptActiveWork()`
- Agent/runtime tool lifecycle handling for queued and running tools when a turn is aborted during failure cleanup

A regression test should inject a host/runtime failure after a tool has entered `running` and another call is `queued`, then assert that:

1. The initiating diagnostic remains recoverable.
2. The Telegram cleanup abort is identified as secondary/internal.
3. Both tools become terminal.
4. The goal is blocked without losing the failure context.
5. The failed session or a failure tombstone remains available after manager persistence/restart.

markustherobot:
## why

A model-driven goal mutation could persist concurrently with later transient assistant and tool projections. When the slower durable write completed, it reinstalled an older snapshot, emitted revisions out of order, and caused the next tool lifecycle event to fail validation. Telegram then replaced the initiating failure with a secondary cleanup interrupt and discarded the failed session record, leaving the original diagnostic unrecoverable.

## what

The local session host now has one per-session mutation queue for runtime events, goal and configuration changes, snapshots, resets, and transient projections. Durable candidates are persisted before replacing the live snapshot, and failed mutations restore host projection state from the last committed snapshot. Failed turns retain the initiating cause chain, block active goals, and reconcile queued or running tools to terminal cancellation states.

Telegram submission failures now persist the public session ID, Tau session ID, complete protocol diagnostic, and failed state before client cleanup. Cleanup interrupts are logged explicitly as internal, the adapter reports the failure through its process logger, and failed records survive restart as tombstones without reconnecting a client.

## details

The Telegram session registry advances to version 2 so state and failure diagnostics are canonical persisted fields. Version 1 registries are normalized on read and rewritten in version 2. Failed tombstones are retained until the session is explicitly closed, which provides a bounded operator-controlled postmortem window without adding a second diagnostics store.

No unresolved implementation questions remain.

fixes #392


src/host/local_session_host.ts
```diff
@@ -1937,6 +1899,27 @@ class LocalHostedSessionHandle implements LocalHostedSession {
...
+  private enqueueMutation<T>(mutation: () => Promise<T> | T): Promise<T> {
+    const result = this.mutationQueue
+      .catch(() => undefined)
+      .then(async () => {
+        try {
+          return await mutation();
+        } catch (error) {
+          if (this.committedSnapshot) {
+            this.goal = structuredClone(this.committedSnapshot.goal);
+            this.restoreProtocolState(this.committedSnapshot);
```
markustherobot, line 1916: **[P1] Reconcile finalized drafts before rollback**

When the store write for `assistant_final` fails, `AgentRuntime` has already appended the final assistant history entry and this handler has already cleared the draft, but this rollback restores the older draft from `committedSnapshot`. `cleanupFailedTurn()` then cannot commit the interrupted assistant because that history ID already exists, and the next snapshot contains both the final history entry and restored draft with the same ID. Snapshot validation rejects that duplicate, so cleanup cannot persist the failed outcome or terminal tools and the original error can be replaced by the validation failure. Reconcile restored drafts against runtime history, or avoid restoring a draft whose ID has already finalized.

src/host/local_session_host.ts
```diff
@@ -907,8 +911,7 @@ class LocalHostedSessionHandle implements LocalHostedSession {
         await this.session.commitUserText(buildGoalContinuationText(this.goal));
       } catch (error) {
         this.rejectCommittedLogicalSteering(committedSteering, error);
-        await this.cleanupFailedTurn().catch(() => undefined);
-        await this.blockActiveGoal().catch(() => undefined);
+        await this.cleanupFailedTurn(rootUserMessage.id, error).catch(() => undefined);
```
markustherobot: **[P2] Persist failure outcomes for committed steering**

If a goal-bound steering message has been committed and a later model, tool, or event-sink failure reaches this catch, `committedSteering` contains its history ID but cleanup records the failed outcome only on the root message. The success path assigns an outcome to every committed steering message, while this path leaves the accepted steering message without a terminal `turn` and rejects its result promise. Pass the committed steering IDs into cleanup and apply the same failed outcome to them.

markustherobot (commented):
I found two failure-recovery gaps that can still leave accepted work without a coherent durable projection; details are inline. I launched 4 subreviews covering host mutation serialization, failed-turn recovery, Telegram tombstones, and cross-module integration/regression coverage; all completed.

markusylisiurunen:
@markustherobot address the findings, but avoid over-engineering or introducing additional complexity for very rare or edge-case issues that almost never happen in the real world.

src/host/local_session_host.ts
markustherobot: Fixed in e3694a7. Rollback restoration now skips a draft when its history ID has already finalized in runtime history, preventing the duplicate final-plus-draft snapshot. The regression forces the assistant-final persistence write to fail and verifies cleanup remains durable.

src/host/local_session_host.ts
markustherobot: Fixed in e3694a7. Failed-turn cleanup now applies the same terminal failed outcome to the root message and every steering message committed for that continuation. The regression injects an event-sink failure after buffered steering is accepted and verifies its durable outcome.

markusylisiurunen:
Reviewed the two remaining recovery edge cases. Both require a persistence failure at a precise event boundary and only affect minor cost accounting or diagnostic presentation, so we’re intentionally leaving them unchanged.

PR #398, opened 2026-08-05
align history worker with Cloudflare limits

markusylisiurunen:
## why

The history Worker exposed a search limit that exceeded D1's maximum bound-parameter count, trusted client-side entry projection instead of enforcing D1 row constraints at the service boundary, and performed schema migrations during Worker requests. Append replication also spent one D1 round trip per entry checking for duplicates.

## what

Align the service contract and deployment flow with Cloudflare's platform boundaries. Remote history search now caps pages at 75 sessions, remote entries are validated at one serialized MiB, indexed entry text is UTF-8 byte-bounded to 512 KiB, and append operations load all existing entry ids with one bounded query.

Replace the Worker's isolate-local runtime migration system with Wrangler D1 migrations applied before every deployment by `tau history setup`, explicitly enable Workers Logs, and update the history documentation and code-mode contract.

## details

Setup applies the schema before every Worker deployment. The initial official migration adopts the current v3 schema idempotently and retains `history_schema_migrations`, keeping the currently deployed Worker compatible until replacement while making interrupted fresh-setup retries safe.


src/core/history/migrations.ts
```diff
@@ -0,0 +1,58 @@
...
+export const HISTORY_INITIAL_MIGRATION_SQL = `CREATE TABLE IF NOT EXISTS sessions (
+  session_id TEXT PRIMARY KEY,
+  attributes_json TEXT NOT NULL,
+  created_at INTEGER NOT NULL,
+  updated_at INTEGER NOT NULL,
+  transcript_revision INTEGER NOT NULL DEFAULT 0,
+  digest_title TEXT,
+  digest_summary TEXT,
+  digest_through_entry_id TEXT,
+  digest_failure_count INTEGER NOT NULL DEFAULT 0,
```
markustherobot, line 12: **[P1] Upgrade released legacy schemas before adopting the baseline**

Upgrading a service created by released v0.3.38 (legacy migration v1), or a database left at v2, keeps the old `sessions` shape because `CREATE TABLE IF NOT EXISTS` does not add these five columns. This migration then drops `history_schema_migrations` and Wrangler records `0001_initial.sql` as applied, while the scheduled Worker reads `digest_failure_count`, so digest processing fails permanently and rerunning setup cannot repair it. Please explicitly complete the legacy v1/v2 upgrades before deleting their marker, with regression tests starting from each legacy schema.

src/core/history/setup.ts
```diff
@@ -40,12 +41,18 @@ export async function setupHistoryService(options: HistorySetupOptions): Promise
...
-    writeWorkerProject(temporaryDirectory, { domain, zoneName, databaseId });
+    writeWorkerProject(temporaryDirectory, {
+      domain,
+      zoneName,
+      databaseId: database.id,
+    });
+    if (database.created) await applyHistoryMigrations(temporaryDirectory, env, stdout);
     stdout(`deploying Worker ${WORKER_NAME} with Wrangler...`);
     await runWrangler(["deploy"], { cwd: temporaryDirectory, env });
+    if (!database.created) await applyHistoryMigrations(temporaryDirectory, env, stdout);
```
markustherobot: **[P2] Check schema readiness before choosing deployment order**

`database.created` only describes this invocation. If setup creates D1 and then its pre-deploy migration fails or the process is interrupted, the retry sees an existing database, deploys the migration-free Worker on the empty or partial schema, and only then attempts migration. Requests and cron fail between those steps, and another migration failure leaves the deployed service unusable. Base the ordering on actual migration readiness (or always establish a compatible schema before deploy), and cover a failed-first-attempt retry in the setup test.

markustherobot (commented):
Found two material migration-lifecycle hazards that can leave the deployed history Worker on an incompatible schema. I launched three read-only subreviews covering deployment and migration safety, Worker limits and append semantics, and cross-layer contracts and tests; all three completed. The search, byte-boundary, and duplicate-query changes otherwise look aligned.

src/core/history/migrations.ts
markusylisiurunen: Direct v1/v2 upgrades are intentionally out of scope for this single-owner service; the only deployment is verified at v3. The legacy marker now remains in place for a safe migration-first cutover.

src/core/history/setup.ts
markusylisiurunen: Setup now always applies migrations before deployment while retaining the legacy marker. Added coverage for retrying after a failed initial migration.

PR #401, opened 2026-08-06
allow disabling Codex accounts

### LINKED ISSUE #400: allow disabling a codex account from selection
it should still stay registered and e.g. `tau auth list` should refresh/display it, but it should just never be selected for an agent session.

markustherobot:
## why

Authenticated Codex accounts sometimes need to remain registered for usage visibility and credential refresh without being eligible for agent sessions. Removing an account loses that visibility, while forcing selection rules alone cannot prevent automatic failover or a session's sticky selection from continuing to use it.

## what

I added account-scoped `tau auth disable codex --account <email-or-id>` and `tau auth enable codex --account <email-or-id>` commands. Disabled state is persisted on the OAuth account, shown by `tau auth list`, and preserved when the same account authenticates again.

Disabled accounts continue through list-time identity and usage refreshes, but are excluded from automatic prioritization, failover, and sticky session selections. `TAU_CODEX_ACCOUNT` now reports an explicit error when it resolves to a disabled account, and selection rechecks the state after asynchronous credential work so a concurrent disable cannot return credentials.

## details

Existing `auth.json` OAuth entries are normalized to `disabled: false` at the storage boundary, keeping the runtime contract explicit without requiring users to re-authenticate. I treated email and account-id matching as identical to logout, per the issue discussion. There are no remaining implementation questions.

fixes #400


src/core/auth/providers/openai_codex.ts
```diff
@@ -224,7 +253,7 @@ export class OpenAICodexAdapter implements AuthProviderAdapter {
     options?: { apiKey?: string },
   ): Promise<boolean> {
     const account = getAccounts(authStorage).find((entry) => entry.accountId === accountId);
-    if (!account) return false;
+    if (!account || account.disabled) return false;
```
markustherobot, line 256: **[P1] Recheck disabled state after sticky usage refresh**

When a sticky account's usage is missing or expired, this guard runs before the awaited `getUsageSnapshot()`. If another process disables the account while that request is pending, the usage update preserves `disabled: true`, but this method still returns `true` from usage alone and `selectCodexAccount()` returns the disabled credential. Re-read the current account after the await and return false when it is missing or disabled.

src/core/auth/auth_manager.ts
```diff
@@ -54,24 +59,35 @@ export class AuthManager {
...
     const removed = adapter.removeAccount(this.authStorage, accountId);
     if (!removed) {
       throw new Error(`account "${accountId}" not found for provider "${providerId}"`);
     }
   }
+
+  setAccountEnabled(providerId: string, accountId: string, enabled: boolean): void {
+    const adapter = this.getAdapter(providerId);
+    this.authStorage.reload();
+    const updated = adapter.setAccountEnabled(this.authStorage, accountId, enabled);
```
markustherobot, line 84: **[P1] Preserve invalid auth storage on account-state changes**

If `auth.json` is malformed or has an unsupported shape, `reload()` exposes an empty provider map plus `invalidReason`. Calling the adapter then enters `AuthStorage.update()`, which saves that empty map and clears the error before this method reports `account not found`, so the new enable/disable commands destroy the original credential file. Check `getInvalidReason()` after reloading and fail before mutation, with a regression test that verifies the file remains unchanged.

markustherobot (commented):
I found two correctness issues around the new disabled-account boundary: one in concurrent selection and one in invalid-storage mutation. I launched 3 subreviews covering storage/concurrency, account-selection semantics, and CLI/contracts; all completed successfully.

markusylisiurunen:
@markustherobot fix the findings if they are relevant and should be fixed

src/core/auth/providers/openai_codex.ts
markustherobot: Fixed in a8cdace. `isAccountUsable()` now rechecks the current enabled state after the awaited usage refresh, with a regression covering sticky selection disabled during that refresh.

src/core/auth/auth_manager.ts
markustherobot: Fixed in a8cdace. Account state changes now stop on `AuthStorage.getInvalidReason()` before mutation, with a regression verifying malformed `auth.json` content remains unchanged.

PR #402, opened 2026-08-06
unify feedback and compaction presentation

markusylisiurunen:
## why

Tau's feedback surfaces had accumulated mixed system-message behavior, inconsistent wording and tone, and duplicate provider failure rendering. Automatic compaction also relied on an accidental visual result: retained model context stayed visible only because the TUI continued reconciling transcript records that should have become immutable client-local history.

## what

This introduces one structured feedback grammar across runtime, protocol, host, SDK, Telegram, and TUI boundaries. Ephemeral activities and notices now use explicit presentation contracts, durable timeline notices carry structured titles, content, severity, and ownership, and provider failures remain recoverable without duplicating inline error output. Requested assistant-turn interruptions are also persisted as host-owned notices so every observer and later recovery sees the same outcome. The TUI renders concise lowercase feedback through dedicated footer and transcript components with semantic feedback colors and bounded presentation-only truncation.

Automatic compaction now creates an explicit transcript projection. A live TUI freezes the prior segment under a `compacted context` boundary, derives the retained range from canonical summary and continuation metadata, and permanently removes frozen messages, notices, and tool cards from active reconciliation. Clients attaching after compaction render only the active projection, and repeated compactions preserve every earlier frozen segment.

## details

The session protocol moves to version 9 and stored session documents to version 6. The storage migration converts legacy timeline notices at the recovery boundary, keeping the current runtime contract canonical while preserving older durable sessions.


src/tui/ui/transcript_notice.ts
```diff
@@ -0,0 +1,69 @@
...
+      this.addChild(new Text(this.theme.palette.textDim(content.join("\n")), 1, 0));
+    }
+  }
+}
+
+export function projectTranscriptNoticeContent(content: readonly string[]): string[] {
+  const lines = content.flatMap((value) =>
+    value.replace(/\r\n?/g, "\n").split("\n").flatMap(splitTranscriptNoticeLine),
+  );
+  if (lines.length <= TRANSCRIPT_NOTICE_CONTENT_MAX_LINES) return lines;
```
markustherobot, line 44: **[P1] Preserve the complete `/help` output**

`showHelp()` now routes the multiline command/keybinding reference through this projection, so every `/help` response is reduced to three leading and three trailing content lines. Most commands are therefore no longer visible. Keep the bounded projection for diagnostics, but render help through an unbounded model or make this cap opt-in.

src/core/session/runner.ts
```diff
@@ -753,7 +752,12 @@ async function* runPreparedToolCall(
   now: () => number = Date.now,
 ): AsyncGenerator<ToolRunnerEvent, CompletedToolRun, void> {
   if (prepared.type === "rejected") {
-    yield { type: "notice", severity: "error", text: prepared.message };
+    yield {
+      type: "feedback",
+      tone: "error",
+      title: prepared.message,
```
markustherobot: **[P2] Keep malformed tool names from aborting the turn**

An unavailable tool name comes directly from model output and is interpolated into this title. If it contains `\n` or `\r`, the new one-line protocol title validator makes `createSessionProtocolEphemeralMessage()` throw, aborting the turn instead of committing the blocked tool result. Use a fixed one-line title and move the model-controlled name into `content` (and avoid the same interpolation in execution-failure titles).

src/tui/ui/footer.ts
```diff
@@ -145,52 +148,98 @@ export class FooterComponent implements Component {
     if (width <= 0) return [""];
 
     const { palette } = this.theme;
-    const isWorking = this.iconIntervalId !== null;
+    if (this.status?.type === "activity") {
```
markustherobot, line 159: **[P2] Show notices while an activity is displayed**

This branch returns before `this.notice` is checked, so a timed notice raised during compaction, diff review, transcription, or another activity expires without ever being shown. For example, submitting input during manual compaction calls `showFooterNotice("wait for tau to become idle...")`, but the footer continues to show only `compacting context`. Give notices precedence or defer their timeout until the activity ends.

src/tui/session_chat_controller.ts
```diff
@@ -742,28 +757,33 @@ export class SessionChatController {
...
 
+    this.assistantInterruptRequested = true;
     try {
       const result = await this.session.interrupt();
-      this.view.addSystemMessage(
-        result.interrupted ? "interrupted" : "interrupt requested",
-        "error",
-      );
+      if (result.interrupted) {
+        this.view.addTranscriptNotice("assistant turn interrupted", "default");
```
markustherobot: **[P2] Expose interruptions to every attached client**

This notice is added only to the TUI that issued `session.interrupt`, while the assistant renderer no longer displays the snapshot's durable `interrupted` state. Other attached clients, and clients opening the recovered session, now see only partial assistant text (or nothing when no text streamed) with no indication that the turn was cancelled. Project interruption from synchronized state or publish a durable/protocol-visible notice.

src/host/local_session_host.ts
```diff
@@ -2320,6 +2343,10 @@ class LocalHostedSessionHandle implements LocalHostedSession {
           cost: { total: cost },
           agent: { type: "main" },
         });
+        const failureNotice = createModelFailureTimelineItem(event.historyEntryId, event.message);
```
markustherobot: **[P2] Avoid sending two Telegram errors for one provider failure**

For Telegram sessions this new durable notice is delivered as a detailed `session-notice`, then `TelegramSessionManager.submitText()` emits `session-turn-failed` for the same failed turn and the adapter sends `turn failed. please try again.` Users now receive two ordered notifications for every provider failure. Suppress the generic turn-failed notification when the detailed model-failure notice was emitted, or consolidate them into one event.

docs/rpc.md
```diff
@@ -190,7 +190,7 @@ params (required):
 
 ```json
 {
-  "version": 8,
+  "version": 9,
```
markustherobot, line 193: **[P2] Update the initialize result version too**

The response envelope was bumped to 9 here, but the same example still returns `"protocolVersion": 8` inside `result` even though the implementation returns `SESSION_PROTOCOL_VERSION` (9). A client following this canonical initialize example can validate against the wrong protocol version; update the nested value as part of the bump.

docs/sdk.md
```diff
@@ -445,6 +445,8 @@ const unsubscribePendingUserMessages = session.onPendingUserMessages(
 const unsubscribeEphemeral = session.onEphemeral((message) => {
   if (message.event.type === "ephemeral-agent.thread-update") {
     console.log(message.event.threadId, message.event.update.lastActivityText);
+  } else if (message.event.type === "feedback.notice") {
+    console.log(message.event.tone, message.event.title, message.event.content);
```
markustherobot: **[P2] Narrow the feedback variant before reading content**

This new TypeScript example does not typecheck: narrowing only on `event.type === "feedback.notice"` leaves both footer and transcript variants, and the footer variant has no `content` property. Check `event.presentation === "transcript"` before reading `content`, or omit that property from the generic notice example.

docs/session-event-protocol-redesign.md
```diff
@@ -657,7 +659,7 @@ Replace the same draft assistant message with the committed final assistant mess
           "id": "notice-1",
           "notice": {
             "severity": "warn",
-            "text": "auto-retrying after transient error",
+            "text": "retrying after transient error",
```
markustherobot: **[P2] Replace the obsolete notice shape in this design document**

This PR changes timeline notices to require `title` and `subject` with optional `content`, but the document still defines `SessionNotice` as `{ severity, text, timestamp }` and this changed example still emits `text`. That now contradicts protocol v9 and the updated RPC documentation. Update both the type definition and example to the canonical structured notice contract.

markustherobot (commented):
Found material issues around feedback visibility, malformed model output, multi-client behavior, Telegram delivery, and public protocol documentation. I launched five focused subreviews covering protocol/storage, host/runtime projection, compaction reconciliation, TUI presentation, and integrations/docs; all five completed.

src/tui/ui/transcript_notice.ts
markusylisiurunen: ah, I forgot help. it doesn't honestly feel like it should be a notice. maybe we need some kind of text component that can be rendered on the timeline nicely instead of trying to use a notice for this

src/tui/ui/footer.ts
markusylisiurunen: precedence for notices that come during actice work

src/tui/session_chat_controller.ts
markusylisiurunen: I believe that first of all this interrupted notice should be host-emitted and it should also be of the durable kind, stored in the snapshot. I believe we have that concept already. I think this should solve this findings. let me know if I'm wrong.

src/host/local_session_host.ts
markusylisiurunen: should only send the specific one. one more thing: we only want to send the notice title and never content to telegram

markusylisiurunen:
@markustherobot address/fix the review findings

src/tui/ui/transcript_notice.ts
markustherobot: I added a dedicated unbounded muted transcript-text component and routed `/help` through it, so the complete command and keybinding reference remains visible. The bounded notice projection remains limited to diagnostics.

src/core/session/runner.ts
markustherobot: I replaced model-controlled tool feedback titles with fixed one-line titles and moved tool names into content. Regression coverage includes a CRLF-bearing unavailable tool name passing through the host protocol without aborting the turn.

src/tui/ui/footer.ts
markustherobot: I gave notices precedence over active-work status in both footer rendering and animation selection. When the notice expires, the active activity resumes normally.

src/tui/session_chat_controller.ts
markustherobot: I moved assistant-turn interruption feedback to the host interrupt path as a durable session-subject timeline notice and removed the initiating TUI copy. This covers zero-output interruptions, all attached observers, and recovered snapshots.

src/host/local_session_host.ts
markustherobot: I changed Telegram notice delivery to send only the notice title. A detailed error notice emitted during a submit now suppresses the generic turn-failed event, while the generic event remains as a fallback when no detailed error notice arrives.

docs/rpc.md
markustherobot: I updated the nested initialize result to `protocolVersion: 9` so the canonical example is internally consistent.

docs/sdk.md
markustherobot: I narrowed the SDK example on `presentation === "transcript"` before reading `content`, so the example now typechecks against the feedback union.

docs/session-event-protocol-redesign.md
markustherobot: I replaced the obsolete design-document notice type and example with the canonical structured `title`, optional `content`, required `subject`, severity, and timestamp contract.

markusylisiurunen:
@markustherobot review

src/protocol/session_protocol.ts
```diff
@@ -2807,6 +3160,7 @@ export function applySessionProtocolDelta(
   }
 
   if (message.delta.type === "snapshot.reset") {
+    validateSnapshotResetCause(snapshot, message);
```
markustherobot, line 3198: **[P2] Preserve the rewind sequence high-water mark**

This only verifies that the cutoff is no greater than the current high-water mark before accepting the reset snapshot wholesale. A rewind from sequence 5 to an empty timeline with `sequence: 0` is therefore valid, so later items can reuse sequence numbers already allocated to durable or ephemeral events in the same epoch. Require the rewind snapshot to retain the current `timeline.sequence`, and cover the regression with a delta-application test.

src/protocol/session_protocol.ts
```diff
@@ -2866,26 +3217,30 @@ export function applySessionProtocolDelta(
...
-        break;
-      case "timeline.replace": {
-        const index = next.timeline.findIndex((item) => item.id === change.item.id);
-        if (index === -1) {
-          next.timeline.push(structuredClone(change.item));
-        } else {
-          next.timeline[index] = structuredClone(change.item);
-        }
+        next.timeline.items.push(structuredClone(change.item));
+        next.timeline.sequence = Math.max(next.timeline.sequence, change.item.sequence);
```
markustherobot: **[P2] Reject appends that reuse allocated sequences**

After `timeline.advance` reserves sequence 5 for an ephemeral item while persisted items end at 1, an append at sequence 2 or 5 still passes because this takes the maximum and the final snapshot remains locally sorted. That defeats the high-water allocation contract and can give durable and ephemeral items the same sequence. Reject each append unless its sequence is strictly greater than the timeline high-water mark before applying it.

docs/session-event-protocol-redesign.md
```diff
@@ -1,745 +1,160 @@
...
-  | Message
-  | {
-      role: "assistant";
-      content: (TextContent | ThinkingContent | ToolCall)[];
-      timestamp: number;
-    };
+interface SessionProtocolNotice {
+  kind: string;
+  version: number;
+  severity: "info" | "warning" | "error";
```
markustherobot: **[P2] Match the notice example to the wire contract**

The canonical example names `SessionProtocolNotice` and advertises `"warning"`, but the exported type is `SessionProtocolTimelineNotice` and protocol v9 accepts `"warn"`. A custom client following this document will construct a payload the validator rejects. Update both the type name and severity literal.

src/tui/session_chat_controller.ts
```diff
@@ -1389,23 +1447,20 @@ export class SessionChatController {
...
-  }
-
   private async syncFromSessionSnapshot(): Promise<boolean> {
     try {
-      this.syncRenderedHistory(await this.session.snapshot());
+      const snapshot = await this.session.snapshot();
+      if (snapshot.timeline.epoch > this.snapshot.timeline.epoch) {
+        this.renderCompactedSnapshot(snapshot);
+      } else {
+        this.syncRenderedHistory(snapshot);
```
markustherobot, line 1484: **[P2] Prune ephemeral notices when recovery adopts a rewind**

If revision-gap recovery fetches a snapshot after a rewind, this same-epoch branch retains every local ephemeral timeline item. The queued rewind reset is then skipped because its `toRevision` is already covered by the fetched snapshot, so the normal cutoff pruning never runs and feedback from the removed tail stays visible. Carry the queued rewind cutoff through recovery, or otherwise reconcile current-epoch ephemeral items before adopting the snapshot.

src/tui/session_chat_controller.ts
```diff
@@ -697,25 +709,20 @@ export class SessionChatController {
...
-        this.view.addSystemMessage(
-          `turn failed: ${result.turn.errorMessage ?? "model provider returned an error"}`,
-          "error",
-        );
-      }
+      await task();
       await this.syncFromSessionSnapshot();
     } catch (error) {
-      this.view.addSystemMessage(`session turn failed: ${formatSessionError(error)}`, "error");
+      this.view.addTranscriptNotice("failed to run assistant turn", "error", [
```
markustherobot, line 726: **[P2] Avoid duplicating host-owned runtime failures**

`LocalHostedSessionHandle.cleanupFailedTurn()` now persists a `tau.turn.failed` notice and emits its reset before rethrowing the runtime error. This catch adds a second local notice for that same failure, so the TUI renders both diagnostics. Synchronize first and add this fallback only when no new host-owned failure notice was projected, while retaining it for transport and pre-submit failures.

src/core/telegram/session_manager.ts
```diff
@@ -944,7 +960,7 @@ class TelegramSessionManagerImpl implements TelegramSessionManager {
       entry.unsubscribeClientEvents = tauSession.onDelta((event) => {
         this.handleClientEvent(entry, event);
       });
-      this.handleSnapshotNotices(entry, await tauSession.snapshot());
+      this.initializeSnapshotDeliveryState(entry, await tauSession.snapshot());
```
markustherobot: **[P2] Deliver warnings created with a fresh Telegram session**

`sessions.create()` can already return `tau.history.unavailable` in its initial snapshot because the host records a history-registration failure during creation. Marking every initial notice as emitted means a fresh Telegram user never receives that warning even though no prior adapter could have delivered it. Use baseline suppression only for recovered sessions; for a new session, pass initial warnings through `handleSnapshotNotices` and rely on the ID set to deduplicate any live race.

markustherobot (commented):
I found correctness gaps around timeline invariants, recovery reconciliation, and feedback delivery. I launched six Luna/high subreviews covering protocol and SDK contracts, host timeline transitions, storage migration and recovery, TUI reconciliation, runtime and tool feedback, and Telegram, presentation, themes, and docs; all six completed.

src/protocol/session_protocol.ts
markusylisiurunen: Fixed in c2519e4: rewind resets now require the snapshot to preserve the current timeline sequence, with regression coverage.

src/protocol/session_protocol.ts
markusylisiurunen: Fixed in c2519e4: timeline appends now reject sequences at or below the current high-water mark, with lower and equal-value regression cases.

docs/session-event-protocol-redesign.md
markusylisiurunen: Fixed in c2519e4: the example now uses SessionProtocolTimelineNotice and the canonical warn severity.

src/tui/session_chat_controller.ts
markusylisiurunen: Fixed in c2519e4: revision-gap recovery discards only pre-recovery ephemeral items; tests cover stale pruning, in-flight arrivals, and frozen epochs.

src/tui/session_chat_controller.ts
markusylisiurunen: Fixed in c2519e4: successfully persisted runtime failures now resolve as failed turn outcomes, leaving the TUI fallback only for unsettled request and transport failures.

src/core/telegram/session_manager.ts
markusylisiurunen: Fixed in c2519e4: recovered sessions still baseline existing notices, while fresh sessions deliver initial snapshot warnings through the deduplicated notice path.

PR #403, opened 2026-08-07
allow live clock and randomness in code mode

markusylisiurunen:
## why

SES secure mode blocks ambient current-time and random-number access by default. That makes ordinary generated JavaScript using `Date.now()`, zero-argument `new Date()`, or `Math.random()` fail with cryptic secure-mode errors, even though these capabilities are appropriate for Tau's one-shot code-mode tools.

## what

Capture the native Date and Math capabilities before SES lockdown and explicitly endow every history, web, and Nook compartment with them. Date is exposed through a coherent hardened facade created after lockdown, preserving `Date.prototype.constructor` identity and constructor-derived calls while retaining live clock access. The shared code-mode contract and progressive documentation advertise live clock and random generation support, with regression coverage across all three tools while retaining the existing sandbox escape checks.


src/core/static/code_mode/web/sandbox_runner.mjs
```diff
@@ -54,6 +54,8 @@ function writeOutput(stream, text) {
 
 const compartment = new Compartment({
   globals: {
+    Date: codeModeDate,
```
markustherobot, line 58: **[P2] Preserve Date's constructor identity**

Because lockdown rewrites the shared `Date.prototype.constructor`, the pre-lockdown constructor endowed here is no longer paired with its prototype: inside every code-mode compartment, `new Date().constructor !== Date`, and both `new Date().constructor()` and `new new Date().constructor()` still throw the secure-mode error this change intends to remove. Code that uses constructor identity for Date detection or constructor-based creation therefore breaks despite the documented standard Date contract. Please expose a coherent hardened Date facade whose prototype points back to the endowed constructor, and cover this identity and constructor-derived path in the sandbox tests.

markustherobot (commented):
Found one Date semantic regression. I launched three read-only subreviews covering SES capability security, distribution packaging, and the public capability contract/tests; the packaging and contract paths were otherwise clean.

markusylisiurunen:
@markustherobot fix the finding

src/core/static/code_mode/web/sandbox_runner.mjs
markustherobot: Fixed in 691dca0. Each runner now creates a hardened Date facade after lockdown with a coherent prototype and constructor identity; all three sandbox tests cover identity, prototype alignment, callable and constructable derived constructors, and hardening.

PR #408, opened 2026-08-07
restore durable turn recovery

markusylisiurunen:
## why

Tau removed the durable request-to-outcome association while separating feedback presentation from session messages. That left SDK consumers to reconstruct logical turns from model history and timeline notices, even though compaction can remove the initiating message and internal continuations can make transcript ordering ambiguous.

Client request recovery is a separate state domain from model history and timeline presentation. Tau needs to own that contract so consumers can determine whether a stable submitted history entry was accepted, is still running, or reached a terminal outcome without reproducing Tau's internal turn semantics.

## what

Add a top-level durable logical-turn ledger to session snapshots, keyed by submitted user history entry ID. Acceptance persists a running receipt atomically with the user message before model work, and settlement persists the terminal outcome before the live response.

The host preserves receipts through compaction and rewind, associates steering and goal continuations at the host logical-turn boundary, rejects explicit ID reuse, and converts interrupted running receipts to aborted outcomes during recovery. Timeline notices remain presentation feedback and are not used for outcome recovery.

The protocol now carries validated `turn.set` changes at version 10, stored sessions migrate to version 8, and the public SDK exposes typed record and outcome lookup helpers. Legacy v5 embedded outcomes become settled records, while v6 and v7 sessions continue to open with an empty ledger rather than inferring discarded outcomes from transcripts.

## details

Regression coverage includes atomic acceptance, terminal settlement, split-turn compaction after the initiating message is removed, hidden-continuation recovery, automatic goal continuation, steering, rewind and ID reuse, storage migration, protocol delta application, and packaged SDK types.


src/host/local_session_host.ts
```diff
@@ -851,10 +859,79 @@ class LocalHostedSessionHandle implements LocalHostedSession {
...
   }
 
+  async acceptTurn(
+    options: Omit<SessionProtocolRecordParams, "sessionId">,
+  ): Promise<SessionProtocolRecordResult> {
+    this.assertActive();
+    const userHistoryEntryId = this.createTurnHistoryEntryId(options.historyEntryId);
+    this.pendingAcceptedTurnHistoryEntryIds.add(userHistoryEntryId);
+    try {
+      await this.session.commitUserText(options.text, { historyEntryId: userHistoryEntryId });
```
markustherobot, line 872: **[P1] Roll back the runtime entry when acceptance persistence fails**

`commitUserText()` appends to `AgentRuntime` before the awaited event sink persists this patch. If that store commit fails after a baseline snapshot exists, `enqueueMutation()` restores `turns` from the committed snapshot but leaves the new runtime history entry in place; a later snapshot or shutdown can then persist the user message without any turn receipt. That violates the atomic acceptance guarantee under a normal storage failure, so the appended runtime entry needs to be rolled back (or acceptance staged) before returning the error.

src/host/local_session_host.ts
```diff
@@ -2306,6 +2418,10 @@ class LocalHostedSessionHandle implements LocalHostedSession {
           ],
           { persist: true },
         );
+        if (turn) {
+          this.pendingAcceptedTurnHistoryEntryIds.delete(event.historyEntryId);
+          this.pendingRuntimeSteeringIds.clear();
```
markustherobot: **[P1] Clear only the steering IDs associated with this message**

While `emitPatch()` is awaiting persistence for one steering continuation, another `steer()` call can add its ID to this shared set. This `clear()` then removes that later ID too; when the later continuation emits its own `user_message`, no running receipt is created, and its result eventually fails in `settleTurn()` as “not accepted.” Capture the IDs associated with the current event and delete only those, or carry the steering association explicitly with the event.

src/host/local_session_host.ts
```diff
@@ -841,6 +842,13 @@ class LocalHostedSessionHandle implements LocalHostedSession {
     options: Omit<SessionProtocolRecordParams, "sessionId">,
   ): Promise<SessionProtocolRecordResult> {
     this.assertActive();
+    if (
+      options.historyEntryId &&
+      (this.turns.has(options.historyEntryId) ||
```
markustherobot: **[P2] Normalize custom IDs before ledger bookkeeping**

The protocol accepts IDs with surrounding whitespace, but `AgentRuntime` trims explicit IDs. For `acceptTurn({ historyEntryId: " turn-1 " })`, the pending set therefore does not match the emitted `turn-1` event, so the message is committed without a receipt and `runAcceptedTurn()` fails. The same mismatch lets `record()` bypass this reuse check after compaction removed the original message. Normalize once before every ledger check, pending-set insertion, and runtime commit (or stop trimming downstream).

src/protocol/session_protocol.ts
```diff
@@ -3391,6 +3433,17 @@ function hasSameFacetSubject(previous: SessionProtocolFacet, facet: SessionProto
...
 
+function setSessionProtocolTurn(
+  turns: SessionProtocolSnapshot["turns"],
+  turn: SessionProtocolTurnRecord,
+): void {
+  const previous = turns[turn.userHistoryEntryId];
+  if (previous?.state === "settled" && JSON.stringify(previous) !== JSON.stringify(turn)) {
+    throw new Error(`settled turn '${turn.userHistoryEntryId}' cannot be changed`);
+  }
+  turns[turn.userHistoryEntryId] = structuredClone(turn);
```
markustherobot: **[P2] Store turn IDs as own properties**

`__proto__` is a valid custom history ID, but assigning it on an ordinary object changes the object prototype instead of creating an enumerable receipt. The record is then omitted by `Object.entries()`, serialization, and subsequent cloning; similarly, the SDK lookup returns inherited values such as `toString` for unknown IDs. Use an own-property-safe map/update plus own-property lookups, or reject prototype-sensitive IDs at the protocol boundary.

markustherobot (commented):
Found four material correctness issues around failure rollback, concurrent steering association, and client-supplied ledger keys. I launched four read-only subreviews covering the host lifecycle, protocol and migrations, handler and SDK contract, and tests plus cross-cutting coverage; all completed successfully.

src/host/local_session_host.ts
markusylisiurunen: Fixed in 26b1060. `AgentRuntime` now rolls back the exact appended user entry and revision when `user_message` delivery fails. Added runtime and hosted-session regressions proving failed acceptance leaves neither message nor receipt and the explicit ID can be retried safely.

src/host/local_session_host.ts
markusylisiurunen: Fixed in 2e162fd. Internal `user_message` events now carry an explicit `input` or `steering` origin, and the host no longer maintains shared pending steering IDs. Added a regression that submits a second steering request while the first continuation is blocked in persistence and verifies both durable receipts settle.

src/host/local_session_host.ts
markusylisiurunen: Fixed in 5d5e84a. Explicit history IDs are now normalized before host ledger checks, pending acceptance, and runtime commit; protocol request parsing also canonicalizes user-message and rewind IDs. The regression accepts a padded ID, compacts away its message, and verifies padded reuse is still rejected by both acceptance and recording.

src/protocol/session_protocol.ts
markusylisiurunen: Fixed in c91bbed. Turn deltas now use own-property-safe reads and `Object.defineProperty` writes, so `__proto__` is retained as an enumerable serializable receipt. SDK record/outcome helpers now reject inherited properties with `Object.hasOwn`. Regressions cover `__proto__` delta serialization and unknown `__proto__`, `constructor`, and `toString` lookups. Full suite: 76 files, 1,051 tests passed.

PR #409, opened 2026-08-07
preserve parent runtime for working-directory subagents

### LINKED ISSUE #407: preserve active reasoning for working-directory subagents
## why

A `spawn_agent` call that omits `model` should inherit the active session persona's model settings. When the call includes `workingDirectory`, Tau currently re-resolves the persona from the target directory's catalog and silently loses session-level settings such as the selected reasoning effort.

This caused a session running `gpt-5.6-sol-chatgpt-coder:high` to spawn the default subagent as `openai-codex/gpt-5.6-sol:low`. The call explicitly supplied the current repository as `workingDirectory` and correctly omitted `model`.

## reproduction

1. Start a session with `gpt-5.6-sol-chatgpt-coder:high`.
2. Call `spawn_agent` without `model`, but include the current directory as `workingDirectory`.
3. Observe that the spawned agent reports `openai-codex/gpt-5.6-sol:low`.
4. Make the same call without `workingDirectory` and observe that it inherits `high`.

## cause

`src/core/tools/spawn_agent.ts` invokes `resolveSubagentRuntime` whenever `workingDirectory` is present, including when it resolves to the current session cwd. `createExecutionEnvironmentSubagentRuntimeResolver` then loads the target runtime config and selects a fresh catalog persona by ID. That persona does not contain the active session's reasoning override, so `resolveSubagentEffectiveSettings` inherits the built-in Sol default of `low`.

`subagents.defaultLaunchModels` does not prevent this because it is an allowlist for explicit launch overrides, not a default model selection.

The working-directory test in `test/spawn_agent_tool.test.js` returns the passed persona from its mocked resolver, so it does not cover the loss caused by selecting a fresh catalog persona.

## expected behavior

Working-directory context resolution should preserve the active session persona's selected model settings while resolving directory-owned config, prompts, skills, model catalog, and subagent definitions. Supplying the current cwd should not change the effective launch model or reasoning.

markustherobot:
## why

Supplying `workingDirectory` to `spawn_agent` re-resolved the active persona from the target directory. That made an incidental directory change replace the parent model, reasoning override, subagent definition, tools, and runtime config.

A working directory should define where the child operates and which prompt context it sees, not which agent Tau launches.

## what

`spawn_agent` now keeps subagent selection, launch authorization, model resolution, inherited settings, tools, and runtime config bound to the parent session's captured turn specification. Omitting `model` inherits the exact active parent model and settings, while an explicit model remains subject to the parent-owned launch allowlist.

A different working directory rebuilds only the child prompt from target environment and repository metadata, AGENTS.md context, target `agentContextFiles`, and target-discovered skills filtered by the parent persona. A working directory that resolves to the parent cwd is treated exactly like omission.

Custom subagent definitions now have one canonical shape: prompt, description, tools, and launch-model allowlist. Per-subagent provider, model, reasoning, and service-tier settings were removed so launch selection has only the two agreed cases.

Regression coverage verifies same-cwd inheritance, cross-directory prompt rebasing, conflicting and absent target personas, source runtime ownership, target-context failure behavior, and the canonical custom-subagent contract.

## details

The parent session remains authoritative for credentials, model notices, auto-compaction, model catalog, subagent availability, and tool-service configuration. Target configuration contributes only context-owned resources (`agentContextFiles` and discovered skills).

Existing unknown-field handling strips the removed custom-subagent model/settings keys. This is an intentional pre-v1 contract simplification rather than a compatibility path. There are no remaining questions.

fixes #407


markusylisiurunen:
@markustherobot can you open an issue to re-introduce subagent specific model overrides in the future, should be thought out extensively to make them work well. keep the issue lightweight.

PR #411, opened 2026-08-07
reconcile Telegram turns durably

markusylisiurunen:
## why

Telegram currently treats the live submit or steering response as the source of truth for turn completion. If that response disappears after Tau accepts or settles the request, the runner cannot distinguish a rejected request from accepted work, and restart recovery has no durable association between Telegram's unresolved work and Tau's outcome.

PR #408 added the canonical turn ledger needed to remove that ambiguity. Telegram should consume that contract directly instead of reconstructing request outcomes from response timing, transcript shape, or timeline presentation.

## what

Persist unresolved canonical turn IDs with Telegram session state and reconcile them against `snapshot.turns` during live updates, ambiguous submit failures, and runner recovery.

Idle Telegram messages now use a persisted explicit history entry ID before submission. Messages arriving during active work retain Tau's canonical steering and batching semantics. One turn-record reconciler handles running and settled records from deltas, snapshots, and live responses, emits failed or blocked outcomes once, and moves the session back to waiting only after all tracked work settles.

Recovered snapshots continue to establish the baseline for historical messages, notices, and unrelated turn outcomes. Only outcomes belonging to persisted unresolved Telegram requests are replayed after chat routing is restored.

## details

The Telegram session-state document moves to version 3 with required `activeTurnIds`; versions 1 and 2 normalize to an empty set. A failed record with unresolved turn ownership is reconnected so the ledger can resolve the ambiguity, while terminal failed records remain tombstones.

If a submit response fails after acceptance, Telegram checks the requested ID in the authoritative snapshot. Accepted work remains attached and continues through turn deltas; unknown IDs follow the existing terminal submit-failure path. Presentation notices remain separate and are not used for request recovery.


src/core/telegram/session_manager.ts
```diff
@@ -1177,6 +1256,131 @@ class TelegramSessionManagerImpl implements TelegramSessionManager {
     return submitPromise;
   }
 
+  private async recoverRejectedSubmit(
+    entry: SessionEntry,
+    tauSession: TelegramTauSession,
+    historyEntryId: string,
+    error: unknown,
+  ): Promise<boolean> {
+    const snapshot = await tauSession.snapshot().catch(() => undefined);
```
markustherobot: **[P1] Preserve the unresolved ID when the fallback snapshot fails**

If `submit()` loses its response and this follow-up `snapshot()` also fails, the catch converts that read failure into `undefined`, so the branch below deletes and persists the ID exactly as if an authoritative snapshot proved non-acceptance. The turn may already have been accepted, and the next runner then has no ownership marker with which to reconcile it. Distinguish snapshot failure from a successful snapshot without the ID, and retain the unresolved ID for restart recovery when the read itself fails.

markustherobot (commented):
I found durability gaps where persisted Telegram ownership is converted back into transient manager or adapter state. I launched four subreviews covering live turn concurrency, persistence and recovery, adapter outcome delivery, and protocol and test integration; all four completed successfully.

src/core/telegram/session_manager.ts
```diff
@@ -1177,6 +1256,131 @@ class TelegramSessionManagerImpl implements TelegramSessionManager {
...
+    await this.persistSessions();
+    this.log(entry, "warn", "submit response recovered from turn ledger", {
+      userHistoryEntryId: historyEntryId,
+      turn,
+      cause: formatErrorDiagnostic(error),
+    });
+    if (turn.state === "settled" && entry.activeTurnIds.size === 0) {
+      this.setState(entry, "waiting-input");
+    }
+    return true;
```
markustherobot, line 1396: **[P1] Do not release a rejected submit on a nonterminal receipt**

A matching `state: "running"` receipt proves acceptance but not completion. After this returns `true`, the submit promise finishes and `activeSubmit` is cleared; if the host rejected because `runAcceptedTurn` or settlement failed, no later `turn.set` exists and the Telegram session remains running forever. Even if work is still live, `/interrupt` refuses it because `interruptSession()` requires `activeSubmit`. Treat only settled records as recovered, or keep an interruptible waiter that follows the running receipt to settlement.

src/core/telegram/session_manager.ts
```diff
@@ -1177,6 +1256,131 @@ class TelegramSessionManagerImpl implements TelegramSessionManager {
...
+    const event: TelegramSessionTurnFailure = {
+      type: "session-turn-failed",
+      sessionId: entry.record.id,
+      projectId: entry.record.projectId,
+      timestamp: this.now().toISOString(),
+      historyEntryId,
+      failure: structuredClone(failure),
+    };
+    if (retain) {
+      const turnFailures = this.recoveredTurnFailures.get(entry.record.id) ?? [];
```
markustherobot: **[P1] Keep recovered failures durable until Telegram delivery**

On restart, reconciliation removes the owned ID and persists that removal, while the failure is retained only in this in-memory map. If adapter startup then fails at `getMe`, or the process exits before the queued notification is sent, the next start has neither the ID nor the failure and silently loses the outcome. Persist a pending recovered-failure record (or retain the owned ID) until the adapter has actually consumed or acknowledged it.

src/core/telegram/session_manager.ts
```diff
@@ -1177,6 +1256,131 @@ class TelegramSessionManagerImpl implements TelegramSessionManager {
...
+    entry: SessionEntry,
+    snapshot: SessionProtocolSnapshot,
+    options: { removeUnknownActiveTurns: boolean; retainFailures: boolean },
+  ): boolean {
+    let changed = false;
+    const snapshotTurnIds = new Set(Object.keys(snapshot.turns));
+    if (options.removeUnknownActiveTurns) {
+      for (const historyEntryId of entry.activeTurnIds) {
+        if (!snapshotTurnIds.has(historyEntryId)) {
+          entry.activeTurnIds.delete(historyEntryId);
```
markustherobot: **[P2] Surface owned IDs that were never accepted**

The ID is durably written before `tauSession.submit()` is called, so a crash between those operations legitimately recovers an active ID absent from the ledger. Deleting it here and then reconnecting as `waiting-input` silently drops that Telegram message, which may already have been acknowledged by the next polling request. Treat an absent owned ID as a recovered rejection that is reported to the chat, or persist enough of the request to replay it idempotently.

src/core/telegram/session_manager.ts
markusylisiurunen: Fixed in 9dcb170. A failed fallback snapshot no longer counts as proof that the turn was absent, so the unresolved ID remains persisted for restart reconciliation.

src/core/telegram/session_manager.ts
markusylisiurunen: Fixed in 9dcb170. A recovered running receipt now keeps an active tracking task, polls the ledger until settlement, and remains interruptible even when no turn delta arrives.

src/core/telegram/session_manager.ts
markusylisiurunen: Fixed in 9dcb170. Turn failure notifications are persisted in the version 4 Telegram state and removed only after the adapter confirms successful Telegram delivery.

src/core/telegram/session_manager.ts
markusylisiurunen: Fixed in 9dcb170. An authoritative snapshot without the owned ID now creates a durable rejection notification, keeps the session usable, and asks the user to resend the message.

PR #412, opened 2026-08-07
prefer structured sources over web retrieval

markusylisiurunen:
## why

The web tool treated any provided URL as sufficient reason to invoke it. That pushed agents toward generic webpage extraction even when local repository data, `gh`, `git`, or another structured source could answer more directly and reliably.

## what

Tighten the web tool description so agents prefer local context, purpose-built CLIs, first-party APIs and SDKs, and structured endpoints before open-web retrieval. GitHub URLs now explicitly prefer `gh` or `git` whenever they can provide the needed information.

Also remove prompt-phrase assertions that pinned guidance wording without testing tool selection or other meaningful behavior. Existing execution, validation, schema, limit, and structured prompt tests remain.

### DIFF
--- a/src/core/tools/web.ts
+++ b/src/core/tools/web.ts
@@ -25,7 +25,11 @@ const WEB_DESCRIPTION = buildCodeModeToolDescription({
   sdkGlobal: "web",
   introduction: [
     "Run a one-shot JavaScript program to search the web and retrieve page content.",
-    "Use this tool only when the user asks to browse or search the web, provides a URL, or otherwise clearly implies that web access is needed.",
+    "Use this tool only when the task requires open-web search or webpage extraction and no more direct or structured source can answer it.",
+    "Before using it, prefer local files and repository data, purpose-built CLIs, first-party APIs and SDKs, and direct structured endpoints.",
+    "A URL alone does not justify using this tool.",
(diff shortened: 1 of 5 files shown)


src/core/tools/web.ts
```diff
@@ -25,7 +25,11 @@ const WEB_DESCRIPTION = buildCodeModeToolDescription({
   sdkGlobal: "web",
   introduction: [
     "Run a one-shot JavaScript program to search the web and retrieve page content.",
-    "Use this tool only when the user asks to browse or search the web, provides a URL, or otherwise clearly implies that web access is needed.",
+    "Use this tool only when the task requires open-web search or webpage extraction and no more direct or structured source can answer it.",
```
markustherobot, line 28: **[P2] Update the documented web-selection policy**

This changes the model-facing eligibility rule, but `README.md:425` still says the minimal description allows `web` whenever a request asks for or clearly implies web access. That now contradicts the structured-source fallback and GitHub `gh`/`git` preference introduced here, so users reading the documented behavior get the old policy. Update the README section to match this hierarchy, as required by the repository's documentation-sync guidance.

markustherobot (commented):
The implementation is internally consistent and the full test suite passes, but the user-facing documentation remains out of sync with the new selection policy. I launched two subreviews covering web-selection semantics and the removed regression coverage; both completed successfully.

PR #415, opened 2026-08-08
make public Nook sites public

### LINKED ISSUE #413: public Nook sites are not public

markusylisiurunen:
## why

Nook recorded public deployment visibility in the Worker, but Cloudflare Access was configured for the complete hostname. Anonymous requests were therefore blocked before the Worker could serve a public site.

## what

Scope Cloudflare Access to the root `/__nook/*` control plane and use a protected auth bootstrap endpoint with hostname-scoped Access cookies for private sites. Public site assets and browser KV can now reach the Worker anonymously, while private sites still validate Access JWTs.

Move CLI and model-tool KV operations to the protected management API so service-token authentication remains isolated from the visibility-aware browser KV routes. Update setup guidance, documentation, and regression coverage for the new routing and authentication contract.

fixes #413


src/nook/worker/index.ts
```diff
@@ -226,10 +226,13 @@ const CLIENT_JS = `(() => {
     });
     if (!response.ok) {
       let message = response.statusText;
+      let authUrl;
       try {
         const payload = await response.json();
         message = payload && payload.error && payload.error.message ? payload.error.message : message;
+        authUrl = payload && payload.error && payload.error.authUrl;
       } catch {}
+      if (response.status === 401 && authUrl) window.location.assign(authUrl);
```
markustherobot, line 235: **[P2] Preserve the app page during KV reauthentication**

When a private site's Access cookie expires while the app is open, this navigates to the server-provided auth URL whose `returnTo` is the KV endpoint itself. After login, a `put` or `delete` is converted into a top-level GET and any KV operation replaces the app with the endpoint's JSON response, so reauthentication loses both the page and the original mutation. Build the auth URL around the current document URL (and reload or retry explicitly) rather than returning the browser to `/<site>/__nook/kv/*`.

markustherobot (commented):
I found one material browser-flow regression. The control-plane KV routing and Access boundary otherwise align with the stated contract. I launched three subreviews covering Worker authentication/routing, the Tau-side client/setup contract, and end-to-end browser/test behavior; all completed successfully.

PR #417, opened 2026-08-08
make subagent activity host-owned

### LINKED ISSUE #404: subagent panel is broken as fuck
for example, it does not show what the subagent is doing.

markusylisiurunen:
## why

The subagent panel depended on a single text facet and reconstructed activity history locally by parsing text prefixes. That made attach behavior incomplete, left each client with different state, and coupled presentation to fragile strings.

## what

This introduces an independently revisioned, non-persisted `session.subagentActivities` channel owned by the host. `session.observe` returns the complete activity baseline, while later messages replace only each changed agent's complete current-run list or explicitly remove an agent. Each run retains up to 64 typed assistant, settled-tool, and notice activities, and follow-up runs reset that list.

The attach bootstrap, transports, SDK, and TUI consume the same authoritative state. Replacement-session handoff performs a final reconciliation after buffered updates, so an agent and its already-received activities cannot become visually desynchronized.

## details

The protocol advances to version 11. Activity text uses field-specific UTF-8 content budgets with middle or head-tail truncation: 512 bytes for short labels, names, and titles, 4 KiB for tool subjects, 32 KiB for assistant text and combined tool details, 8 KiB for combined metadata, and 32 KiB for combined notice content.

Tool activities appear only after settlement with terminal producer-owned presentation. Exact tool lookup still uses the raw model-provided name, while rejected-call presentation names are normalized before protocol validation. Thinking is omitted, typed activity entries have no dedicated tool-call-ID field, and bounded diagnostic notices may still mention an ID.

Activity state survives client detach while the hosted runtime remains live, but is discarded on restart or recovery. Completed-agent histories are intentionally retained without a session-wide eviction policy; per-agent updates avoid retransmitting unrelated histories on every activity.

fixes #404


src/core/subagents/agent_supervisor.ts
```diff
@@ -422,38 +416,89 @@ export class AgentSupervisor {
...
       type: "subagent_activity",
       state: this.toSnapshot(record),
-      text: normalized,
+      activity: {
+        type: "tool",
+        toolName,
+        outcome,
+        presentation: {
+          action: actionByStatus[outcome],
+          ...structuredClone(terminalPresentation),
```
markustherobot, line 499: **[P1] Bound terminal presentations before publishing**

A normal subagent `edit` can produce more than 64 diff lines because its canonical presentation deliberately uses `detailTruncation: false`. Forwarding that presentation unchanged conflicts with the new wire schema's 64-detail limit, so `createSessionProtocolSubagentActivitiesMessage()` throws through the awaited event sink after the file was successfully edited, marks the child run failed, and leaves invalid activity state that also breaks later `session.observe` validation. Apply an activity-channel truncation policy with an omission marker before storing/publishing the presentation, and cover an edit over 64 lines.

src/core/subagents/agent_supervisor.ts
```diff
@@ -422,38 +416,89 @@ export class AgentSupervisor {
...
+    if (event.type === "feedback") {
+      await this.emit({
+        type: "subagent_activity",
+        state: this.toSnapshot(record),
+        activity: {
+          type: "notice",
+          severity: event.tone === "error" ? "error" : "info",
+          title: event.title,
+          ...(event.presentation === "transcript" && event.content
+            ? { content: event.content }
```
markustherobot, line 476: **[P2] Keep tool-call IDs out of notice activities**

When `runPreparedToolCall()` catches a tool execution exception, its transcript feedback content explicitly includes `tool call id: ...`. Forwarding `event.content` verbatim exposes that ID through `session.subagentActivities` to SDK clients even though this channel's contract says tool-call IDs remain internal, and it also bypasses the bounded notice projection used by the main host. Project and sanitize feedback content before emitting the activity, stripping internal diagnostics and enforcing the protocol limits.

src/host/local_session_host.ts
```diff
@@ -1862,12 +1890,22 @@ class LocalHostedSessionHandle implements LocalHostedSession {
...
           this.agentCostTotals.delete(id);
         }
       }
+      for (const id of this.subagentActivitiesByAgent.keys()) {
+        if (!this.session.hasSubagent(id)) {
+          this.subagentActivitiesByAgent.delete(id);
+          activitiesChanged = true;
+        }
+      }
+      if (activitiesChanged) {
```
markustherobot: **[P2] Publish activity cleanup only after the mutation commits**

On rewind, supervisor origins are removed before `history_rewound` calls this reconciliation, so this publishes the replacement activity state before the following `commitSnapshot()` succeeds. If the store commit fails, the request fails and snapshot-owned agent state is restored, but observers have already received the deletion and the transient revision is not rolled back. Stage this cleanup until the durable mutation commits, or restore the activity map/revision and publish a compensating replacement on failure.

docs/rpc.md
```diff
@@ -190,7 +191,7 @@ params (required):
 
 ```json
 {
-  "version": 10,
+  "version": 11,
```
markustherobot, line 194: **[P2] Update the initialize protocol version**

This response envelope is updated to version 11, but the same example still returns `"protocolVersion": 10` a few lines below. Runtime initialization returns 11, so the canonical RPC example now documents an internally inconsistent handshake; update the nested value as part of the version bump.

markustherobot (commented):
I found material issues at the producer/protocol and lifecycle boundaries, including one expected tool path that can fail the subagent after successful work. I launched 4 read-only subreviews covering supervisor activity projection, host/protocol ownership and ordering, SDK/transports, and TUI rendering; all completed successfully.

src/core/subagents/agent_supervisor.ts
markusylisiurunen: the tool call id policy does not mean that e.g. an error message cannot include the id. the point was that the subagent activity entry doesn't have a separate tool call id field because that is not useful there

markusylisiurunen: but should be addressed if something to address beyond this

src/host/local_session_host.ts
markusylisiurunen: try to find a simple remidiation which is not easy to break in the future accidentally

markusylisiurunen:
@markustherobot address the unresolved findings

src/core/subagents/agent_supervisor.ts
markustherobot: Fixed in 9cba628. The host now projects terminal tool presentations before storing or publishing activity state, middle-truncating details to 64 lines with an omission marker and bounding metadata. The regression covers a canonical 65-line edit presentation and verifies publication succeeds.

src/core/subagents/agent_supervisor.ts
markustherobot: Addressed the remaining boundary issue in 9cba628. Per the clarification, textual diagnostic IDs remain valid, but there is still no dedicated tool-call-ID field. Notice titles and content now go through the canonical bounded projection before host storage, with coverage for oversized diagnostic content.

src/host/local_session_host.ts
markustherobot: Fixed in 9cba628. Missing-agent cleanup is now an option owned by commitSnapshot: snapshot-owned agents are reconciled before persistence, while transient activity deletion and publication happen only after the durable commit succeeds. A failed-rewind regression verifies that the activity state and revision remain unchanged and observers receive no cleanup message.

docs/rpc.md
markustherobot: Fixed in 9cba628. The nested initialize result now reports protocolVersion 11, matching the response envelope and runtime.

markusylisiurunen:
@markustherobot how many of the most recent subagent activity entries are stored in the live snapshot now? for the tool results in subagent activity entries, I'd prefer a size based limit instead of this line-based on the protocol level because lines can be short or long.

markusylisiurunen:
@markustherobot review

src/host/local_session_host.ts
```diff
@@ -2819,43 +2854,72 @@ class LocalHostedSessionHandle implements LocalHostedSession {
...
+    if (changed) {
+      this.publishSubagentActivities();
+    }
+  }
+
+  private publishSubagentActivities(): void {
+    this.subagentActivitiesRevision += 1;
+    const message = createSessionProtocolSubagentActivitiesMessage({
+      sessionId: this.sessionId,
+      state: this.subagentActivities(),
```
markustherobot: **[P1] Bound full replacements across completed agents**

Each spawn adds an activity entry, but `AgentSupervisor` retains completed records until rewind, so the active limit of 8 does not cap sequential agents. This call clones and sends every retained history on every new activity; with 64 entries per agent and 32 KiB text budgets, a long-running session can grow this channel to hundreds of MiB and make each subsequent publication increasingly expensive. Add a session-wide agent or byte retention limit (for example, evict activity histories for the oldest completed agents) and enforce the same limit in the protocol schema.

src/tui/session_chat_controller.ts
```diff
@@ -1125,6 +1150,14 @@ export class SessionChatController {
     this.sendPendingIdleNotification();
   }
 
+  private onSdkSubagentActivities(message: SessionProtocolSubagentActivitiesMessage): void {
+    if (message.state.revision <= this.subagentActivities.revision) {
```
markustherobot: **[P1] Reconcile activities after buffered agent deltas**

If an `agent.set` delta and its activity replacement arrive after `session.snapshot()` returns but before the controller installs listeners, the SDK already holds the latest activity state while buffering the delta. The controller seeds that state, drains the delta and creates an empty panel entry, then discards the immediate activity callback here because the revision is equal. The agent can remain blank until another activity arrives. Reconcile the authoritative state after applying buffered agent deltas, or allow the equal-revision baseline callback to synchronize newly added agents.

src/protocol/session_protocol.ts
```diff
@@ -451,6 +500,213 @@ export function projectSessionProtocolNoticePresentation(
...
+      type: "notice",
+      severity: activity.severity,
+      title: truncateUtf8Middle(activity.title, SESSION_PROTOCOL_MAX_SUBAGENT_SHORT_TEXT_BYTES),
+      ...(content.length > 0 ? { content } : {}),
+    };
+  }
+
+  return {
+    type: "tool",
+    toolName: truncateUtf8Middle(activity.toolName, SESSION_PROTOCOL_MAX_SUBAGENT_SHORT_TEXT_BYTES),
```
markustherobot: **[P2] Normalize model-provided tool names before validation**

An unavailable tool name comes directly from model output, and admission does not require it to be one line. This projection only truncates the raw name, so a name containing `\n` or `\r` reaches the one-line activity schema and makes publication throw, failing the subagent on an otherwise handled blocked call. Normalize line breaks before storing the activity, or validate and canonicalize the name at admission, and cover the malformed-name path.

markustherobot (commented):
I found material gaps in aggregate state bounding and in normalization/reconciliation at the new transient channel boundary. I launched 5 read-only subreviews covering supervisor activity production, host ownership and ordering, protocol projection and schema, SDK/transport delivery, and TUI reconciliation; all completed successfully.

src/host/local_session_host.ts
markusylisiurunen: Addressed in 3ff5dd2. We intentionally retain completed-agent histories for the live runtime without an aggregate eviction policy, but removed the repeated whole-session amplification: session.observe still returns the complete baseline, while each later message carries a complete replacement only for the changed agent (or explicit removals). The host no longer clones, validates, or sends every retained agent on each activity. Per-run and per-field bounds remain enforced.

src/tui/session_chat_controller.ts
markusylisiurunen: Fixed in 3ff5dd2. After the replacement-session controller drains all buffered snapshot and side-channel messages, it now performs one final authoritative reconciliation from the resulting snapshot agents and activity state. This avoids coupling hydration to equal-revision callbacks. A regression reproduces the buffered agent.set plus already-seeded activity revision and verifies the activity is rendered immediately.

src/protocol/session_protocol.ts
markusylisiurunen: Fixed in 3ff5dd2. Exact registry lookup still uses the raw model-provided name, so malformed input cannot become an executable known tool. After admission rejects the call, the presentation copy collapses whitespace, trims, and falls back to a generic invalid-name label; the activity protocol projection applies the same defensive normalization before byte bounding. Regressions cover a CRLF-containing unavailable tool name through runtime activity production and host publication.

PR #419, opened 2026-08-09
provide a code-mode runtime for client tools

### LINKED ISSUE #418: provide a code-mode runtime for command client tools
## why

Tau's existing command-backed client tools are enough to advertise a `{ code: string }` tool and invoke a local executable, but authors currently have to build the code-mode behavior themselves. That includes evaluating model-generated JavaScript safely, exposing a documented API namespace, routing API calls to trusted integrations, capturing console output, and fitting the result back into Tau's command client-tool contract.

Tau should make this easy without introducing another client-tool configuration shape. Keeping the current `clientTools` contract means the same feature works in the local TUI and `tau attach`, retains the existing global-only executable boundary and project allowlist, and requires no new host or session protocol concept.

## proposed shape

A code-mode tool remains an ordinary command client tool in global config:

```json
{
  "clientTools": [
    {
      "name": "linear",
      "defaultEnabled": true,
      "description": "Run one-shot JavaScript against Linear. First print docs with console.log(docs).",
      "parameters": {
        "type": "object",
        "properties": {
          "code": { "type": "string" }
        },
        "required": ["code"],
        "additionalProperties": false
      },
      "command": ".config/tau/tools/linear"
    }
  ]
}
```

The referenced executable owns the code-mode API and uses a Tau-provided authoring/runtime helper:

```ts
await runTauCodeModeCommand({
  name: "linear",
  documentation: `
# linear

- linear.issues.get(id)
- linear.issues.search(options)
`,
  api: {
    issues: {
      get: /* trusted handler */,
      search: /* trusted handler */,
    },
  },
});
```

From the model's perspective, the tool supports programs such as:

```js
const issue = await linear.issues.get("TAU-123");
console.log(issue);
```

The helper should provide the standard code-mode experience: an isolated environment for model-generated code, `docs` and console globals, structured calls into explicitly exposed trusted handlers, bounded execution, cancellation, and final textual output compatible with the existing command client-tool result.

The trusted executable may hold authenticated clients and other local capabilities, but generated code should only reach them through the API methods the executable deliberately exposes.

## desired outcome

- Authors can build a command-backed code-mode tool without implementing an evaluator, bridge, output capture, or Tau command framing themselves.
- Existing `clientTools` config, `enabledClientTools`, and `--no-client-tools` behavior remain unchanged.
- API documentation and method definitions live with the executable rather than being duplicated in config.
- Structured method results remain available to the running program before it prints its final answer.
- SDK users can reuse the same underlying code-mode runtime without the command wrapper.
- The feature does not require a new host-owned tool or session protocol execution path.

A small manifest or config-snippet generator could reduce the remaining repetition in the ordinary client-tool name, description, and `{ code: string }` schema, while still producing the current config format.

Code-mode-specific tool-card presentation is not required for the initial capability; the executable can continue to appear as an ordinary client tool.

markusylisiurunen:
## why

Command-backed client tools can already expose a `{ code: string }` surface, but every author currently has to implement isolated JavaScript evaluation, trusted API bridging, output capture, cancellation, and Tau's command framing themselves. The same runtime concerns were also duplicated across Tau's built-in code-mode tools.

## what

Add a shared code-mode runtime for built-in tools, programmatic SDK client tools, and standalone command-backed client-tool executables. The SDK now provides `createTauCodeModeClientTool()` and the lower-level executor, while `@markusylisiurunen/tau/code-mode` adds the executor, opt-in description builder, and `runTauCodeModeCommand()` adapter.

## details

The runtime exposes one declared API namespace through an SES worker, keeps trusted handlers in the parent, and carries arguments and results across a JSON boundary. It generates installed-version runtime documentation, applies common timeout, output, memory, and bridge-call limits, and offers best-effort caller-owned output persistence without assuming a shared filesystem. Worker bridge requests and responses are capped at 1 MiB, handler shutdown has a bounded drain, and sandbox workers do not inherit caller `execArgv`.

The existing web, history, and Nook tools now use the same generic runner. Their static documentation contains only API-specific guidance, while the runtime prepends canonical capabilities and limits. The shared bridge budget is 128 calls per execution with no more than eight unresolved concurrently.

fixes #418


markustherobot (commented):
Found three material issues around the hard isolation guarantees and public-runtime portability. I launched four Luna subreviews covering worker bridge isolation, built-in migrations, packaging/public API, and output/failure semantics; all four completed.

src/code_mode/runtime.ts
```diff
@@ -0,0 +1,420 @@
...
+        name: options.name,
+        methods: methods.map(({ id, path }) => ({ id, path })),
+      },
+      signal,
+      timeoutMs,
+      handleRequest: async (request, requestSignal) => {
+        const method = methods[request.methodId];
+        if (!method) throw new Error(`unsupported ${options.name} API method`);
+        const args = parseBridgeArguments(request.argsJson, options.name, method.path);
+        const value = await method.handler(args, { signal: requestSignal, invocation });
```
markustherobot, line 185: **[P1] Keep handler cancellation bounded**

Once this handler starts, a timeout or abort terminates the worker, but `executeCodeModeWorker()` then waits for every in-flight handler promise to settle. A handler that ignores `requestSignal` (for example, a stuck SDK call or `new Promise(() => {})`) therefore makes `executeTauCodeMode()` hang forever and can also block `client.close()`, despite the configured timeout. Race or bound the post-termination drain so a non-cooperative handler cannot prevent the outer execution from settling.

src/code_mode/runtime.ts
```diff
@@ -0,0 +1,420 @@
...
+        docs,
+        name: options.name,
+        methods: methods.map(({ id, path }) => ({ id, path })),
+      },
+      signal,
+      timeoutMs,
+      handleRequest: async (request, requestSignal) => {
+        const method = methods[request.methodId];
+        if (!method) throw new Error(`unsupported ${options.name} API method`);
+        const args = parseBridgeArguments(request.argsJson, options.name, method.path);
```
markustherobot, line 184: **[P1] Bound JSON bridge payloads outside the worker heap**

`argsJson` is parsed in the host with no byte limit, and handler results are likewise fully stringified before being posted back. Generated code can enqueue several very large argument payloads before the parent enforces the 8-call concurrency limit, so the message queue and host-side parsing bypass the worker's 128 MiB heap limit and can exhaust the Tau process; an unexpectedly large trusted-handler result has the same effect. Enforce a per-call byte cap before posting/parsing arguments and before returning serialized results, ideally with worker-side accounting so fan-out cannot first queue unbounded data.

src/code_mode/runtime.ts
```diff
@@ -0,0 +1,420 @@
...
+
+  const methods = registerMethods(options.api);
+  const timeoutMs = options.timeoutMs ?? TAU_CODE_MODE_DEFAULT_TIMEOUT_MS;
+  const invocation = options.invocation ?? null;
+  const signal = options.signal ?? new AbortController().signal;
+  const docs = buildRuntimeDocumentation(options.name, options.documentation, timeoutMs);
+  const startedAt = Date.now();
+  let execution: TauCodeModeExecutionCapture;
+  try {
+    execution = await executeCodeModeWorker({
```
markustherobot, line 171: **[P2] Sanitize inherited worker exec arguments**

The worker inherits the consumer process's `execArgv`. For example, calling this public executor from `node --input-type=module --eval ...` makes Node apply `--input-type` to the file-backed worker and every execution fails with `ERR_INPUT_TYPE_NOT_ALLOWED`. Pass a sanitized `execArgv` when creating the worker so valid parent launch modes do not break the published runtime.

src/code_mode/runtime.ts
markusylisiurunen: let's assume not malicious intent by default. we can implement some limits for safety but I'd like to keep this simple. it does not have to cover every possible edge case

src/code_mode/runtime.ts
markusylisiurunen: again, let's try to keep things simple and avoid fragile assumptions or heuristics

markusylisiurunen:
@markustherobot address the findings and in your response explain why they resolve the findings

src/code_mode/runtime.ts
markustherobot: Fixed in c416ec8. Worker exit now aborts active handler signals and races their settlement against a fixed one-second drain. Cooperative handlers still clean up, but a handler that never settles can no longer block the outer execution or client shutdown indefinitely. The regression test cancels an execution backed by a permanently pending handler and verifies that it settles.

src/code_mode/runtime.ts
markustherobot: Fixed in c416ec8 with a simple 1 MiB per-request/per-response cap. The worker enforces payload, total-call, and pending-call limits before posting requests, the parent validates request size before parsing, and serialized handler results are rejected before being posted back. This bounds queued bridge data to the eight allowed capped requests without introducing payload heuristics. Tests verify oversized arguments never reach the handler and oversized results never cross back to the worker.

src/code_mode/runtime.ts
markustherobot: Fixed in c416ec8 by starting the packaged worker with `execArgv: []`. The sandbox runner imports everything it needs, so it does not need caller process flags, and clearing them is deterministic rather than maintaining a fragile flag allowlist or denylist. A regression test launches the public executor through `node --input-type=module --eval` and verifies the file-backed worker succeeds.

PR #421, opened 2026-08-09
expose execution environments to client tools

markusylisiurunen:
## why

Client-provided tools run outside the session host, but workspace-aware tools need the same execution-environment access regardless of whether they are implemented directly through the SDK or as TUI command processes. SDK tools could previously capture a session and call `exec` themselves, while command tools had no canonical equivalent.

## what

Every SDK client-tool context now includes an execution-environment facade backed by the existing `session.exec` and `session.cancelExec` methods. Tool cancellation is composed into environment commands automatically, and the ordinary session facade shares the same execution implementation.

TUI command tools now use a version 2 bidirectional NDJSON process protocol. The exported `runTauClientToolCommand` helper presents command implementations with the same context as in-process SDK tools and relays environment execution through the observing TUI. Code-mode command tools use this shared adapter, while trusted code-mode API handlers can selectively expose the facade through their declared APIs.

The built-in diff review client tool now captures its workspace snapshot through the canonical client-tool execution context rather than a captured session execution path. Documentation and public package type coverage are updated for the new SDK and command contracts.


src/tui/command_client_tools.ts
```diff
@@ -39,9 +39,85 @@ export function createCommandClientTools(
...
+                createTauClientToolCommandExecResponse({
+                  requestId: frame.requestId,
+                  error: error instanceof Error ? error.message : String(error),
+                }),
+              ),
+          )
+          .finally(() => activeExecutions.delete(frame.requestId));
+      };
+      const consumeStdout = (text: string): void => {
+        stdoutBuffer += text;
```
markustherobot, line 127: **[P1] Bound the protocol frame buffer**

This parser runs from a second stdout listener and appends every chunk independently of `spawnWithCapture`'s 1 MiB capture buffer. If a command emits an unterminated line and ignores `SIGTERM`, the capture layer freezes and requests termination, but `stdoutBuffer` keeps growing throughout the two-second grace period, so a streaming command can exhaust the TUI despite the documented output cap. Track protocol-frame bytes and stop buffering/terminate as soon as the same bound is crossed.

src/tui/command_client_tools.ts
```diff
@@ -50,14 +126,39 @@ export function createCommandClientTools(
...
+          createTauClientToolCommandInvoke({
+            sessionId: context.sessionId,
+            callId: context.callId,
+            arguments: args,
+          }),
+        )}\n`,
+        onSpawn: (child) => {
+          observedStdout = true;
+          writeToCommand = (frame) => {
+            child.stdin?.write(`${JSON.stringify(frame)}\n`);
```
markustherobot: **[P1] Apply backpressure to execution responses**

Each completed `exec` can contain up to 24 MiB, but this writes the serialized response without checking the return value or awaiting `drain`, and incoming `exec` frames have no in-flight or queued-byte bound. A command that runs several environment commands concurrently can therefore make Node queue hundreds of megabytes on the TUI side faster than the child consumes stdin. Serialize these writes with backpressure and enforce a bounded response queue/concurrency limit, cancelling outstanding executions when the bound is exceeded.

src/sdk/client_tool_command.ts
```diff
@@ -0,0 +1,387 @@
...
+    const inputFailure = new Promise<never>((_resolve, reject) => {
+      rejectInput = reject;
+    });
+    void (async () => {
+      try {
+        while (true) {
+          const next = await iterator.next();
+          if (next.done) {
+            inputEnded = true;
+            rejectPending(new Error("client-tool command input closed during execution"));
```
markustherobot: **[P2] Settle the handler when protocol input closes**

When stdin reaches EOF after the invocation, this rejects only currently pending `exec()` calls. If the handler is awaiting other work, `inputFailure` never rejects and the controller is not aborted, so the detached command can remain alive indefinitely after its protocol peer disappears. Treat EOF as a terminal peer disconnect by aborting the handler and rejecting `inputFailure` (and keep stdin open in any one-shot version 2 harness until the result arrives).

markustherobot (commented):
I found material boundedness and lifecycle gaps in the version 2 process relay. I launched four subreviews covering the command protocol helper, TUI relay/process lifecycle, SDK/session and diff-review wiring, and code-mode/public API compatibility; all four completed successfully.

src/tui/command_client_tools.ts
markusylisiurunen: Fixed in 849faaea. Protocol frames are now capped at 1 MiB independently of capture, and a breach clears the parser buffer and aborts the command.

src/tui/command_client_tools.ts
markusylisiurunen: Fixed in 849faaea. Responses are serialized with stdin backpressure, and at most eight executions may remain active through response delivery; exceeding the limit cancels the outstanding work.

src/sdk/client_tool_command.ts
markusylisiurunen: Fixed in 849faaea. Protocol EOF now aborts the handler controller, rejects pending executions and the input race, and the one-shot harness keeps stdin open until result.

PR #424, opened 2026-08-09
support persistent Telegram directories

markusylisiurunen:
## why

Telegram projects currently always own generated repository workspaces. Personal or operational bots also need to work from an existing long-lived directory whose contents are shared across sessions and must never be removed by Tau.

## what

Add persistent directory projects through a required `directory` field. Every session for that project starts from the same existing directory, loads its normal Tau configuration and context, and omits repository metadata.

Tau treats these directories as externally owned: creation validates them without cloning or provisioning, restart recovery reuses them, startup orphan cleanup preserves them, and session closure or `/new` never removes them. Repository and composite project behavior remains unchanged, and composites continue to accept repository members only.

## details

Relative directory paths resolve from the Telegram config file. The directory must already exist, and concurrent sessions may use it without serialization.


src/core/telegram/session_manager.ts
```diff
@@ -903,7 +914,9 @@ class TelegramSessionManagerImpl implements TelegramSessionManager {
     }
 
     for (const project of Object.values(this.projects)) {
-      if (project.workspaceRoot) {
+      if (isDirectoryProject(project)) {
+        preservedWorkspacePaths.push(project.directory);
```
markustherobot, line 918: **[P1] Canonicalize persistent paths before startup cleanup**

`cleanupWorkspaceRootsOnStartup` compares these preserved paths to cleanup roots using only `resolve()`. If a workspace root reaches the same tree through a symlink alias (for example `/alias -> /real` with `directory: /real/me`), `/real/me` is not recognized beneath `/alias`, so startup pruning removes `/alias/me` and deletes the externally owned contents. Canonicalize existing roots and persistent directories with `realpath()` (or reject aliasing/overlapping ownership) before any recursive cleanup.

src/core/telegram/session_manager.ts
```diff
@@ -943,11 +956,16 @@ class TelegramSessionManagerImpl implements TelegramSessionManager {
...
 
-    let workspacePath = resolveWorkspacePath({
-      workspaceRoot: entry.project.workspaceRoot ?? this.workspaceRoot,
-      projectId: entry.record.projectId,
-      sessionId: entry.record.id,
-    });
+    const projectWorkspaceRoot = isDirectoryProject(entry.project)
+      ? this.workspaceRoot
+      : (entry.project.workspaceRoot ?? this.workspaceRoot);
+    let workspacePath = isDirectoryProject(entry.project)
```
markustherobot, line 962: **[P1] Verify the recovered Tau cwd before accepting the session**

This chooses and validates the directory from the current config, but `client.sessions.observe(tauSessionId)` restores the durable snapshot's original `executionEnvironment.cwd`. If a project directory changes from `/old` to `/new`, the manager reports recovery at `/new` while tool execution still targets `/old` (or later fails there), which can modify the wrong long-lived workspace. Compare the observed snapshot cwd with the configured directory and fail clearly on mismatch, or add an explicit migration that actually changes the execution environment.

markustherobot (commented):
Found two material ownership and recovery issues: startup cleanup can cross a persistent-directory boundary, and recovered sessions can silently retain a stale execution cwd. I launched 3 subreviews covering project config, workspace ownership/cleanup, and session lifecycle; all completed successfully.

src/core/telegram/session_manager.ts
markusylisiurunen: Reviewed and accepted as unsupported operator configuration. Managed workspace roots and persistent directories must not overlap through distinct symlink aliases; no code change planned.

src/core/telegram/session_manager.ts
markusylisiurunen: Fixed in 9ce58a5. Recovery now validates the durable Tau cwd against the configured persistent directory and fails clearly on mismatch, with regression coverage.

PR #426, opened 2026-08-10
add agent-scoped code-mode scratch files

### LINKED ISSUE #420: code-mode method(s) to read/write to files
in addition to the `console.*` methods to produce output, it would be helpful to allow code-mode code to read/write to files on disk. basic form of serde. one execution could for example write a json array to a file, then a later tool call could read it/parse it. the API should not be a full file system API but a narrow interface to interact with the file system in a scoped and restricted way.

markustherobot:
## why

Code-mode executions were isolated one-shot programs with no narrow way to hand text to later code-mode calls or to Bash. That made multi-step workflows depend on console output even when the execution environment already provided the right lifecycle boundary for temporary data.

## what

This adds an agent-scoped `files` global to Tau's built-in, SDK, and command-backed code-mode paths. It reads, atomically writes, lists, and removes real UTF-8 files under a deterministic execution-environment temporary directory derived from the owning agent ID. Returned paths are directly usable by Bash, later calls reopen the real files, names are restricted to basenames, and each agent receives a 128-file, 64 MiB quota.

The change also carries required `agentId` separately from `sessionId` through client-tool calls, SDK contexts, and version 3 command-tool invocation frames. Session protocol version 12 reflects the required wire field, and `ChatRuntime` now owns session identity explicitly instead of deriving it from main-agent identity.

## details

Scratch directories and files are owner-only, reject symlinks and non-regular targets, and use same-directory atomic replacement. Their contents are not persisted in session snapshots; recovery deterministically reaches the same directory when the execution environment retained its temporary filesystem.

Standalone `executeTauCodeMode()` callers opt in by supplying `{ agentId, adapter }`. Without that capability, `files` is neither exposed nor mentioned in generated runtime documentation. `files.list()` returns at most 128 file metadata entries plus total file and byte counts, so Bash-created over-quota contents remain observable without expanding the bridge response without bound. Removing a missing file is treated as an error.

No unresolved implementation questions remain.

fixes #420


src/code_mode/files.ts
```diff
@@ -0,0 +1,360 @@
...
+    if (stats.isSymbolicLink() || !stats.isFile()) continue;
+    files.push({ name, path, bytes: stats.size });
+  }
+  files.sort((left, right) => left.name.localeCompare(right.name));
+  return files;
+}
+
+function read(root, name) {
+  const path = getFilePath(root, name);
+  const descriptor = openSync(path, constants.O_RDONLY | constants.O_NOFOLLOW);
```
markustherobot: **[P2] Open special files without blocking**

`openSync(..., O_RDONLY)` runs before the `fstatSync().isFile()` check, so a Bash-created FIFO blocks here waiting for a writer and the intended non-regular-file rejection is never reached. I reproduced `files.read()` hanging until the code-mode timeout. Open nonblocking before validating the descriptor (or use an equivalent race-safe approach) so FIFOs and other special files fail promptly.

src/code_mode/files.ts
```diff
@@ -0,0 +1,360 @@
...
+function write(root, name, content) {
+  if (typeof content !== "string") fail("scratch file content must be a string");
+  const bytes = Buffer.byteLength(content, "utf8");
+  if (bytes > MAX_OPERATION_BYTES) fail("scratch file exceeds the 1 MiB write limit");
+
+  const path = getFilePath(root, name);
+  const existing = getExistingFile(path);
+  const files = listFiles(root);
+  const totalBytes = files.reduce((total, file) => total + file.bytes, 0);
+  const resultingFiles = files.length + (existing ? 0 : 1);
```
markustherobot, line 292: **[P2] Serialize quota accounting across executions**

The queue in `createTauCodeModeFilesApi()` is local to one code-mode execution, while the quota is agent-wide. Two concurrent executions for the same agent can both observe 127 files, both pass this calculation, and both rename successfully; I reproduced a directory ending with 129 files. Protect the quota calculation and commit with a target-side lock or another transaction shared by all processes using this agent root.

src/code_mode/files.ts
```diff
@@ -0,0 +1,360 @@
...
+  const resultingBytes = totalBytes - (existing?.size ?? 0) + bytes;
+  if (resultingFiles > MAX_FILES) fail("scratch directory exceeds the 128-file limit");
+  if (resultingBytes > MAX_TOTAL_BYTES) fail("scratch directory exceeds the 64 MiB total limit");
+
+  const temporaryPath = join(root, ".tau-write-" + randomUUID());
+  let descriptor;
+  try {
+    descriptor = openSync(
```
markustherobot, line 300: **[P2] Recover abandoned staging files**

If cancellation, timeout, or process termination lands after this file is created but before the `finally` block runs, `.tau-write-*` remains in the scratch root. Because `listFiles()` treats it as an ordinary user file, it is exposed by `files.list()` and consumes file and byte quota after recovery. Keep staging artifacts outside the public namespace or clean stale staging files before listing and quota accounting.

src/host/local_session_host.ts
```diff
@@ -310,7 +310,7 @@ export class LocalSessionHost implements TauSessionHost {
       createdAt,
       persona: bootstrap.persona,
       backend: executionEnvironment.getToolExecutionBackend(),
-      clientTools: (sessionId) => this.clientToolBroker.getToolDefinitions(sessionId),
+      clientTools: () => this.clientToolBroker.getToolDefinitions(sessionId),
```
markustherobot, line 313: **[P2] Rebind client tools when the session ID resets**

This closure permanently binds definitions to the original `sessionId`, but `ChatRuntime.reset()` now assigns a new session ID and rebuilds the registry. After reset, retained client tools dispatch with the old routing ID and `executionEnvironment.exec()` targets a session the host has removed, while detached old definitions can also disappear entirely. Either keep the explicit session identity stable across an agent reset or re-key client attachment and resolve definitions from the runtime's current session ID.

src/core/tools/nook.ts
```diff
@@ -397,6 +398,7 @@ function executeNookProgram(
       },
     },
     code,
+    agentId,
```
markustherobot, line 401: **[P2] Update Nook's docs for the new files capability**

Passing `agentId` here makes `executeInternalCodeMode()` expose `files`, but `src/core/static/code_mode/nook/documentation.md:1` still says Nook methods are the only platform capability and `src/nook/README.md:123` still says generated code has no filesystem access. Since Nook requires agents to read these docs before use, the documentation directly contradicts the injected API. Update the Nook-specific guidance to distinguish the bounded scratch `files` capability from ambient filesystem access.

src/tui/command_client_tools.ts
```diff
@@ -176,6 +176,7 @@ export function createCommandClientTools(
         input: `${JSON.stringify(
           createTauClientToolCommandInvoke({
             sessionId: context.sessionId,
+            agentId: context.agentId,
```
markustherobot, line 147: **[P3] Update the remaining v2 diagnostic**

This invocation now speaks command protocol v3, but the successful-exit-without-result path at line 246 still reports `returned no version-2 result frame`. That sends integration authors to the wrong contract while debugging a malformed v3 command tool; update the diagnostic to version 3 and cover it in the command-tool tests.

markustherobot (commented):
Found several issues to address before merge, centered on scratch-file safety and lifecycle, quota correctness, session-reset routing, and integration guidance. I launched four subreviews covering filesystem safety, the code-mode bridge, identity propagation, and integration/docs; all four completed successfully.

src/code_mode/files.ts
markusylisiurunen: We are intentionally keeping the scratch quota cooperative rather than adding cross-process lock and stale-lock recovery machinery for a harmless rare overshoot. `files.list()` no longer truncates at 128 entries: it returns every regular file present, subject only to the shared 1 MiB code-mode response limit. Regression coverage verifies that all 129 files are reported.

src/code_mode/files.ts
markusylisiurunen: Fixed by moving atomic-write staging into a separate owner-validated, agent-scoped sibling directory on the same temporary filesystem. Interrupted `.tau-write-*` artifacts therefore remain outside the public scratch namespace and quota accounting while rename remains atomic. Regression coverage simulates an abandoned staging artifact and verifies that `files.list()` does not expose it.

src/host/local_session_host.ts
markusylisiurunen: Resolved by removing the obsolete runtime reset/rekey lifecycle rather than adding routing migration machinery. `ChatRuntime.reset()` had no production callers after `/new` and the old RPC reset operation moved to host-created sessions, and it did not coherently reset host-owned state. The otherwise-unused `AgentRuntime.reset()`, host snapshot-rekey bookkeeping, obsolete test, and now-dead session-store deletion contract were removed as well. Session IDs and client-tool routing are now stable for each hosted session.

PR #429, opened 2026-08-10
enable command client tools in Telegram

### LINKED ISSUE #427: make configured client tools available to Telegram sessions
## why

Telegram sessions currently cannot use configured command-backed client tools. Project-level `enabledClientTools` can select global definitions, but the Telegram runner creates SDK sessions without a client that advertises or executes those tools. The same workspace therefore has different tool capabilities in the TUI and Telegram even when its Tau configuration selects the same tools.

Personal and operational Telegram workspaces need access to integrations such as Gmail, Google Calendar, and Todoist without requiring an attached TUI.

## what

Provide a canonical way for the Telegram runner to make the configured client tools selected by each project's `enabledClientTools` available to its sessions. The implementation may let the Telegram-owned SDK client advertise and execute command tools, or establish another ownership model suitable for non-TUI clients, but the resulting tools should use the same definitions and execution contract as TUI client tools.

## details

The solution should preserve the existing command-tool boundaries and behavior: global-only executable definitions, project allowlist selection, argument validation, owning session and agent identities, execution-environment access, cancellation, process-group cleanup, protocol limits, and shutdown handling. It should not require a TUI process to remain attached to a Telegram session.

markusylisiurunen:
## why

Telegram sessions currently create SDK clients without advertising configured command-backed client tools. This leaves the same project with different integrations depending on whether it is used from the TUI or Telegram, and forces command execution to remain coupled to TUI wiring.

## what

Move the existing command client-tool runner to a shared client-neutral module and use it from both the TUI and Telegram. Telegram now resolves normal Tau configuration from each prepared session workspace, advertises that workspace's selected command tools through its SDK client, and keeps TUI-only tools out of the Telegram client.

## details

The existing SDK delegation and command execution contracts remain unchanged, including schema validation, execution-environment access, cancellation, process-group cleanup, timeouts, and protocol limits. Workspaces without an explicit selection receive default-enabled definitions, exact `enabledClientTools` allowlists are respected, and an empty allowlist disables configured tools. Reconnected sessions resolve the selection again from their recovered workspace.

fixes #427


src/core/telegram/local_session_client.ts
```diff
@@ -0,0 +1,37 @@
...
+};
+
+export async function createLocalTelegramSessionClient(options: {
+  client: TelegramSessionClientOptions;
+  hostConfig: Config;
+  configDeps: ConfigDeps;
+  deps?: Partial<LocalTelegramSessionClientDependencies>;
+}): Promise<TelegramSessionClient> {
+  const deps = { ...defaultDependencies, ...options.deps };
+  const workspaceConfig = deps.loadConfig(options.client.cwd, options.configDeps);
```
markustherobot, line 26: **[P1] Preserve global client tools outside `HOME`**

`loadConfig()` deliberately omits `~/.config/tau/config.json` when this prepared cwd is outside the runner user's home. The documented Telegram deployment uses `/etc/tau/telegram.json` with `/var/lib/tau/telegram-workspaces`, so those sessions never see any global `clientTools` definitions and advertise no tools, even if their project config selects them. Resolve the runner-owned global definitions independently, then apply the workspace's project-level selection, and cover an outside-`HOME` workspace in the regression test.

README.md
```diff
@@ -657,7 +657,7 @@ see [docs/telegram.md](docs/telegram.md) for the config schema, project selectio
 
 ### command client tools
 
-Command client tools extend the standard TUI without modifying Tau. Define them only in `~/.config/tau/config.json`; Tau requires each parameter schema to be an object with root `type: "object"`, trusts the remaining schema contents, advertises the effective tools to the session host, and runs their commands on the machine where `tau` or `tau attach` is running. Every definition must set `defaultEnabled`. Without a workspace selection, only default-enabled tools are advertised.
+Command client tools extend Tau clients without modifying the session host. Define them only in `~/.config/tau/config.json`; Tau requires each parameter schema to be an object with root `type: "object"`, trusts the remaining schema contents, advertises the effective tools to the session host, and runs their commands on the owning TUI or Telegram runner machine. Every definition must set `defaultEnabled`. Without a workspace selection, only default-enabled tools are advertised.
```
markustherobot, line 660: **[P2] Update the rest of this section for Telegram ownership**

This paragraph now covers Telegram, but lines 691-693 still say every command runs on the TUI machine, from the TUI cwd, with the TUI environment. Telegram commands instead run from the Telegram runner's cwd with that process's environment, so command authors can make incorrect path and environment assumptions. Use owning-client wording throughout those two paragraphs.

markustherobot (commented):
I found one deployment-blocking configuration gap and one documentation inconsistency. I launched three focused subreviews covering configuration/security boundaries, SDK lifecycle/reconnect behavior, and the refactor/test/package surface; all completed successfully, and I verified the reported issues against the repository before including them here.

src/core/telegram/local_session_client.ts
markusylisiurunen: This is intentional. Telegram resolves configuration from the prepared workspace exactly as if a local TUI were launched there, so a workspace outside HOME should not inherit global configuration.

README.md
markusylisiurunen: Fixed in 02d2c37b. The detailed command documentation now uses owning-client wording and distinguishes the TUI and Telegram opt-outs.

PR #430, opened 2026-08-10
add built-in Tau documentation

### LINKED ISSUE #428: built-in shipped documentation for Tau
the Tau agent should have access to built-in Tau documentation so it knows how to manage and configure itself.

markusylisiurunen:
## why

Tau agents need reliable documentation for using, managing, and configuring the installed Tau version. The existing README and topic guides mixed operational guidance with reference and maintainer material, while remote execution boundaries meant agents could not safely rely on host filesystem paths.

## what

This adds one canonical documentation corpus for both people and agents. The new `docs/tau/` manual is organized around user intent, configuration scope, and the client, host, execution-environment, and Telegram-runner boundaries. README is now a concise landing page, and the overlapping model and Telegram pages point into the canonical corpus.

Tau now packages the corpus with each build and exposes it through an intrinsic host-owned `tau_docs({ path })` tool for main agents and subagents. Agents begin with `index.md` and follow exact flat Markdown links, without a search/list API or system-prompt changes. Build and corpus checks enforce the manifest, flat paths, valid links, package inclusion, and a 24 KiB limit per document.

fixes #428


src/core/tools/catalog.ts
```diff
@@ -96,6 +97,7 @@ export const ToolCatalog = {
     const enabledToolNames = new Set<string>(options.persona.tools);
     return new ToolRegistry([
       ...tools.filter((tool) => enabledToolNames.has(tool.schema.name)),
+      createTauDocsToolDefinition(),
```
markustherobot, line 100: **[P2] Reserve the intrinsic name from client tools**

A client can still advertise a tool named `tau_docs` because `ClientToolBroker.HOST_TOOL_NAMES` was not updated. Registration succeeds, but when the next turn merges that definition with this always-present one, `ToolRegistry` throws `duplicate tool 'tau_docs'` and the session cannot run. Reserve `TOOL_NAME_TAU_DOCS` in the broker and add it to the collision test.

docs/config-reference.md
```diff
@@ -0,0 +1,422 @@
...
+
+The default is `false`. When enabled, at least one valid custom persona must be available or session startup fails. This field does not disable user or project personas.
+
+### `agentContextFiles`
+
+An array of non-empty paths to additional `AGENTS.md`-style context files:
+
+```json
+{
+  "agentContextFiles": ["docs/AI_GUIDE.md", "services/payments/AGENTS.md"]
```
markustherobot, line 83: **[P2] Use an actual `AGENTS.md` path**

The runtime context resolver rejects any configured file whose basename is not exactly `AGENTS.md`, so `docs/AI_GUIDE.md` is silently omitted rather than injected. The same invalid example is repeated in `configuration.md`. Replace it with an `AGENTS.md` path and state here that this is not a general arbitrary-file include.

docs/configuration.md
```diff
@@ -0,0 +1,210 @@
...
+}
+```
+
+Tau resolves the paths as `/home/ada/work/ledger/docs/AI_GUIDE.md` and `/home/ada/work/ledger/tools/review-ui`. In an attached session, `diffTool` used by the TUI comes from the client-local load, so this resolved command is a client path.
+
+Other strings that happen to contain paths are not automatically rebased unless their field contract says so. In particular, hosted execution-environment `home` values are passed as configured.
+
+## Invalid and unknown fields
+
+Each `config.json` must contain a JSON object. Tau reports malformed JSON, wrong types, invalid enum values, unknown model targets, and fields used at a forbidden scope as startup warnings. It keeps valid fields from the same file and continues merging other levels.
```
markustherobot: **[P2] Do not promise startup warnings that session creation drops**

When `tau serve`, `tau rpc`, or the SDK creates a session from an execution environment, `resolveRuntimeConfig()` returns diagnostics but both hosted bootstrap paths discard `runtime.warnings`; prompt-context `unknownSkills` are also ignored until `/reload`. Invalid target-side config, model, persona, or skill selection can therefore fall back silently despite this guarantee. Either surface those diagnostics during creation or scope the warning claim to local startup and reload.

docs/credentials.md
```diff
@@ -0,0 +1,200 @@
...
+tau auth disable codex --account developer@example.com
+tau auth enable codex --account developer@example.com
+tau auth logout codex --account developer@example.com
+```
+
+Login adds a new account or refreshes the matching account. Disable keeps the account stored but removes it from automatic selection. Logout removes it.
+
+### Account selection and failover
+
+Without an override, Tau chooses among enabled accounts with usable quota and keeps the selected account stable for a session. If a provider error reveals that the selected account's tracked quota is exhausted, Tau clears that session selection so a later attempt can choose another usable account. Disabled, expired, unrefreshable, or exhausted accounts are not suitable candidates.
```
markustherobot: **[P2] Keep refreshable expired accounts eligible**

An expired Codex access token is not itself an unsuitable account: account selection calls `getApiKeyForAccount()`, which refreshes expired credentials before checking quota and can select the account normally. This wording tells users to treat a recoverable state as unusable. Remove `expired` here or qualify it as expired credentials that cannot be refreshed.

docs/tau/telegram.md
```diff
@@ -0,0 +1,342 @@
...
+
+These command processes run on the Telegram runner machine with the runner process environment. They can reach the session workspace only through their explicit execution-environment facade, despite physical co-location. Telegram does not advertise TUI-only `diff_review` or `prefill_input` tools.
+
+Tool selection occurs when the session client is created or reconnected. `/reload` does not rebuild the Telegram client advertisement. Restart the runner or create a new session client after changing command client-tool selection. See [client tools](client-tools.md).
+
+## Persistence and restart behavior
+
+Normal shutdown interrupts live work, waits for owned activity to settle, disconnects session clients, and persists active records as waiting for input. It preserves repository and composite workspaces rather than deleting them. On restart, Tau reconnects those records to their durable session snapshots.
```
markustherobot: **[P2] Qualify recovery for sessions still being prepared**

Shutdown only changes `running` records to `waiting-input`. A `queued` or `preparing-workspace` record without a persisted Tau session ID remains in that transient state, and startup calls `initializeSession()` again rather than reconnecting a durable snapshot. This is reachable during a slow clone or client/session creation, so limit this promise to records with a persisted Tau session ID and describe reinitialization for preparation-stage records.

markustherobot (commented):
I found several correctness gaps across the intrinsic-tool boundary and the new canonical documentation contracts. I launched five subreviews covering runtime integration, packaging, configuration, session/tool behavior, and History/Nook/Telegram/security; all completed, and I verified the retained findings against the repository.

src/core/tools/catalog.ts
markusylisiurunen: Already addressed in 59bf3338: the broker uses the shared HOST_TOOL_NAMES list, which includes tau_docs, and the collision test covers every reserved host name.

docs/config-reference.md
markusylisiurunen: Changed the contract instead: explicit agentContextFiles now accepts any eligible text filename, while automatic discovery remains AGENTS.md-only. Added runtime and TUI coverage and updated the docs.

docs/configuration.md
markusylisiurunen: Scoped this to observable behavior: /reload is now the documented way to see current configuration warnings, and the broad startup-warning promise is removed.

docs/credentials.md
markusylisiurunen: Updated the wording so refreshable expired credentials remain eligible. Only credentials that cannot be refreshed are excluded.

docs/tau/telegram.md
markusylisiurunen: Updated the docs to distinguish completed sessions, which reconnect, from sessions interrupted during preparation, which prepare again. The explanation is now user-facing rather than state-machine-specific.

PR #432, opened 2026-08-10
remove rpc transport

markusylisiurunen:
## why

Tau already has a complete remote-session path through `tau serve` and WebSocket clients. Keeping a second stdio RPC path adds subprocess lifecycle, framing, public API, CLI, test, and documentation surface for a connection mode we no longer want to support.

## what

Remove the `tau rpc` mode, stdio session transport, process-specific transport errors, and command-based `tau attach -- ...` form. `tau attach` now connects through WebSocket only.

Keep the session protocol and SDK transport contract independent of WebSocket so another transport can still be introduced without changing session semantics. Move the shared protocol-handler coverage out of the deleted RPC server test owner, retain generic custom-transport SDK coverage, and update the public and contributor documentation to the canonical connection model.


docs/ownership-and-scope.md
```diff
@@ -39,9 +39,9 @@ For repository and composite projects it prepares managed workspaces. A configur
 | `tau` | Local TUI process | In-process on the same machine | Local `cwd` where Tau started |
 | `tau attach … ws://…` | Machine running `tau attach` | Machine running `tau serve` | Environment selected or restored by that host |
 | `tau attach … -- <command>` | Machine running `tau attach` | Machine running the protocol command, often reached through SSH | Environment selected or restored by that host |
-| `tau rpc` or `tau serve` | A separate protocol client | The server process | Local or configured hosted environment chosen by the client |
+| `tau serve` | A separate protocol client | The server process | Local or configured hosted environment chosen by the client |
```
markustherobot, line 41: **[P2] Remove the obsolete command-attachment row**

The table still advertises `tau attach … -- <command>` immediately above this line, but `parseAttachArgs` now rejects `--` and the new CLI regression test codifies that removal. Delete the stale row so the ownership guide does not send users to an unsupported connection mode.

markustherobot (commented):
Found one documentation inconsistency in an otherwise coherent removal. The runtime and public API cleanup align with the intended WebSocket-only built-in connection model, retained protocol coverage is sound, and CI is green. I launched three subreviews covering CLI/TUI behavior, the SDK/package surface, and protocol tests/documentation; all three completed successfully.

PR #433, opened 2026-08-11
render granular client tool presentations

### LINKED ISSUE #422: allow client tools to render more granular UI, such as the subject
now, for example, client code-mode tools don't really render very nicely and the code cannot be seen.

markusylisiurunen:
## why

Client tools currently render with only their registered name, hiding invocation-specific context such as submitted code, review commands, or draft text. The owning client has that context, but presentation must remain optional so ordinary client tools can rely entirely on Tau's defaults.

## what

Client tools may provide partial presentation overrides containing `subject`, `subjectWrap`, `details`, and `metadata`. Acknowledgement presentation applies while the call runs, while result presentation independently applies to the terminal state. Tau owns lifecycle actions and operation, supplies every omitted field, and uses a complete fallback when no client result arrives.

Explicit presentation text is preserved unchanged after protocol safety validation. Tau's canonical display truncation applies only to generated defaults. The SDK and code-mode packages expose a generic, opt-in `truncateTauClientToolText` helper for clients that want concise text.

The SDK's `describe` callback is optional and has no execution-environment authority. Successful results are accepted only after the host has completed acknowledgement authorization, while explicit preparation failures may be reported before acknowledgement. Code mode, diff review, and prefill input provide invocation subjects for both running and terminal cards.

Command client tools use the breaking version 4 flow `prepare -> ready(presentation?) -> execute -> result(ok, presentation?)`, with execution-environment requests available only after authorization. The coordinated session wire contract advances to protocol version 13 without compatibility shims.

The branch also retains the independent contributor policy that agents do not query or report remote CI unless explicitly requested.

fixes #422


src/protocol/session_protocol.ts
```diff
@@ -239,8 +239,29 @@ export type SessionProtocolEphemeralSubmitParams = SessionProtocolSessionIdParam
...
+  details: Array<{
+    text: string;
+    tone?: "added" | "removed";
+    wrap: "word" | "character";
+  }>;
+  metadata: string[];
+};
 export type SessionProtocolClientToolAckParams = SessionProtocolSessionIdParams & {
   callId: string;
+  presentation: SessionProtocolClientToolPresentation;
```
markustherobot: **[P1] Advance the session protocol version**

This makes `presentation` required while `SESSION_PROTOCOL_VERSION` remains 12. A pre-change client therefore still negotiates successfully with a new host, then every client-tool acknowledgement is rejected as invalid params and the call times out. Increment the session protocol version (and the versioned method docs/tests) so incompatible peers fail during negotiation instead.

src/sdk/session.ts
```diff
@@ -178,25 +179,30 @@ class TauSdkClientImpl implements TauSdkClient {
...
+      signal: abortController.signal,
+      executionEnvironment: this.createClientToolExecutionEnvironment(
+        message.sessionId,
+        abortController.signal,
+      ),
+    };
 
     try {
-      const result = await tool.execute(message.arguments, {
+      const presentation = await tool.describe(message.arguments, context);
```
markustherobot: **[P1] Keep execution authority out of `describe`**

`describe` receives the full context, including `executionEnvironment`, before the acknowledgement has been accepted. If a descriptor calls `context.executionEnvironment.exec()` to derive its subject, workspace work runs before `tool_call_started` is committed and still runs when the presentation is rejected or persistence fails. Pass a reduced describe context without the execution-environment facade, and construct the execution context only after `accepted: true`.

src/sdk/client_tool_command.ts
```diff
@@ -391,11 +484,20 @@ export async function runTauClientToolCommand(handler: TauClientToolCommandHandl
...
     };
 
+    const presentation = await definition.describe?.(preparation.arguments);
+    readySent = true;
+    await writeFrame({
+      version: TAU_CLIENT_TOOL_COMMAND_PROTOCOL_VERSION,
+      type: "ready",
+      ...(presentation === undefined ? {} : { presentation }),
+    });
+    await Promise.race([executionReady, inputFailure]);
```
markustherobot: **[P1] Make command preparation observe termination**

Both the direct `await definition.describe?.(...)` and this authorization wait ignore `controller.signal`. Because the helper installs SIGINT/SIGTERM handlers, a signal received during an asynchronous descriptor or after `ready` suppresses Node's default termination but never settles either wait, leaving the command alive indefinitely while stdin remains open. Race the preparation waits against input failure and controller abort so cleanup always runs.

src/host/client_tool_broker.ts
```diff
@@ -97,14 +110,38 @@ export class ClientToolBroker {
...
+      await pending.emitActivity({
+        type: "tool_call_started",
+        toolCallId: pending.toolCall.id,
+        toolName: pending.toolCall.name,
+        presentation: parsedPresentation,
+      });
+      if (pending.settled || this.pendingCalls.get(callId) !== pending) {
+        return false;
+      }
+      pending.presentation = parsedPresentation;
```
markustherobot: **[P2] Preserve the presentation before awaiting persistence**

The client presentation is stored only after `emitActivity` resolves. If interruption, detach, or the execution timer settles the call while that awaited mutation is queued, `complete()` resolves with the generic tool-name presentation, so the finished card loses the subject and operation that the started activity carried. Store the parsed presentation before awaiting the activity while still withholding acknowledgement acceptance until the await succeeds.

src/host/client_tool_broker.ts
```diff
@@ -97,14 +110,38 @@ export class ClientToolBroker {
...
+      return true;
+    } catch (error) {
+      this.complete(
```
markustherobot: **[P2] Cancel prepared work when acknowledgement persistence fails**

When `emitActivity` rejects, `complete()` removes the call without sending `session.clientTool.cancel`. The acknowledgement request returns an internal error, the SDK's catch path does not abort its controller, and a command process that already emitted `ready` remains blocked until its potentially very long execution timeout. Settle this through a path that also notifies the client, or explicitly abort client-side state on acknowledgement failure.

src/protocol/session_protocol.ts
```diff
@@ -1910,10 +1931,55 @@ const sessionProtocolEphemeralCloseParamsSchema = z
...
+      operation: clientToolPresentationLabelSchema.optional(),
+      subject: z
+        .string()
+        .min(1)
+        .refine((value) => {
+          const lines = value.split("\n");
+          return lines.length <= 8 && lines.every((line) => Array.from(line).length <= 512);
+        }, "must contain at most 8 lines of 512 characters"),
+      subjectWrap: z.enum(["word", "character"]),
+      details: z.array(
```
markustherobot: **[P2] Bound client-supplied detail collections**

This schema bounds the subject and labels but allows unlimited detail text, detail entries, and metadata entries. `buildTauClientToolPresentation({ detailTruncation: false })` can therefore create a multi-megabyte acknowledgement that validates and is persisted and broadcast in `tool_call_started`, contrary to the bounded presentation contract. Add count or aggregate UTF-8 byte limits here and enforce the same limits in the public builder/canonical parser.

src/core/tools/presentation.ts
```diff
@@ -261,6 +267,37 @@ function truncateLines(
   ];
 }
 
+export function truncateToolRunSubject(
+  subject: string,
+  options: ToolCardSubjectTruncation = {},
+): string {
+  const maxLines = options.maxLines ?? TOOL_CARD_SUBJECT_MAX_LINES;
+  const maxLineChars = options.maxLineChars ?? TOOL_CARD_MAX_LINE_CHARS;
+  if (!Number.isInteger(maxLines) || maxLines < 1 || maxLines > TOOL_CARD_SUBJECT_MAX_LINES) {
```
markustherobot, line 336: **[P2] Handle one-line middle truncation**

`maxLines: 1` is accepted here, but the default middle strategy computes `tailCount === 0` and `truncateLines` then calls `bounded.slice(-0)`, which returns the entire input. For example, truncating three lines with `{ maxLines: 1 }` returns the omission marker plus all three lines. Special-case the one-line/zero-tail case so the exported helper honors its requested bound.

markustherobot (commented):
I found compatibility and lifecycle issues in the new describe/acknowledge gate, plus boundary gaps in the public presentation helpers. I launched 4 subreviews covering host/protocol lifecycle, command process flow, SDK presentation contracts, and tests/docs, then revalidated their candidates and the two follow-up commits against current head `c225b7e`. The build and focused tests pass, but the inline issues remain.

src/protocol/session_protocol.ts
markusylisiurunen: Fixed in 3d5ad190. The session protocol is now v13, including negotiation fixtures, integration assertions, and the versioned protocol documentation.

src/sdk/session.ts
markusylisiurunen: Fixed in 3d5ad190. `describe` now receives `TauSdkClientToolDescribeContext` without execution-environment authority, and the full execution context is constructed only after the acknowledgement is accepted.

src/sdk/client_tool_command.ts
markusylisiurunen: Fixed in 3d5ad190. Input, asynchronous description, authorization, and handler waits now race controller abort. SIGINT/SIGTERM regressions cover both description and post-ready authorization waits.

src/host/client_tool_broker.ts
markusylisiurunen: Fixed in 3d5ad190. The parsed presentation is stored before awaiting the started-activity mutation while acknowledgement acceptance still waits for that mutation to commit.

src/host/client_tool_broker.ts
markusylisiurunen: Fixed in 3d5ad190. A failed acknowledgement mutation now settles through `fail()` and sends `session.clientTool.cancel` with the new `host-failed` reason. The regression also verifies prepared presentation is preserved.

src/protocol/session_protocol.ts
markusylisiurunen: Fixed in 3d5ad190. Protocol and canonical parsing now enforce large UTF-8 safety limits for the aggregate presentation, every field, and collection counts. Raw command presentation validation uses the same high subject and label limits.

src/core/tools/presentation.ts
markusylisiurunen: Fixed in 3d5ad190. Zero-tail middle truncation no longer uses `slice(-0)`, and the exported helper now has a `maxLines: 1` regression.

markusylisiurunen:
@markustherobot review

src/host/client_tool_broker.ts
```diff
@@ -325,6 +400,33 @@ function createClientToolPresentation(toolName: string, content: string, duratio
...
+    details: defaults.details,
+    detailTruncation: false,
+    metadata: defaults.metadata,
+  });
+  return {
+    ...resolved,
+    subject: presentation?.subject ?? resolved.subject,
+    subjectWrap: presentation?.subjectWrap ?? resolved.subjectWrap,
+    details:
+      presentation?.details?.map((line) => ({ ...line, wrap: line.wrap ?? "word" })) ??
```
markustherobot, line 427: **[P2] Revalidate the resolved presentation size**

An override containing 1,024 details of 1,000 bytes each is below the protocol's 1 MiB aggregate cap, but this mapping adds `wrap: "word"` to every entry after that validation, and the canonical lifecycle fields add further bytes. The resulting activity exceeds `SESSION_PROTOCOL_MAX_CLIENT_TOOL_PRESENTATION_BYTES`; `parseToolRunPresentation`, which the TUI uses for current-version facets, rejects it, so a protocol-valid acknowledgement can later break rendering. Validate the final merged object, or reserve enough canonical overhead at the protocol boundary so every accepted override resolves to a valid `ToolRunPresentation`.

markustherobot (commented):
I found one presentation-boundary issue in the otherwise coherent authorization and phase-specific override flow. I launched 4 subreviews covering host/protocol lifecycle, command process flow, SDK/presentation APIs, and tests/docs integration; the SDK and integration subreviews failed because the provider was overloaded, and I independently covered those surfaces before publishing.

src/host/client_tool_broker.ts
markusylisiurunen: Fixed in 8b3f9a26. The untrusted partial override remains capped at 1 MiB, while a fully resolved canonical `ToolRunPresentation` now has a separate 2 MiB bound that accounts for Tau-added action labels, operation, and default wraps. The broker revalidates the merged presentation before emitting it. Regressions cover the reported 1,024 × 1,000-byte detail override through protocol acceptance, host resolution above 1 MiB, and canonical/TUI parsing.

PR #434, opened 2026-08-11
add telegram voice responses

markusylisiurunen:
## why

Telegram users can read final assistant responses in chat, but there is no persistent way to receive those responses as audio when listening is more convenient.

## what

Add `/tts_on` and `/tts_off` with a per-bot, per-chat preference that survives project changes, new sessions, and runner restarts. After a successful logical response settles, the runner keeps text authoritative and asynchronously sends a Gemini-generated Telegram voice note.

## details

The shared speech pipeline rewrites responses into plain spoken paragraphs with `gemini-3.6-flash`, then synthesizes them with `gemini-3.1-flash-tts-preview` and the Despina voice. TUI playback uses 500-character chunks at concurrency six; Telegram uses 1,000-character chunks at concurrency three. Both preserve paragraph and Unicode boundaries, use a brisk synthesis direction, and apply pitch-preserving 1.1× playback.

Source and rewritten text are each limited to 10,000 Unicode characters, cumulative PCM to 32 MiB, and every TTS request to 8,192 output tokens. Rewriting has a one-minute deadline, conversion has two minutes, and the complete Telegram job has five minutes through upload retries.

Telegram validates each ordered WAV chunk and writes it to a controlled temporary file. One runner-side `ffmpeg` concat operation produces Ogg Opus, and startup removes temporary directories left by crashes. Voice jobs are serialized per session, remain live-only across restarts, and recheck opt-in state before delivery.

Generation, conversion, timeout, size, and exhausted upload failures leave text unaffected. While voice responses remain enabled, the bot sends `voice response failed. please try again.` and keeps detailed diagnostics in runner logs.

Existing version 1 project preferences migrate lazily to the combined version 2 project and TTS preference shape, preserving the selected project and defaulting TTS off. The two added commands reduce the visible project-command capacity from 92 to 90 per bot.

### DIFF
--- a/src/core/telegram/adapter.ts
+++ b/src/core/telegram/adapter.ts
@@ -255,6 +255,7 @@ const MAX_TELEGRAM_ATTACHMENTS_PER_TURN = 10;
 const MAX_TELEGRAM_ATTACHMENT_FILE_BYTES = 20 * 1024 * 1024;
 const MAX_TELEGRAM_ATTACHMENT_TOTAL_BYTES = 50 * 1024 * 1024;
 const MAX_TELEGRAM_VOICE_BYTES = 50 * 1024 * 1024;
+const TELEGRAM_TTS_JOB_TIMEOUT_MS = 5 * 60_000;
 const MAX_TELEGRAM_GROUP_PENDING_MESSAGES = 50;
 const TELEGRAM_ATTACHMENT_TEMP_DIR_PREFIX = "tau-telegram-attachments-";
 const TELEGRAM_REASONING_EFFORTS = [
@@ -951,10 +952,6 @@ class TelegramDeliveryError extends Error {
   }
 }
 
-function isAbortError(error: unknown): boolean {
-  return error instanceof Error && error.name === "AbortError";
-}
-
 function getErrorCode(error: unknown): string | undefined {
   if (typeof error !== "object" || error === null || !("code" in error)) {
     return undefined;
@@ -1292,7 +1289,6 @@ class TelegramAdapterImpl {
   private readonly chatTypesByChat = new Map<number, string>();
   private readonly typingIntervalsBySessionChat = new Map<string, ReturnType<typeof setInterval>>();
   private readonly lastCommandBySession = new Map<string, string>();
-  private readonly finalAssistantTextBySession = new Map<string, string>();
   private readonly pendingAttachmentsBySession = new Map<string, TelegramPendingAttachment[]>();
   private readonly pendingGroupMessagesByChat = new Map<number, TelegramGroupPendingMessage[]>();
   private readonly pendingAttachmentTempDirBySession = new Map<string, string>();
@@ -1301,6 +1297,7 @@ class TelegramAdapterImpl {
   private readonly inFlightUpdateTasks = new Set<Promise<void>>();
   private readonly notificationQueueTailByChat = new Map<number, Promise<void>>();
   private readonly inFlightNotificationTasks = new Set<Promise<void>>();
+  private readonly ttsQueueTailBySession = new Map<string, Promise<void>>();
   private readonly inFlightTtsTasks = new Set<Promise<void>>();
 
   private readonly unsubscribeSessionEvents: () => void;
@@ -1386,6 +1383,7 @@ class TelegramAdapterImpl {
       await this.loopPromise;
       await this.waitForInFlightUpdateTasks();
       await Promise.allSettled(Array.from(this.inFlightTtsTasks));
+      this.ttsQueueTailBySession.clear();
       await Promise.allSettled(Array.from(this.inFlightNotificationTasks));
 
       for (const sessionId of Array.from(this.chatsBySession.keys())) {
@@ -2962,6 +2960,13 @@ class TelegramAdapterImpl {
       return;
     }
 
+    if (event.type === "session-response-completed") {
+      if (this.chatsBySession.has(event.sessionId)) {
+        this.startTtsNotification(event.sessionId, event.text);
+      }
+      return;
+    }
+
     if (event.type === "session-turn-failed" || event.type === "session-turn-rejected") {
       if (!this.chatsBySession.has(event.sessionId)) {
         return;
@@ -3010,7 +3015,6 @@ class TelegramAdapterImpl {
     }
 
     if (event.state === "running") {
-      this.finalAssistantTextBySession.delete(event.sessionId);
       this.startTypingIndicators(event.sessionId);
       if (this.isVerboseSession(event.sessionId)) {
         this.notifyLifecycle(event.sessionId, event.projectId, "started");
@@ -3019,7 +3023,6 @@ class TelegramAdapterImpl {
     }
 
     if (event.state === "failed") {
-      this.finalAssistantTextBySession.delete(event.sessionId);
       this.stopTypingIndicators(event.sessionId);
       this.notifyLifecycle(event.sessionId, event.projectId, "failed");
       return;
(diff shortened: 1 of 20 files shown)


src/core/telegram/adapter.ts
```diff
@@ -3020,6 +3106,60 @@ class TelegramAdapterImpl {
...
+      }
+
+      const chatIds = this.chatsBySession.get(sessionId);
+      if (!chatIds) {
+        return;
+      }
+      await Promise.all(
+        [...chatIds]
+          .filter((chatId) => this.projectPreferences.isTtsEnabled(this.ownerIdForChat(chatId)))
+          .map(async (chatId) => await this.enqueueVoiceNotification(sessionId, chatId, voice)),
```
markustherobot: **[P1] Preserve response order across TTS jobs**

When two turns finish while the first Gemini conversion is still running, each generation races independently and this line queues whichever voice finishes first. The chat can therefore receive response A text, response B text, response B voice, then response A voice. Keep a per-chat or per-session FIFO for voice jobs so speech delivery follows response order without blocking text notifications.

src/core/telegram/adapter.ts
```diff
@@ -2961,6 +3038,11 @@ class TelegramAdapterImpl {
       if (this.isVerboseSession(event.sessionId)) {
         this.notifyLifecycle(event.sessionId, event.projectId, "finished");
       }
+      const finalAssistantText = this.finalAssistantTextBySession.get(event.sessionId);
+      this.finalAssistantTextBySession.delete(event.sessionId);
+      if (finalAssistantText) {
```
markustherobot: **[P2] Gate TTS on the logical turn outcome**

This text can come from an earlier `stop`/`length` subturn during steering or an active-goal continuation. If a later continuation fails, blocks, or is interrupted, the manager still transitions from `running` to `waiting-input`, so the adapter can synthesize this stale response after reporting the unsuccessful turn. Associate the candidate with a successful logical-turn settlement, or clear it on every non-completed outcome before starting TTS.

src/core/telegram/tts.ts
```diff
@@ -0,0 +1,164 @@
...
+  const spawn = options.deps?.spawn ?? spawnWithCapture;
+  const chunks: Buffer[] = [];
+
+  for await (const chunk of streamSpeechAudio({
+    apiKey: options.apiKey,
+    sourceText: options.sourceText,
+    fetchImpl: options.fetchImpl,
+    signal: options.signal,
+  })) {
+    chunks.push(chunk.audio);
```
markustherobot: **[P2] Bound PCM before collecting every chunk**

A long but valid response retains every WAV chunk here, then `concatenateWaveAudio` allocates another complete PCM buffer; the 50 MB check happens only after writing, encoding, and reading the result. Large responses can consume hundreds of megabytes and kill the runner before the eventual upload-size rejection. Track a cumulative audio budget while consuming chunks, or stream PCM into `ffmpeg` instead of retaining and duplicating the whole response.

src/core/telegram/tts.ts
```diff
@@ -0,0 +1,164 @@
...
+  }
+
+  const waveAudio = concatenateWaveAudio(chunks);
+  const temporaryDirectory = await mkdtemp(join(tmpdir(), TELEGRAM_TTS_TEMP_DIR_PREFIX));
```
markustherobot, line 46: **[P2] Sweep TTS directories left by crashes**

The `finally` cleanup only runs during an orderly process lifetime. A crash or kill after this call leaves `tau-telegram-tts-*` WAV and Ogg files, including assistant audio, in the system temp directory indefinitely. Add a startup sweep for this prefix, matching the existing stale attachment-directory cleanup.

markustherobot (commented):
I found several issues in the asynchronous voice lifecycle: jobs can be delivered out of order, unsuccessful logical runs can reuse stale speech text, and conversion lacks bounded buffering and crash cleanup. I launched 4 subreviews covering event and delivery flow, audio generation, preference migration, and integration regressions; all completed successfully.

markusylisiurunen:
@markustherobot review

src/core/telegram/session_manager.ts
```diff
@@ -1376,6 +1392,7 @@ class TelegramSessionManagerImpl implements TelegramSessionManager {
       .finally(() => {
         if (entry.activeSubmit === trackedSubmit) {
           entry.activeSubmit = undefined;
+          this.emitCompletedResponseIfReady(entry);
```
markustherobot, line 1396: **[P1] Block completion after rejected steering**

A `tauSession.steer()` transport or pre-acceptance rejection has no `requestedHistoryEntryId`, so the catch path marks the session failed without setting `turnSequenceUnsuccessful` or clearing `finalAssistantResponse`. If the original turn already committed a `stop` response, this finalizer still emits `session-response-completed` after the failed steering sequence and the adapter speaks that stale answer. Invalidate the candidate on every unrecovered submit/steer failure (and cancellation), or make the completion gate require a successful, non-cancelled session state.

src/core/telegram/adapter.ts
```diff
@@ -3020,6 +3100,130 @@ class TelegramAdapterImpl {
     );
   }
 
+  private startTtsNotification(sessionId: string, sourceText: string): void {
+    const chatIds = this.chatsBySession.get(sessionId);
+    if (
+      !this.geminiApiKey ||
```
markustherobot: **[P2] Report missing credentials for persisted opt-ins**

If a chat enabled TTS previously and the runner restarts without the formerly available Google key, its persisted preference remains enabled but this branch silently drops every voice job before logging or sending the documented failure notice. Treat the missing key as a voice-job failure for opted-in chats (or explicitly disable and report the preference) so credential removal does not turn the feature off invisibly.

src/core/telegram/tts.ts
```diff
@@ -0,0 +1,185 @@
...
+  try {
+    const entries = await readdir(systemTmpDir, { withFileTypes: true, encoding: "utf8" });
+    await Promise.allSettled(
+      entries
+        .filter(
+          (entry) => entry.isDirectory() && entry.name.startsWith(TELEGRAM_TTS_TEMP_DIR_PREFIX),
+        )
+        .map(
+          async (entry) =>
+            await rm(join(systemTmpDir, entry.name), { recursive: true, force: true }),
```
markustherobot, line 22: **[P2] Avoid sweeping another runner's live TTS directory**

This startup sweep removes every matching directory in the shared system temp directory. During an overlapping restart, or when two Tau Telegram runners use the same OS account, one process can delete WAV inputs or the Ogg output of another process's active five-minute job and make that response fail. Only remove directories demonstrably older than the maximum job lifetime, or use an ownership/lock marker before deleting them.

markustherobot (commented):
I found remaining lifecycle and operational gaps around unsuccessful-sequence gating, persisted credential state, and temporary workspace ownership. I launched 4 subreviews covering response settlement, Telegram delivery, speech/audio processing, and preference migration; all completed successfully.

src/core/telegram/session_manager.ts
markusylisiurunen: Fixed in d9c9c06. Unrecovered submit or steering failures now invalidate the retained response candidate, as do cancellation and every non-completed or rejected turn outcome. The completion gate also explicitly rejects cancelled and failed session state. Added regressions for steering rejected before acceptance and cancellation after a candidate was committed. The broader host-owned logical-response completion contract is tracked in #435.

src/core/telegram/adapter.ts
markusylisiurunen: Fixed in d9c9c06. A persisted opt-in no longer exits early when the Google credential is absent. The serialized voice job logs the missing credential and sends the standard failure notification while preserving the preference, so restoring the runner credential resumes voice generation. Added a regression confirming generation is not invoked without a key.

src/core/telegram/tts.ts
markusylisiurunen: No code change. Tau supports at most one Telegram runner process per machine; multiple bots are adapters within that single runtime, and the TTS sweep runs once before adapters start. Under that operating invariant, an orderly restart has no overlapping owner and crash leftovers have no live owner, so immediate cleanup is intentional.

PR #436, opened 2026-08-13
isolate history replication failures

markusylisiurunen:
## why

Remote history replication currently uses one endpoint-wide FIFO outbox. A permanent rejection for one session is retried from the head forever, preventing every later session from reaching the shared history service, while the replication error itself is silently discarded.

## what

Replicate ordered batches through independent per-session lanes. Permanent operation-domain rejections quarantine only the affected session and retain its pending operations plus a durable diagnostic, while unrelated lanes continue. Transient transport and endpoint failures remain pending for a later retry.

Preserve remote HTTP status and error codes, emit structured host diagnostics for all replication failures, and migrate existing outboxes by backfilling their session lane from the durable operation payload.

### DIFF
--- a/src/core/history/history_manager.ts
+++ b/src/core/history/history_manager.ts
@@ -1,5 +1,5 @@
 import { LocalHistoryStore } from "./local_history_store.js";
-import { RemoteHistoryClient } from "./remote_history_client.js";
+import { RemoteHistoryClient, RemoteHistoryError } from "./remote_history_client.js";
 import type {
   HistoryEntry,
   HistoryQuery,
@@ -10,8 +10,15 @@ import type {
 } from "./types.js";
 
 const REPLICATION_BATCH_SIZE = 10;
+const REPLICATION_LANE_BATCH_SIZE = 10;
 const REPLICATION_BATCH_BYTES = 6 * 1024 * 1024;
 const REPLICATION_TIMEOUT_MS = 15_000;
+const PERMANENT_REPLICATION_ERROR_CODES = new Set([
+  "immutable_conflict",
+  "invalid_request",
+  "not_found",
+  "request_too_large",
+]);
 
 export class HistoryManager {
   private readonly replicationRuns = new Map<string, Promise<void>>();
@@ -151,7 +158,9 @@ export class HistoryManager {
       }
     });
     this.replicationRuns.set(target.endpoint, run);
-    void run.catch(() => undefined);
+    void run.catch((error) => {
+      reportReplicationFailure({ endpoint: target.endpoint, error });
+    });
   }
 
   private async replicate(endpoint: string): Promise<void> {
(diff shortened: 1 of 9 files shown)


src/core/history/local_history_store.ts
```diff
@@ -364,10 +445,42 @@ export class LocalHistoryStore {
...
+      );
     `);
+    const outboxColumns = new Set(
+      (this.database.prepare("PRAGMA table_info(history_outbox)").all() as SqlRow[]).map((row) =>
+        String(row.name),
+      ),
+    );
+    if (!outboxColumns.has("session_id")) {
+      this.transaction(() => {
+        this.database.exec("ALTER TABLE history_outbox ADD COLUMN session_id TEXT");
```
markustherobot, line 475: **[P2] Recheck the outbox schema under the migration lock**

Every Tau host opens the same machine-local `history.sqlite`. If two hosts first open a pre-PR database concurrently, both can observe that `session_id` is absent; after one commits this `ALTER`, the other acquires `BEGIN IMMEDIATE` and reruns it, gets a duplicate-column error, and disables history for that host process. Recheck `PRAGMA table_info` after acquiring the write lock (or otherwise make the migration idempotent) before altering the table.

src/core/history/history_manager.ts
```diff
@@ -179,6 +224,43 @@ function formatFailureReason(error: unknown): string {
...
+  sessionId?: string;
+  operationId?: string;
+  error: unknown;
+  quarantined?: boolean;
+}): void {
+  const error = input.error;
+  console.error(
+    JSON.stringify({
+      event: "history_replication_failed",
+      endpoint: input.endpoint,
```
markustherobot, line 250: **[P1] Keep endpoint credentials out of failure logs**

History config accepts HTTP(S) URLs containing userinfo, but this new diagnostic writes the endpoint verbatim. Node's `fetch` also rejects such URLs with an error message containing the full credential-bearing URL, which is logged again below, so a normal replication attempt with `https://user:secret@host` exposes the secret in host logs. Reject URL userinfo at the config boundary (preferred because this fetch path cannot use it), or sanitize both the endpoint and error before logging.

src/core/history/remote_history_client.ts
```diff
@@ -129,17 +140,23 @@ export class RemoteHistoryClient implements HistoryQuery {
...
+        value.error !== null
+          ? value.error
+          : undefined;
+      const code =
+        remoteError && "code" in remoteError && typeof remoteError.code === "string"
+          ? remoteError.code
+          : undefined;
+      const message =
+        remoteError && "message" in remoteError && typeof remoteError.message === "string"
+          ? remoteError.message
```
markustherobot: **[P2] Bound structured error fields before recording them**

These strings can consume nearly the full 16 MiB response allowance. The manager now logs `code` and `message` on every retry and persists the full message once per quarantined session, so a malformed or failing endpoint can amplify oversized errors into repeated log growth and a large local database. Bound both structured fields at this remote-response boundary before constructing `RemoteHistoryError`.

markustherobot (commented):
I found material issues in the concurrent upgrade path and in how replication diagnostics handle externally supplied data. I launched 3 subreviews covering scheduler semantics, SQLite migration/persistence, and remote error-contract/test alignment; all completed successfully.

src/core/history/history_manager.ts
markusylisiurunen: isn't the config just a regular domain and the api key is separately? I think that should be the right way and should not sanitize it here, that feels overly defensive.

markusylisiurunen:
@markustherobot address the findings

src/core/history/local_history_store.ts
markustherobot: Rechecked `PRAGMA table_info(history_outbox)` after acquiring `BEGIN IMMEDIATE`, so a concurrent opener that already completed the migration causes this opener to skip the `ALTER`.

src/core/history/history_manager.ts
markustherobot: Agreed that the API key should remain separate. I enforced that contract at the config boundary: `history.endpoint` now rejects URL userinfo, and the config error does not echo the credential-bearing value.

src/core/history/remote_history_client.ts
markustherobot: Bound structured remote error codes to 128 UTF-8 bytes and messages to 4 KiB before constructing `RemoteHistoryError`, with a multibyte regression test.

PR #437, opened 2026-08-13
set noninteractive Bash environment

markusylisiurunen:
## why

Bash tools run without a TTY, but inherited terminal, color, pager, and login-startup settings can still produce interactive behavior or unstable command output across execution environments.

## what

Set `NO_COLOR=1`, `FORCE_COLOR=0`, `TERM=dumb`, and `PAGER=cat` as fixed Bash environment overrides alongside Tau's existing noninteractive Git values. Apply the same authoritative baseline to local, Cloudflare Sandbox, and Fly Sprite execution, and reassert it after login startup files run. Commands can still assign their own environment explicitly.

Backend tests verify the values override conflicting execution-environment and login-profile values while preserving explicit command assignments.


src/core/tools/execution_backend.ts
```diff
@@ -65,7 +65,10 @@ export type ListDirResult = {
   entries: ListDirEntry[];
 };
 
-export const NONINTERACTIVE_GIT_ENV = {
+export const BASH_ENVIRONMENT_OVERRIDES = {
+  NO_COLOR: "1",
+  TERM: "dumb",
+  PAGER: "cat",
```
markustherobot, line 72: **[P1] Reapply overrides after login startup**

These values are only installed in the environment used to start `bash -lc`, so Bash subsequently sources the execution environment's login files and can replace them before the requested command runs. A routine `.bash_profile` containing `export PAGER=less` (or assignments for `TERM`/`NO_COLOR`) restores those values, contradicting the fixed contract and potentially re-enabling an interactive pager until timeout. Reassert the overrides after startup initialization, before executing the requested command, and cover a conflicting profile in a regression test.

src/core/tools/execution_backend.ts
```diff
@@ -65,7 +65,10 @@ export type ListDirResult = {
   entries: ListDirEntry[];
 };
 
-export const NONINTERACTIVE_GIT_ENV = {
+export const BASH_ENVIRONMENT_OVERRIDES = {
+  NO_COLOR: "1",
```
markustherobot, line 69: **[P2] Neutralize force-color precedence**

`NO_COLOR=1` does not guarantee colorless output when the inherited environment contains `FORCE_COLOR`, which Node and Chalk deliberately give precedence. Because every unlisted environment key is retained here, `FORCE_COLOR=1 node -e 'console.log({a:1})'` still emits ANSI escapes and Node also writes a warning that `NO_COLOR` was ignored, defeating the predictable-output goal. Add an authoritative `FORCE_COLOR=0` override and test it against a conflicting input rather than relying on `NO_COLOR` alone.

markustherobot (commented):
I found two environment-precedence gaps that keep the new baseline from being authoritative in realistic configurations. I launched two subreviews covering backend/startup semantics and the tests/public behavior contract; both completed successfully.

markusylisiurunen:
@markustherobot address the findings but avoid adding too much complexity if e.g. the FORCE_COLORS stuff is explicitly trying to be overridden like in the FORCE_COLOR=1 example.

src/core/tools/execution_backend.ts
markustherobot: Fixed in 5d31090. Bash commands now prepend exports derived from the shared override map, so the values are reapplied after login startup. The local regression test uses a conflicting `.bash_profile`.

src/core/tools/execution_backend.ts
markustherobot: Fixed in 5d31090. The baseline now sets `FORCE_COLOR=0` and tests conflicting execution and profile values. An explicit assignment inside the command, such as `FORCE_COLOR=1 node ...`, remains allowed and is covered.

markusylisiurunen:
@markustherobot can you now explain with simple examples how these behave after this pr? no jargon, narrative examples.

markusylisiurunen:
@markustherobot how does Tau restore after the .bash_profile file? show me with pseudo-code samples what happens when and how

PR #439, opened 2026-08-14
retain failed speech and stream OpenAI transcription

### LINKED ISSUE #438: when speech transcription fails, the full audio is lost. that's annoying as fuck.

markustherobot:
## why

Failed speech transcription currently deletes the completed WAV, so a transient provider outage or malformed response loses the original voice input. Existing providers also wait until recording finishes before uploading audio, leaving avoidable latency for interactive dictation.

## what

- Retain one failed local recording and report its exact path.
- Add `/listen retry` to transcribe retained audio with the current provider and fresh conversation context, plus `/listen discard` for explicit deletion.
- Keep retained audio until a replacement capture is producing audio, so replacement startup failures do not destroy the recoverable recording.
- Retry transient Gemini failures once on Gemini 3.7 Flash, then use Gemini 3.6 Flash as a final fallback. Permanent request failures stop immediately.
- Add `speechToText.provider: "openai"` using `OPENAI_API_KEY` or `apiKeys.openai`, and make OpenAI the default provider when the setting is absent.
- Stream `/listen` microphone audio to `gpt-live-transcribe` while recording, then insert only the final transcript after recording stops.
- Support OpenAI for Telegram by decoding downloaded audio to PCM with runner-side `ffmpeg`, with owner cancellation, a five-minute decoded-audio limit, and WebSocket backpressure.
- Cover provider configuration, credentials, Realtime protocol events, live capture, retained-audio replay, lifecycle cancellation, resource bounds, and Telegram transcription.
- Document the provider defaults, credential ownership, ffmpeg requirement, and privacy-sensitive retention lifecycle.

## details

OpenAI is the default provider. Its transcription path opens the official transcription-specific Realtime WebSocket, configures a `gpt-live-transcribe` session with 24 kHz mono PCM and conversation context, streams audio with `input_audio_buffer.append`, and manually commits the turn when capture ends. Partial deltas remain internal; Tau waits for `conversation.item.input_audio_transcription.completed` before changing the editor. Mistral and Gemini remain selectable through `speechToText.provider`.

The TUI writes a WAV while streaming PCM so the recovery contract remains available when the connection or transcription fails. `/listen retry` decodes that retained WAV and replays it through a fresh provider request. The retained path is client-local controller state, not session state, and exiting leaves the file at the reported path for manual recovery. Starting a replacement preserves the old WAV through temp-file creation, provider setup, and capture startup, then removes it only after the new capture emits PCM.

OpenAI batch replay waits for the Realtime session before starting `ffmpeg`. Decoder cancellation follows socket failure, transcription abort, TUI disposal, and Telegram shutdown. Decoded PCM is limited to five minutes, and ffmpeg stdout pauses while the WebSocket send buffer exceeds the configured high-water mark.

The Gemini fallback remains inside the configured Google provider and also applies to Telegram through the shared speech-to-text owner. Retryable failures include transport errors, HTTP 408/429/5xx responses, provider availability statuses, and invalid model output. Retry delays use short backoff and honor `Retry-After` up to five seconds.

fixes #438

### DIFF
--- a/src/core/utils/openai_transcription.ts
+++ b/src/core/utils/openai_transcription.ts
@@ -14,6 +14,10 @@ const OPENAI_TRANSCRIPTION_COMPLETION_TIMEOUT_MS = 30_000;
 const OPENAI_TRANSCRIPTION_CONTEXT_TOKENS = 900;
 const OPENAI_TRANSCRIPTION_FFMPEG_TIMEOUT_MS = 5 * 60 * 1_000;
 const OPENAI_TRANSCRIPTION_FFMPEG_OUTPUT_LIMIT_BYTES = 20_000;
+const OPENAI_TRANSCRIPTION_MAX_PCM_BYTES =
+  OPENAI_TRANSCRIPTION_SAMPLE_RATE * 2 * (OPENAI_TRANSCRIPTION_FFMPEG_TIMEOUT_MS / 1_000);
+const OPENAI_TRANSCRIPTION_BACKPRESSURE_HIGH_BYTES = 1024 * 1024;
+const OPENAI_TRANSCRIPTION_BACKPRESSURE_LOW_BYTES = 256 * 1024;
 
 const eventSchema = z.object({ type: z.string() }).passthrough();
 const errorEventSchema = z.object({
@@ -30,6 +34,7 @@ const transcriptionCompletedEventSchema = z.object({
 });
 
 export type OpenAITranscriptionWebSocket = {
+  readonly bufferedAmount?: number;
   on(event: "open", listener: () => void): unknown;
   on(event: "message", listener: (data: unknown) => void): unknown;
   on(event: "error", listener: (error: Error) => void): unknown;
@@ -46,7 +51,11 @@ export type OpenAITranscriptionWebSocketFactory = (
 
 export type OpenAIStreamingTranscription = {
   appendAudio(audio: Buffer): void;
-  finish(options?: { audio?: Buffer; spawnImpl?: typeof spawnWithCapture }): Promise<string>;
+  finish(options?: {
+    audio?: Buffer;
+    signal?: AbortSignal;
+    spawnImpl?: typeof spawnWithCapture;
+  }): Promise<string>;
   abort(): void;
 };
 
@@ -58,17 +67,21 @@ export type StartOpenAITranscriptionOptions = {
 
 export type TranscribeOpenAIAudioOptions = StartOpenAITranscriptionOptions & {
   audio: Buffer;
+  signal?: AbortSignal;
   spawnImpl?: typeof spawnWithCapture;
 };
 
 class OpenAIStreamingTranscriptionImpl implements OpenAIStreamingTranscription {
   private readonly socket: OpenAITranscriptionWebSocket;
+  private readonly abortController = new AbortController();
+  private readonly audioDrainWaiters = new Set<() => void>();
   private ready = false;
   private aborted = false;
   private failure?: Error;
   private completedTranscript?: string;
   private hasAudio = false;
   private pendingAudio: Buffer[] = [];
+  private pendingAudioBytes = 0;
   private readyTimeout?: ReturnType<typeof setTimeout>;
   private completionTimeout?: ReturnType<typeof setTimeout>;
   private resolveReady?: () => void;
@@ -142,6 +155,7 @@ class OpenAIStreamingTranscriptionImpl implements OpenAIStreamingTranscription {
     this.hasAudio = true;
     if (!this.ready) {
       this.pendingAudio.push(audio);
+      this.pendingAudioBytes += audio.length;
       return;
     }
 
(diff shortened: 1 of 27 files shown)


markustherobot (commented):
Reviewed the full speech-retention and fallback change. Three Luna subreviews covered the provider/shared transcription contract, retained-file lifecycle, and command/Telegram integration in addition to the complete maintainer pass. I found one lifecycle correctness issue around safe replacement; the provider retry and integration paths otherwise align, and the focused test suite passes.

src/tui/session_chat_controller.ts
```diff
@@ -791,6 +807,20 @@ export class SessionChatController {
       return;
     }
 
+    const retainedAudioPath = this.retainedListenAudioPath;
+    if (retainedAudioPath) {
+      try {
+        await deleteListenTempFile(retainedAudioPath);
```
markustherobot: **[P2] Keep the retained file until replacement capture starts**

When `/listen` is used after a failed transcription, this unlinks the only recoverable WAV before `mktemp` and ffmpeg have established the replacement. If temp-file creation fails or ffmpeg exits immediately (for example, because the audio device is unavailable), the new path is cleaned up and the previous audio is permanently gone even though no replacement was recorded. Keep the old path until the new capture has started successfully, and preserve it if replacement startup fails.

markusylisiurunen:
@markustherobot review

markustherobot (commented):
I reviewed the complete current PR with four Luna subreviews covering the OpenAI Realtime/ffmpeg path, TUI retained-audio lifecycle, shared provider retries, and Telegram/config integration. I found three material lifecycle and resource issues. The remaining provider/config integration aligns with the intended behavior, and the focused suite passed (191 tests across seven files).

src/tui/session_chat_controller.ts
```diff
@@ -791,21 +811,39 @@ export class SessionChatController {
...
```
markustherobot: **[P2] Keep retained audio until replacement capture is established**

When `/listen` is used after a failed transcription, this deletes the only recoverable WAV before `mktemp`, transcription setup, or ffmpeg has established a replacement. If startup then fails because the audio device is unavailable, ffmpeg is missing, or temp-file creation fails, the previous recording is permanently lost. Preserve the old path through startup and delete it only once the new capture is successfully running.

src/core/utils/openai_transcription.ts
```diff
@@ -0,0 +1,417 @@
...
+    }
+
+    this.sendAudio(audio);
+  }
+
+  async finish(
+    options: { audio?: Buffer; spawnImpl?: typeof spawnWithCapture } = {},
+  ): Promise<string> {
+    if (!this.hasAudio && options.audio) {
+      await decodeAndStreamOpenAIAudio({
```
markustherobot: **[P2] Cancel decoding with the transcription lifecycle**

`finish()` waits for ffmpeg before checking `readyPromise`, but neither the decoder nor the socket wait receives an owner cancellation signal. If the Realtime socket fails during decode, or Telegram/TUI shuts down, ffmpeg keeps running while the owner waits for this task, potentially until the five-minute decode timeout and then the completion timeout. Thread an `AbortSignal` through `finish`, pass it to `spawnWithCapture`, and trigger it from `fail()`, `abort()`, and owner shutdown.

src/core/utils/openai_transcription.ts
```diff
@@ -0,0 +1,417 @@
...
+        "-c:a",
+        "pcm_s16le",
+        "-f",
+        "s16le",
+        "pipe:1",
+      ],
+      {
+        input: args.audio,
+        timeoutMs: OPENAI_TRANSCRIPTION_FFMPEG_TIMEOUT_MS,
+        maxCaptureBytes: OPENAI_TRANSCRIPTION_FFMPEG_OUTPUT_LIMIT_BYTES,
```
markustherobot, line 415: **[P2] Bound or backpressure decoded PCM**

This limit applies only to captured stderr because `captureOutput` is `"stderr"`; decoded stdout remains unbounded. A compressed Telegram audio file can expand to hundreds of megabytes of PCM, and ffmpeg drains it as fast as possible while `ws.send()` queues base64 strings without backpressure, which can exhaust the runner's memory. Enforce a decoded-byte or duration limit and pause stdout when the socket backlog grows.

markusylisiurunen:
@markustherobot address the findings, and can we also change the default speech provider from mistral to openai?

src/tui/session_chat_controller.ts
markustherobot: I changed replacement startup so the retained WAV is deleted only after the new ffmpeg capture emits PCM. Temp-file, provider, and capture startup failures preserve the old recording, with regression coverage for an immediate audio-device failure.

src/tui/session_chat_controller.ts
markustherobot: I kept the retained WAV authoritative through temp-file creation, transcription setup, and capture startup. It is removed only after the replacement emits audio, and startup failure now leaves the original recording intact.

src/core/utils/openai_transcription.ts
markustherobot: I now wait for Realtime readiness before decoding and pass a lifecycle AbortSignal into ffmpeg. Socket failure, transcription abort, TUI disposal, and Telegram shutdown all cancel the decoder, with focused cancellation tests.

src/core/utils/openai_transcription.ts
markustherobot: I bounded decoded PCM to five minutes and added high/low-water WebSocket backpressure. ffmpeg stdout pauses while buffered sends exceed the high-water mark and resumes after the backlog drains; both behaviors have regression coverage.

PR #441, opened 2026-08-14
preserve metadata while pursuing goals

### LINKED ISSUE #440: pursuing goal should not hide the footer metadata

markusylisiurunen:
## why

Goal pursuit replaced the footer's normal duration, working directory, context usage, and cost metadata with a long activity label.

## what

Render a compact accent-colored `goal` badge before the normal footer metadata. Keep the separator and metadata dim, and account for the badge when compacting the status line to fit the terminal width.

Update the footer component test to cover the badge styling and preserved metadata.

fixes #440

### DIFF
--- a/src/tui/ui/footer.ts
+++ b/src/tui/ui/footer.ts
@@ -174,14 +174,14 @@ export class FooterComponent implements Component {
     const activityStyle = this.working || isCompleting ? palette.brandAccent : palette.textDim;
     const icon = activityStyle(iconChar);
     const iconWidth = visibleWidth(iconChar);
-    const availableWidth = Math.max(0, width - iconWidth - 3);
-    const goalStatus = this.working && this.status?.pursuingGoal ? "pursuing goal" : "";
-    const rawText = goalStatus || this.buildStatusLine(availableWidth);
-    const style = goalStatus ? activityStyle : palette.textDim;
-    const text = truncateFromEndByWidth(rawText, availableWidth);
+    const goalStatus = this.working && this.status?.pursuingGoal ? "goal" : "";
+    const goalPrefix = goalStatus ? `${activityStyle(goalStatus)} ${palette.textDim("·")} ` : "";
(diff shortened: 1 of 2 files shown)


src/tui/ui/footer.ts
```diff
@@ -174,14 +174,14 @@ export class FooterComponent implements Component {
...
-    const text = truncateFromEndByWidth(rawText, availableWidth);
+    const goalStatus = this.working && this.status?.pursuingGoal ? "goal" : "";
+    const goalPrefix = goalStatus ? `${activityStyle(goalStatus)} ${palette.textDim("·")} ` : "";
+    const goalPrefixWidth = goalStatus ? visibleWidth(goalStatus) + 3 : 0;
+    const availableWidth = Math.max(0, width - iconWidth - goalPrefixWidth - 3);
+    const text = truncateFromEndByWidth(this.buildStatusLine(availableWidth), availableWidth);
     const padding = " ".repeat(Math.max(0, availableWidth - visibleWidth(text)));
 
-    return [` ${icon} ${style(text)}${padding} `];
+    return [` ${icon} ${goalPrefix}${palette.textDim(text)}${padding} `];
```
markustherobot, line 184: **[P2] Keep the goal prefix within the render width**

When active goal work is rendered at widths 3-10, such as while a terminal pane is being resized, `availableWidth` clamps to zero but this still emits the complete prefix, so `render(10)` returns an 11-column line. `TuiMainScreen` treats a non-image line wider than the terminal as a component contract violation and can stop with an error on the next animation render. Omit or truncate the badge and separator when the fixed prefix does not fit, and add a narrow-width regression asserting the rendered line never exceeds `width`.

markustherobot (commented):
I found one actionable footer rendering issue in the constrained-width path. I launched two subreviews, covering width/rendering behavior and state/test integration; both completed successfully. Local build, the focused UI component suite, and the full 1,172-test suite pass.

PR #443, opened 2026-08-15
stream Gemini speech playback

### LINKED ISSUE #442: stream Gemini speech playback in the TUI
## why

Shared Gemini speech generation currently divides rewritten responses into fixed-size segments: 500 characters for progressive TUI delivery and 1,000 characters for complete Telegram delivery. This bounds failures and output size, but the segment durations are uneven and the separate policies can produce unnecessarily short requests.

The TUI additionally waits for each complete segment, wraps it in a temporary WAV file, and starts `afplay` separately for playback. `gemini-3.1-flash-tts-preview` supports incremental raw PCM output through `streamGenerateContent`, so the TUI can begin playback sooner and keep one continuous audio sink. Telegram must still upload a finished OGG file, but it should use the same balanced segmentation policy.

## what

- Replace fixed character chunking in shared Gemini speech generation with one balanced duration-based partitioning policy used by every consumer, including TUI `/speak` and Telegram voice generation:
  - Treat two minutes of raw Gemini-generated audio as the maximum segment duration, before applying Tau's 1.15x playback rate or any output encoding.
  - Estimate the complete rewritten transcript's duration, calculate `ceil(total duration / 120 seconds)` segments, and divide it into approximately equal speech-weight sections at natural sentence or paragraph boundaries.
  - For example, an estimated 150-second transcript should become two roughly 75-second segments, not one 120-second segment followed by 30 seconds.
- Add a Gemini TTS streaming path specifically for TUI playback using `streamGenerateContent`.
- Feed TUI 24 kHz mono 16-bit PCM into one long-lived, gapless local playback process rather than creating a WAV file and launching `afplay` for every segment.
- Decouple TUI generation from playback with a bounded ordered PCM buffer. Begin generating at least the next transcript segment while the current segment is playing, retain one-segment read-ahead, and feed buffered audio continuously into the same playback process so request startup does not create audible gaps. Apply backpressure rather than allowing unbounded pre-generation.
- Keep Telegram completion-oriented: generate the shared balanced segments completely, concatenate them, encode the result, and upload the finished OGG.
- Retry a TUI streaming request automatically only when it fails before emitting audible audio. If playback has begun, stop and report the failure rather than replaying already-heard speech.
- Preserve cancellation, terminal finish-reason validation, and the cumulative 32 MiB raw-audio limit in both delivery paths.

Add focused shared coverage for balanced partitioning and consumer coverage for TUI incremental playback, Telegram balanced complete generation, stream failure before and after audio emission, cancellation, output limits, and continuous playback across transcript segments. Update the `/speak` documentation if its local program requirements change.

markusylisiurunen:
## why

Gemini 3.1 TTS can emit audio incrementally, but `/speak` waited for complete fixed-size chunks and restarted playback for each one. The TUI therefore paid avoidable startup latency and could introduce seams, while TUI and Telegram used different, uneven segmentation policies.

## what

Use one shared balanced speech segmentation policy for both TUI and Telegram. Segments estimate at most two minutes of unmodified Gemini speech, distribute the full transcript evenly, and prefer nearby sentence or paragraph boundaries.

Stream TUI speech as 24 kHz PCM through one long-lived `ffplay` process. Generate one future segment while the current segment plays, preserve ordering with bounded read-ahead, validate stream completion, and retry only before any audio from the active segment is emitted.

Keep Telegram completion-oriented for Ogg Opus encoding while using the same shared segments. Update speech documentation and cover balancing, SSE parsing, prefetch, cancellation, limits, retries, continuous playback, and Telegram encoding behavior.

fixes #442

### DIFF
--- a/src/core/utils/gemini_speech.ts
+++ b/src/core/utils/gemini_speech.ts
@@ -52,6 +52,10 @@ export type GeminiSpeechOptions = {
   onSegmentProgress?: (progress: GeminiSpeechSegmentProgress) => void | Promise<void>;
 };
 
+export type GeminiSpeechPcmOptions = GeminiSpeechOptions & {
+  initialBufferBytes?: number;
+};
+
 export type GeminiSpeechAudioChunk = {
   index: number;
   total: number;
@@ -137,7 +141,7 @@ export async function* generateGeminiSpeechAudio(
 }
 
 export async function* streamGeminiSpeechPcm(
-  options: GeminiSpeechOptions,
+  options: GeminiSpeechPcmOptions,
 ): AsyncGenerator<GeminiSpeechPcmChunk> {
   const abortController = createLinkedAbortController(options.signal);
   let completed = false;
@@ -168,6 +172,7 @@ export async function* streamGeminiSpeechPcm(
       ...prepared,
       spokenText: prepared.spokenSegments[0]!,
       signal: abortController.signal,
+      initialBufferBytes: Math.max(0, Math.trunc(options.initialBufferBytes ?? 0)),
     });
     const firstIterator = firstStream[Symbol.asyncIterator]();
     let nextAudio = firstIterator.next();
@@ -463,17 +468,34 @@ async function synthesizeSpeechAudioSegment(
 type StreamSpeechSegmentArgs = SynthesizeSpeechAudioSegmentArgs;
 
 async function* streamSpeechSegmentWithRetries(
-  args: StreamSpeechSegmentArgs,
+  args: StreamSpeechSegmentArgs & { initialBufferBytes: number },
 ): AsyncGenerator<Buffer> {
   const maxAttempts = Math.max(1, Math.trunc(args.maxAttempts));
   let lastError: unknown;
 
   for (let attempt = 1; attempt <= maxAttempts; attempt += 1) {
+    const bufferedAudio: Buffer[] = [];
+    let bufferedAudioBytes = 0;
     let emittedAudio = false;
     try {
       for await (const audio of requestGeminiSpeechStream(args)) {
+        if (!emittedAudio && bufferedAudioBytes < args.initialBufferBytes) {
+          bufferedAudio.push(audio);
+          bufferedAudioBytes += audio.length;
+          if (bufferedAudioBytes < args.initialBufferBytes) {
+            continue;
+          }
+          emittedAudio = true;
+          yield Buffer.concat(bufferedAudio);
+          continue;
+        }
+
         emittedAudio = true;
         yield audio;
+      }
+      if (bufferedAudio.length > 0 && !emittedAudio) {
+        emittedAudio = true;
+        yield Buffer.concat(bufferedAudio);
       }
       return;
     } catch (error) {
--- a/src/core/utils/gemini_speech.ts
+++ b/src/core/utils/gemini_speech.ts
@@ -6,15 +6,16 @@ const DEFAULT_GEMINI_SPEECH_REWRITE_MODEL = "gemini-3.7-flash";
 const DEFAULT_GEMINI_SPEECH_REWRITE_THINKING_LEVEL = "low";
 const DEFAULT_GEMINI_SPEECH_TTS_MODEL = "gemini-3.1-flash-tts-preview";
 const DEFAULT_GEMINI_TTS_VOICE_NAME = "Despina";
-const DEFAULT_TTS_SAMPLE_RATE_HZ = 24000;
-const DEFAULT_TTS_CHANNEL_COUNT = 1;
-const DEFAULT_TTS_BITS_PER_SAMPLE = 16;
+export const GEMINI_SPEECH_SAMPLE_RATE_HZ = 24000;
+export const GEMINI_SPEECH_CHANNEL_COUNT = 1;
+export const GEMINI_SPEECH_BITS_PER_SAMPLE = 16;
 const DEFAULT_TTS_MAX_ATTEMPTS = 3;
 const SPEECH_REWRITE_TIMEOUT_MS = 60_000;
-const PROGRESSIVE_TTS_CONCURRENCY = 6;
 const COMPLETE_TTS_CONCURRENCY = 3;
-const PROGRESSIVE_SPEECH_CHUNK_CHARACTERS = 500;
-const COMPLETE_SPEECH_CHUNK_CHARACTERS = 1000;
+const MAX_SPEECH_SEGMENT_SECONDS = 120;
+const ESTIMATED_SPEECH_CHARACTERS_PER_SECOND = 17;
+const MAX_SPEECH_SEGMENT_WEIGHT =
+  MAX_SPEECH_SEGMENT_SECONDS * ESTIMATED_SPEECH_CHARACTERS_PER_SECOND;
 const MAX_SPEECH_SOURCE_CHARACTERS = 10_000;
 const MAX_SPOKEN_TEXT_CHARACTERS = 10_000;
 const MAX_SPEECH_PCM_BYTES = 32 * 1024 * 1024;
@@ -32,9 +33,8 @@ const errorPayloadSchema = z.object({
 });
 
 export type GeminiSpeechStage = "rewriting" | "generating";
-export type GeminiSpeechDeliveryMode = "progressive" | "complete";
 
-export type GeminiSpeechChunkProgress = {
+export type GeminiSpeechSegmentProgress = {
   ready: number;
   total: number;
 };
@@ -42,7 +42,6 @@ export type GeminiSpeechChunkProgress = {
 export type GeminiSpeechOptions = {
   apiKey: string;
   sourceText: string;
-  deliveryMode: GeminiSpeechDeliveryMode;
   rewriteModel?: string;
   ttsModel?: string;
   voiceName?: string;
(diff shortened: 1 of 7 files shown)


src/core/utils/gemini_speech.ts
```diff
@@ -559,55 +833,113 @@ function exceedsUnicodeCharacterLimit(text: string, limit: number): boolean {
...
-    if (characters[index] === "\n" && characters[index + 1] === "\n") {
-      return index;
-    }
+function selectSpeechBoundary(
+  naturalCandidates: number[],
+  fallbackCandidates: number[],
+  cumulativeWeights: number[],
+  idealWeight: number,
+): number {
+  const fallback = closestSpeechBoundary(fallbackCandidates, cumulativeWeights, idealWeight);
```
markustherobot, line 948: **[P1] Handle weighted intervals with no character boundary**

For valid mixed-weight text, the admissible interval can contain no cumulative character boundary. For example, `"これは音声です。".repeat(1020)` makes the first interval exactly weight 2040 but no prefix has that weight, so `fallbackCandidates` is empty and `closestSpeechBoundary` returns `undefined`; every resulting slice then becomes the full 8,160-character transcript and it is synthesized 11 times. Choose a reachable boundary while preserving nonempty/max-weight constraints, and add a regression for this case.

src/core/utils/gemini_speech.ts
```diff
@@ -432,17 +634,69 @@ async function requestGeminiGenerateContent(
...
+      if (done) {
+        const payload = parseSseEvent(buffered);
+        if (payload !== undefined) {
+          yield payload;
+        }
+        return;
+      }
+    }
+  } finally {
+    reader.releaseLock();
```
markustherobot, line 702: **[P2] Cancel failed SSE bodies before retrying**

When parsing or finish validation fails before audio, the retry path opens another request, but this `releaseLock()` does not cancel the current response body. If a later retry succeeds, the outer controller completes without aborting, so the abandoned Gemini stream can remain open and continue consuming a connection and generation work. Cancel the reader/body on abnormal exit, or give each attempt a controller that is aborted before retrying.

src/tui/speech_playback.ts
```diff
@@ -51,84 +90,126 @@ export async function runSpeechPlaybackTask(args: {
       }
 
       playbackStarted = true;
-      totalChunks = chunk.total;
+      totalSegments = chunk.total;
       refreshSpeechProgress();
+      await writePlaybackAudio(playbackInput, chunk.audio, abortController.signal);
```
markustherobot: **[P2] Await player shutdown on cancellation**

If Escape lands after the iterator has produced a chunk but before the loop body runs, the early return above skips the catch block's `await playback`. The `finally` block sends the abort, but `runSpeechPlaybackTask` resolves immediately, so the controller can clear the task or finish disposal while the detached process group is still in SIGTERM/SIGKILL cleanup. Let this write path observe the abort, or explicitly await `playback` before returning.

markustherobot (commented):
Found a high-impact segmentation failure for valid CJK input, plus stream and player lifecycle failures that can leak work or mask playback errors. I launched 3 scoped subreviews covering Gemini streaming semantics, segmentation and Telegram, and TUI playback lifecycle; all completed.

src/tui/speech_playback.ts
```diff
@@ -51,84 +90,126 @@ export async function runSpeechPlaybackTask(args: {
...
+      throw outcome.error;
     }
-  } catch (err) {
-    if (args.signal.aborted) {
+    if (args.signal.aborted || outcome.result.aborted) {
       return;
     }
-    const error = err as NodeJS.ErrnoException;
-    if (error.name === "AbortError") {
+    if (outcome.result.exitCode !== 0) {
```
markustherobot, line 113: **[P2] Do not treat every zero ffplay exit as successful playback**

`ffplay` can emit fatal errors such as `audio open failed` and `Failed to open file 'pipe:0' or configure filtergraph` while still exiting with code 0. With this exit-code-only check, `/speak` can clear its status without playing audio or showing a failure. Make fatal playback errors produce a nonzero result, or inspect the captured stderr for fatal termination before reporting success.

src/tui/speech_playback.ts
```diff
@@ -51,84 +90,126 @@ export async function runSpeechPlaybackTask(args: {
...
-    if (error.code === "ENOENT") {
-      throw new Error("afplay not found.");
+    if (
+      isMissingExecutableError(error) ||
+      ("error" in outcome && isMissingExecutableError(outcome.error))
+    ) {
+      throw new Error("ffplay not found. Install ffmpeg to use /speak.");
     }
-    throw err;
+    throw error;
```
markustherobot, line 140: **[P2] Preserve player diagnostics when stdin closes early**

If `ffplay` exits while Gemini is still generating, the next `input.write` rejects with `ERR_STREAM_DESTROYED` or `EPIPE`. This catch awaits the completed player but ignores its result unless it was `ENOENT`, then rethrows the low-level stream error and discards the captured stderr, exit code, and signal. After the cancellation and missing-executable checks, prefer `playbackFailure(outcome.result)` when the player has already terminated.

src/tui/speech_playback.ts
markusylisiurunen: if there is no robust way to detect actual failure, then I think we should accept that limitation and not try to find fragile heuristics around it (e.g. detect random potential words from the output, like "failed").

markusylisiurunen:
@markustherobot address the findings

src/core/utils/gemini_speech.ts
markustherobot: Fixed in 9895164. Segmentation now computes the minimum feasible segment count from reachable character boundaries and only selects cuts whose suffix remains partitionable. The reported Japanese transcript is covered by a regression that verifies complete, nonduplicated text and the 2,040-weight cap.

src/core/utils/gemini_speech.ts
markustherobot: Fixed in 9895164. The SSE reader now cancels its body on every abnormal exit before releasing the lock, while preserving the original parse or validation error. A retry regression verifies cancellation before the next request succeeds.

src/tui/speech_playback.ts
markustherobot: Fixed in 9895164. I removed the pre-write early return so cancellation is observed by the write path, which aborts and awaits the player cleanup. The regression keeps the task pending until the mocked detached player finishes shutdown.

src/tui/speech_playback.ts
markustherobot: I verified that ffplay does not expose a supported exit-on-error option here (`-xerror` is rejected as an unknown option). Per the guidance in this thread, I left the zero-exit limitation as-is and did not add stderr keyword heuristics.

src/tui/speech_playback.ts
markustherobot: Fixed in 9895164. When ffplay has already terminated independently, a stdin write failure now reports the captured playback result and stderr instead of the low-level EPIPE. Cancellation-induced exits still preserve the original cancellation path, and the diagnostic behavior has a regression.

markusylisiurunen:
@markustherobot review

src/core/utils/gemini_speech.ts
```diff
@@ -401,6 +460,132 @@ async function synthesizeSpeechAudioChunk(args: SynthesizeSpeechAudioChunkArgs):
...
+  args: StreamSpeechSegmentArgs,
+): AsyncGenerator<Buffer> {
+  const maxAttempts = Math.max(1, Math.trunc(args.maxAttempts));
+  let lastError: unknown;
+
+  for (let attempt = 1; attempt <= maxAttempts; attempt += 1) {
+    let emittedAudio = false;
+    try {
+      for await (const audio of requestGeminiSpeechStream(args)) {
+        emittedAudio = true;
```
markustherobot, line 493: **[P2] Keep retries available until buffered audio starts playback**

With the new 500 ms startup buffer, a transient stream failure after the first sub-threshold PCM event sets `emittedAudio` and disables retries even though `ffplay` has not been spawned and the user has heard nothing. This turns a recoverable pre-playback failure into a failed `/speak`; keep the initial buffered window inside the retry owner, or otherwise let the consumer discard and retry it until playback actually starts.

src/tui/speech_playback.ts
```diff
@@ -16,119 +27,204 @@ export async function runSpeechPlaybackTask(args: {
...
-      if (playback.exitCode !== 0) {
-        const detail =
-          playback.exitCode !== null
-            ? `afplay exited with code ${playback.exitCode}`
-            : playback.closeSignal
-              ? `afplay terminated by signal ${playback.closeSignal}`
-              : "afplay exited";
-        throw new Error(detail);
-      }
+      await writePlaybackAudio(playbackInput!, chunk.audio, abortController.signal);
```
markustherobot: **[P2] Observe early player exit while waiting for Gemini**

After startup, this loop only notices that `ffplay` exited when another chunk reaches `writePlaybackAudio`. If the player terminates after accepting the startup buffer while the next SSE read is pending, `/speak` remains stuck in `playing speech` and continues generation despite there being no audio sink. Race the active player outcome with stream progress and abort/report any exit that occurs before stdin is intentionally ended.

markustherobot (commented):
Found two playback lifecycle regressions around failure handling before and after playback starts. I launched 3 scoped subreviews covering Gemini streaming and segmentation, TUI player lifecycle, and Telegram complete-generation compatibility; all completed.

markusylisiurunen:
@markustherobot address the findings

src/core/utils/gemini_speech.ts
markustherobot: Fixed in 111bb74. The 500 ms PCM startup buffer now lives inside the retrying first-stream generator, so a failed sub-threshold attempt is discarded and retried before any bytes reach the playback consumer. A regression verifies that only the successful retry is emitted.

src/tui/speech_playback.ts
markustherobot: Fixed in 111bb74. The playback task now observes the active ffplay outcome concurrently and aborts the linked Gemini stream if the player settles before stdin shutdown begins, then reports the captured player diagnostics. A stalled-SSE regression verifies prompt cancellation and failure.

PR #444, opened 2026-08-15
upgrade Gemini Flash persona

markusylisiurunen:
## why

Gemini 3.7 Flash is available in the current pi-ai model catalog and should replace Gemini 3.6 Flash as Tau's built-in Gemini persona.

## what

Updates the built-in chat persona and its regression coverage to use `gemini-3.7-flash`. The persona retains the model's supported low, medium, and high reasoning efforts.

The repository was already pinned to the latest published pi-ai and pi-tui release, `0.84.2`, so no dependency version change was required.

### DIFF
--- a/src/core/personas.ts
+++ b/src/core/personas.ts
@@ -212,10 +212,10 @@ type PersonaSpec = {
 
 const PERSONA_SPECS: PersonaSpec[] = [
   {
-    id: "gemini-3.6-flash",
-    description: "Gemini 3.6 Flash",
+    id: "gemini-3.7-flash",
+    description: "Gemini 3.7 Flash",
     provider: "google",
-    modelId: "gemini-3.6-flash",
+    modelId: "gemini-3.7-flash",
     allowedReasoningLevels: ["low", "medium", "high"],
     settings: { reasoning: "medium" },
     skills: "*",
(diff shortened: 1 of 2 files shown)


src/core/personas.ts
```diff
@@ -212,10 +212,10 @@ type PersonaSpec = {
 
 const PERSONA_SPECS: PersonaSpec[] = [
   {
-    id: "gemini-3.6-flash",
-    description: "Gemini 3.6 Flash",
+    id: "gemini-3.7-flash",
```
markustherobot, line 215: **[P1] Preserve the persona when recovering existing sessions**

A stored session whose durable `personaId` is `gemini-3.6-flash-chat` no longer matches the current catalog after this replacement. `applySnapshotSettingsToBootstrap()` then silently falls back to the configured default (normally Opus), while the committed snapshot still reports the removed Gemini ID, so the next turn can run a different model/provider or fail for a user who only has Google credentials. Add owning-boundary recovery normalization from the missing 3.6 persona to the 3.7 persona, persist the canonical ID, and cover restart plus next-turn behavior with a representative older snapshot.

markustherobot (commented):
Reviewed the complete persona/catalog change and the surrounding configuration and recovery paths. I found one durable-session compatibility issue that needs addressing before this is safe to ship; fresh-session model resolution and the pinned catalog otherwise look consistent. I launched 2 subreviews covering persona resolution/recovery and pi-ai catalog/tests, and both completed successfully.

markusylisiurunen:
@markustherobot does this not require docs changes?

PR #446, opened 2026-08-15
increase Telegram attachment limits

### LINKED ISSUE #445: increase telegram attachment count limits

markustherobot:
## why

Telegram turns were limited to 10 attachments and 50 MiB in aggregate, which is too restrictive for larger batches of otherwise supported files. Issue #445 established a higher count and aggregate allowance.

## what

- Raised the per-turn attachment count limit to 32.
- Raised the per-turn aggregate attachment limit to 100 MiB.
- Kept the 20 MiB per-file limit because Tau uses Telegram's hosted Bot API download endpoint.
- Added an end-to-end regression covering 32 queued attachments totaling exactly 100 MiB and rejection of the 33rd attachment.
- Updated the Telegram documentation with the new limits.

## details

I treated the latest 100 MiB request as replacing the earlier 72 MiB proposal while retaining the agreed 32-file count and 20 MiB per-file cap. There are no remaining questions.

fixes #445


markustherobot (commented):
The runtime limit changes are internally consistent, but the regression coverage does not fully lock down the new aggregate boundary. I launched two read-only subreviews covering attachment runtime/resource behavior and the docs/test contract.

test/telegram_adapter.test.js
```diff
@@ -2172,6 +2172,88 @@ describe("telegram adapter", () => {
     }
   });
 
+  it("accepts 32 attachments totaling 100 MiB per turn", async () => {
```
markustherobot, line 2175: **[P2] Exercise the aggregate rejection path**

This case reaches 32 queued files at exactly 100 MiB, so the 33rd file is rejected by the count check before the aggregate-size check runs. It would still pass if the total cap were removed or raised, leaving the changed aggregate contract unprotected. Add a separate case with fewer than 32 files that reaches exactly 100 MiB, then queues one more byte and asserts the `exceeds per-turn total limit (100 MiB)` warning.

PR #447, opened 2026-08-16
refresh model catalogs from pi.dev

markusylisiurunen:
## why

Tau currently receives updated model metadata only when its pinned `pi-ai` dependency changes. This leaves newly published models, corrected limits, compatibility flags, and pricing unavailable until the next Tau release.

## what

Add a host-owned remote model catalog backed by the version-compatible provider shards published at `pi.dev`. Tau restores the cached catalog during bootstrap, performs a startup-only asynchronous freshness check, preserves successful provider updates when other shards fail, and keeps live sessions on their captured catalog until `/reload`.

Add `tau models refresh` for forced best-effort refreshes, `TAU_OFFLINE` for disabling automatic checks, a versioned atomic cache at `~/.config/tau/models-store.json`, and documentation for precedence and trust boundaries.


src/core/models/remote_catalog.ts
```diff
@@ -0,0 +1,389 @@
...
+  }
+
+  async apply(updates: readonly ProviderUpdate[]): Promise<StoreDocument> {
+    if (updates.length === 0) return this.read();
+    mkdirSync(dirname(this.path), { recursive: true, mode: PRIVATE_DIRECTORY_MODE });
+    const release = await acquireLock(`${this.path}.lock`);
+    try {
+      const document = this.read();
+      for (const update of updates) {
+        document.providers[update.provider] = structuredClone(update.entry);
```
markustherobot: **[P2] Prevent stale concurrent refreshes from replacing newer shards**

When two Tau processes refresh the same provider, each fetches from the pre-lock snapshot, and the slower one blindly replaces whatever is current after acquiring the lock. A delayed `304` for an old ETag (or an older `200` from a lagging edge) can therefore overwrite a newer shard committed by the other process and mark the stale data fresh for four hours. Carry the expected ETag/revision into `apply()` and skip or retry the update when the locked document no longer matches it (and compare `lastModified` before replacing).

markustherobot (commented):
I found one concurrency issue in the cross-process cache publication path. I launched three subreviews covering core storage/networking, host lifecycle and session capture, and model merge/external contracts; all completed.

PR #449, opened 2026-08-16
enrich OpenAI live transcription context

### LINKED ISSUE #448: voice input fails
```
 failed to transcribe speech
 Invalid 'session.audio.input.transcription.prompt': string too long. Expected a string with maximum length 1024, but got
 a string with length 6342 instead.
 recording retained at /var/folders/5r/rd6xn8fd4mxgdr1k307bs8n80000gn/T/tau-listen.o2oEcu
 run /listen retry to try again, or /listen discard to delete it
```

markustherobot:
## why

OpenAI Realtime rejects transcription prompts longer than 1,024 characters, while Tau previously placed token-budgeted raw conversation context in that field. The fix needs to preserve the shared recent-context contract without imposing Realtime's narrow prompt limit on Gemini or OpenAI file transcription.

## what

OpenAI live transcription now starts a direct `gpt-5.6-luna` Responses API request alongside the Realtime connection. Luna receives the existing provider-neutral recent speech context and returns relevance-ordered transcription hints through structured output. Tau validates those hints and sends them through `gpt-live-transcribe`'s native `keywords` field while keeping `prompt` short and static.

Audio remains buffered until keyword preparation and Realtime session configuration complete. Extraction timeout, provider failure, malformed output, and unusable hints all fall back to transcription without keywords. Cancellation aborts both the keyword request and Realtime session. OpenAI file transcription retains its larger raw-context prompt, and Gemini and Mistral behavior is unchanged.

Regression coverage exercises the direct Luna request, exact supplied context, audio buffering, Unicode and keyword normalization, failure fallback, cancellation, the bounded Realtime prompt, and the independently larger file-transcription prompt. The TUI speech documentation now describes the additional OpenAI request and fallback behavior.

## details

I chose a five-second best-effort extraction deadline with low reasoning effort to overlap normal recording startup without making enrichment a long-lived dependency. Provider-boundary safety limits are 100 keywords, 100 Unicode characters per keyword, and 1,024 aggregate keyword characters; Luna is not asked to satisfy those limits. Responses storage is disabled. No keyword request is made when the shared context is empty. There are no remaining implementation questions.

fixes #448


src/core/utils/openai_transcription.ts
```diff
@@ -480,7 +547,91 @@ function createPcmWav(pcm: Buffer): Buffer {
...
+        model: OPENAI_TRANSCRIPTION_KEYWORD_MODEL,
+        instructions: OPENAI_TRANSCRIPTION_KEYWORD_INSTRUCTIONS,
+        input: formattedContext,
+        reasoning: { effort: "low" },
+        max_output_tokens: 2_048,
+        store: false,
+        text: OPENAI_TRANSCRIPTION_KEYWORD_TEXT_CONFIG,
+      }),
+    });
+    if (!response.ok) return [];
```
markustherobot: **[P2] Consume failed keyword responses before returning**

When the Responses request returns a realistic 4xx/5xx (for example, rate limiting or a temporary provider outage), this exits without consuming or cancelling `response.body`. Node’s Undici-backed `fetch` retains connection resources for an unread body until collection, so repeated `/listen` attempts during an outage can accumulate unreusable connections and eventually stall later requests. Consume a bounded error body or await `response.body?.cancel().catch(() => {})` before falling back to an empty keyword list.

markustherobot (commented):
Found one resource-lifecycle issue in the new best-effort fallback. I launched three subreviews covering async lifecycle, provider contracts, and integration/tests; all three completed.

PR #451, opened 2026-08-17
make history replication race-safe

### LINKED ISSUE #450: make history replication safe across concurrent Tau processes
## problem

Multiple Tau processes on the same host share the local history SQLite database and durable replication outbox. Each process independently scans that outbox, so starting a new Tau session while another session is writing history can cause both processes to send the same pending operation concurrently.

The history Worker checks the `operations` table before applying an operation, but that check happens outside the D1 mutation batch. Two requests can both observe that the operation is absent. One commits successfully, while the other encounters a uniqueness conflict and returns:

```json
{"error":{"code":"internal_error","message":"Internal server error"}}
```

The local client logs this as `history_replication_failed`. Replication later recovers because the operation was committed by the competing request, but the failure is noisy and the Worker is not fully idempotent under concurrent delivery.

## expected behavior

Concurrent delivery of the same replication operation should have exactly-once effects and return success for both requests. At-least-once outbox delivery, including retries after ambiguous HTTP outcomes, must not produce an internal service error.

## proposed direction

After a D1 mutation batch fails, re-check the exact operation marker. Because the marker and operation effects are committed in the same atomic batch, an existing marker proves that another request committed the operation and the duplicate can be treated as already applied. Re-throw the original error when the marker is still absent.

Add focused regression coverage for the concurrent-commit path and retain unexpected storage failures as `500` responses.

markusylisiurunen:
## why

Multiple Tau processes can concurrently deliver the same operation from their shared history outbox. The Worker’s idempotency check occurs before its D1 mutation batch, allowing both requests to pass the check and one to surface a uniqueness conflict as an internal service error even though the competing request committed the operation.

Replication failures were also written directly with `console.error`. An asynchronous failure could therefore interleave raw JSON with the active terminal renderer and corrupt the TUI.

## what

Reconcile a failed mutation batch against the exact operation marker. If another request committed the operation, report it as already applied; if the marker remains absent or reconciliation is unavailable, preserve the original storage failure.

Route structured replication diagnostics through an injected owner instead of writing from history core. Headless commands retain structured stderr logs, the TUI shows a concise footer notice, and in-process SDK clients can opt into diagnostics with `onDiagnostic` without receiving unsolicited process output.

Add regression coverage for concurrent commits, genuine storage failures, diagnostic reporter isolation, TUI presentation, and the published SDK type boundary.

closes #450


src/history/worker/index.ts
```diff
@@ -483,8 +479,21 @@ export async function applyOperation(database: D1Database, operation: Operation)
       .prepare("INSERT INTO operations (operation_id, session_id, applied_at) VALUES (?, ?, ?)")
       .bind(operation.id, operation.sessionId, appliedAt),
   );
-  await database.batch(statements);
-  return true;
+  try {
+    await database.batch(statements);
+    return true;
+  } catch (error) {
+    if (await operationExists(database, operation.id)) return false;
```
markustherobot: **[P2] Preserve the batch error when reconciliation fails**

When a D1 outage or transient storage failure makes both `database.batch()` and this follow-up SELECT reject, the `operationExists()` error escapes and the original mutation failure is never rethrown. That defeats the stated error-preservation behavior and hides the actual batch diagnostic from `history_request_failed` logs. Catch a failed reconciliation read separately and rethrow the original batch error unless the marker was positively found.

markustherobot (commented):
I found one reliability issue in the failure-reconciliation path. I launched 2 focused subreviews covering D1 race and storage correctness, plus history test and client integration behavior; both completed successfully.

src/history/worker/index.ts
markusylisiurunen: Fixed in 111ca68. Reconciliation reads are now best-effort, and a failed read rethrows the original batch error. Added regression coverage for both failures occurring together.

PR #454, opened 2026-08-17
improve compacted history retrieval

markusylisiurunen:
## why

Agents can confuse current-session auto-compaction archives with the separate history collection, then recover context through stale remote history, guessed searches, or oversized transcript dumps. The archive formats also lacked local documentation that survives compaction as an explicit lookup contract.

## what

Add a private `README.md` beside each agent's numbered auto-compaction archive pairs, documenting file semantics, the JSON shape, sequencing across compactions, and adaptable bounded lookup examples. Require the continuation to load that guide into model context while retaining concise source-selection and retrieval defaults.

Distinguish archive entry references from history transcript ids, prefer execution-environment archives for current-session recovery, and expand the history tool's progressive documentation with bounded overview and selected-entry drill-down examples.


src/core/session/compaction.ts
```diff
@@ -627,15 +627,15 @@ export function buildAutoCompactionContinuationMessage(args: {
...
-      "When the compaction summary mentions a history entry id, search the numbered text transcripts for that id first, then inspect the corresponding JSON record if the text transcript is truncated or incomplete.",
-      "When retained output is marked as truncated, search the text transcript by tool name and distinctive surrounding text, then inspect the paired JSON record for the complete output.",
-      "Prefer narrow searches and bounded reads of the text transcripts; their tool results are middle-truncated. The JSON files retain untruncated archived content and may be very large.",
-      "When available, delegating a precise archive lookup to a low-effort subagent can preserve this context more efficiently than reading large sections directly.",
-      "These files are temporary and may no longer exist.",
+      "Before continuing, ensure the archive guide's full contents are present in the current model context. If they are not already visible in full, read the guide now with an execution-environment file tool. This is required even when no archive lookup is currently planned.",
```
markustherobot, line 634: **[P1] Keep compaction usable without filesystem tools**

`tools: []` is a supported main-persona and subagent configuration, but those runtimes receive only intrinsic `tau_docs` (plus main-session goal tools), none of which can read this execution-environment path. Because archiving still succeeds and this hidden continuation makes the read mandatory, such an agent cannot satisfy the continuation contract after automatic compaction. Include the guide content through a guaranteed channel, or make the requirement conditional with a usable fallback when no read-capable tool is present.

src/core/static/code_mode/history/documentation.md
```diff
@@ -122,3 +122,71 @@ Entry timestamps are Unix milliseconds. The transcript is the active flat conver
...
+  const kept = max - marker.length;
+  const head = Math.ceil(kept / 2);
+  return chars.slice(0, head).join("") + marker + chars.slice(-(kept - head)).join("");
+};
+const render = (entry) =>
+  entry.type === "tool"
+    ? `${text(entry.arguments)} ${text(entry.result)}`
+    : text(entry.content);
+
+const page = await history.read({ sessionId, limit: 100 });
```
markustherobot, line 157: **[P1] Project history before crossing the code-mode bridge**

This projection runs only after `history.read` has returned the complete raw page through code mode's 1 MiB response bridge. History pages are allowed to reach 12 MiB, and a single local entry can exceed 1 MiB, so realistic sessions with large tool or image payloads fail before `render()` can bound anything. The overview needs a server-side bounded/projection API (or equivalent response shaping) rather than a post-read JavaScript projection.

src/core/static/code_mode/history/documentation.md
```diff
@@ -122,3 +122,71 @@ Entry timestamps are Unix milliseconds. The transcript is the active flat conver
...
+After selecting references, scan transcript pages in code and print only matching entries instead of emitting every page:
+
+```js
+const sessionId = "selected-session-id";
+const wantedSuffixes = ["12345678", "90abcdef"];
+const found = new Set();
+let cursor;
+
+do {
+  const page = await history.read({ sessionId, limit: 100, cursor });
```
markustherobot, line 179: **[P1] Omit the undefined cursor on the first read**

Code mode rejects `undefined` anywhere in API arguments as non-JSON-serializable. Since `cursor` is initially undefined, the first loop iteration always throws before `history.read` runs. Build the options with a conditional cursor property so the initial request omits it.

src/core/static/code_mode/history/documentation.md
```diff
@@ -122,3 +122,71 @@ Entry timestamps are Unix milliseconds. The transcript is the active flat conver
...
+  for (const entry of page.entries) {
+    const ids = [entry.id, ...entry.sourceIds];
+    const matches = wantedSuffixes.filter((suffix) =>
+      ids.some((id) => id.endsWith(suffix)),
+    );
+    if (matches.length === 0) continue;
+
+    for (const suffix of matches) found.add(suffix);
+    const label = entry.type === "tool" ? `tool ${entry.name} ${entry.outcome}` : entry.type;
+    console.log(`[${label} id=${entry.id}] ${middle(render(entry), 4000)}`);
```
markustherobot: **[P1] Make the drill-down program self-contained**

History code-mode calls are one-shot and do not retain JavaScript definitions from an earlier invocation. This block calls `middle(render(entry), 4000)` without defining either helper, so it throws as soon as a selected reference matches. Include the helper definitions in this program (or present one complete combined example).

markustherobot (commented):
The recovery guidance is not yet reliable across Tau's supported tool configurations and code-mode boundaries. I found several execution-path gaps in the new operational documentation that should be addressed before merge.

I launched three read-only subreviews covering archive writer lifecycle, compaction integration, and history retrieval documentation, then independently verified the candidate findings against this head.

PR #455, opened 2026-08-18
refine tool result messages

markusylisiurunen:
## why

Built-in tools returned inconsistent validation, failure, cancellation, and execution-status text. Some responses exposed internal `(tau)` markers or made appended status indistinguishable from command and program output, while similar file and subagent operations used different wording and outcome classifications.

## what

Align built-in tool responses around focused argument guidance, concise success text, contextual `Could not ...` failures, and operation-specific cancellation messages. Bash and code-mode execution now separate Tau-authored notices from captured output with bracketed status text, and file, goal, documentation, and subagent tools use stricter argument boundaries and consistent semantic outcomes.

## details

Adds focused regression coverage for exact model-facing text, unknown arguments, filesystem and supervisor failures, termination behavior, and code-mode output separation.

### DIFF
--- a/src/core/tools/bash.ts
+++ b/src/core/tools/bash.ts
@@ -11,6 +11,7 @@ import {
   type TruncationResult,
   truncateForTokens,
 } from "../utils/truncate.js";
+import { formatZodError } from "../utils/zod.js";
 import type { ToolActivity } from "./activity.js";
 import type { ToolExecutionBackend } from "./execution_backend.js";
 import {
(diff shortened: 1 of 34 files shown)


src/core/tools/bash.ts
```diff
@@ -460,43 +533,54 @@ export function createBashToolDefinition(backend: ToolExecutionBackend, cwd: str
...
+            });
 
             const outputPolicy = getBashOutputPolicy({
               mode: "model",
               maxOutputTokens,
               hasMaxOutputTokens,
             });
             const truncationInfo = await prepareBashOutput(
-              aborted ? stripBashAbortNote(output) : output,
+              termination ? stripBashTerminationNote(output, termination.backendNote) : output,
```
markustherobot: **[P2] Preserve the termination reason in the tool card**

When Bash times out or exits by signal, this strips the backend marker before building `truncationInfo`, and the canonical replacement is appended only to `toolText`. The TUI renders `buildBashPresentation(...)`, not the model-facing result, so the card now shows only partial output plus a generic cancelled/failed state and loses whether the command timed out or which signal ended it. Add the canonical notice as a separate presentation detail or metadata item so it remains distinct from captured output without disappearing from the client.

src/core/tools/view_image.ts
```diff
@@ -316,7 +317,10 @@ export function createViewImageToolDefinition(backend: ToolExecutionBackend): Ag
           return { content: outcome.content, outcome: outcome.outcome, uiEvent };
         } catch (e) {
           const errorMessage = e instanceof Error ? e.message : String(e);
-          return blocked(`Tool view_image failed: ${errorMessage}`, "failed");
+          if ((e as NodeJS.ErrnoException).code === "ENOENT") {
```
markustherobot: **[P2] Keep missing-image classification consistent on hosted backends**

This branch works only for the local backend, which preserves Node's `ENOENT` code. The Fly worker serializes only `err.message` and reconstructs a plain `Error`, while the Cloudflare bridge also throws a new HTTP `Error`, so the common missing-path case falls through to `failed` with `Could not view image...` for hosted sessions instead of the new corrective `blocked` result. Preserve a generic not-found code through the execution-backend boundary (or otherwise normalize it there) so this tool has the same outcome contract in local and hosted environments.

markustherobot (commented):
I found two cross-boundary regressions in the new tool-result contract, covering presentation and hosted-environment consistency.

I launched three read-only subreviews covering process output semantics, subagent and goal outcomes, and validation/file tools; all completed successfully, and I independently verified their candidates against the full PR.

PR #456, opened 2026-08-18
refine code mode boundaries

markusylisiurunen:
## why

Code-mode rejected options objects containing `undefined`, even though those properties should behave as omitted optional fields. Its progressive-disclosure instruction also encouraged agents to reload tool documentation for each task instead of reusing documentation already visible in conversation context.

## what

Omit `undefined` object properties when serializing code-mode API arguments, while continuing to reject direct undefined arguments and undefined array entries so they never become `null`.

Require code-mode documentation to be read and visible before API use, but explicitly avoid reloading it while it remains in context. Align the bundled product documentation and add regression coverage for the serialization behavior.


docs/tools.md
```diff
@@ -141,15 +141,15 @@ The command uploads the PDF to Mistral and attempts to delete the remote upload
 
 `web`, `history`, and `nook` each run a one-shot JavaScript program in Tau's restricted code-mode runtime. They intentionally disclose their exact API at use time rather than embedding signatures in this page.
 
-The first useful call must only print the tool's documentation:
+Before using a tool's API, its documentation must be visible in the conversation context. If it is not already visible, make one documentation-only call:
```
markusylisiurunen: all of these docs edits are written as if they are instructions to an agent. they are not. they document Tau's behavior to both human or agent readers who are interested in understanding how Tau works. they should not be written as instructions to the Tau agent. make sure AGENTS.md is clear about this as well so this won't happen again.

src/code_mode/runtime.ts
```diff
@@ -160,9 +160,10 @@ export function buildTauCodeModeToolDescription({
   if (!trimmedDescription) throw new Error("code-mode tool description must not be empty");
   return [
     trimmedDescription,
-    "When this tool is useful, your first call must be a documentation-only program that does nothing except print docs with console.log(docs).",
+    "Before using the API, check whether this tool's documentation is already visible in the conversation context.",
```
markusylisiurunen: I find the "When this tool is useful" wording better honestly but this should just be clear that docs should be read once when they aren't already visible and then the tool can be used normally afterwards. think hard about the wording.

markusylisiurunen:
@markustherobot address my comments

docs/tools.md
markustherobot: Rewrote the changed public documentation as third-person product behavior for human and agent readers, and added an AGENTS.md rule separating public docs from model-facing instructions.

src/code_mode/runtime.ts
markustherobot: Restored the “When this tool is useful” lead-in and clarified the conditional lifecycle: a documentation-only first call when docs are absent, followed by normal API use without reloading while they remain visible.

PR #459, opened 2026-08-20
refine Telegram composite project config

markusylisiurunen:
## why

Composite workspaces currently inject every discovered member `AGENTS.md` into the parent session, even before work targets that repository. Composite projects also need a project-scoped way to allow model overrides for the built-in default subagent.

## what

Generate an empty composite root config by default and instruct the agent to read each relevant repository's root `AGENTS.md` before making changes. Add composite-only `subagents.defaultLaunchModels` configuration and write it through to the generated root config so the normal Tau runtime owns model resolution and enforcement.

Update Telegram configuration validation, workspace coverage, and public documentation for the new contract.

### DIFF
--- a/src/core/telegram/workspace.ts
+++ b/src/core/telegram/workspace.ts
@@ -610,39 +610,6 @@ async function prepareRepositoryWorkspace(options: {
   return projectCwd;
 }
 
-async function pathIsFile(path: string): Promise<boolean> {
-  const stats = await stat(path).catch((error) => {
-    if (getErrorCode(error) === "ENOENT") {
-      return undefined;
-    }
-    throw error;
-  });
-  return stats?.isFile() ?? false;
-}
-
-async function collectMemberAgentContextFiles(options: {
-  workspacePath: string;
-  memberProjectId: string;
-  project: TelegramRepositoryProjectConfig;
-}): Promise<string[]> {
-  const memberPath = join(options.workspacePath, options.memberProjectId);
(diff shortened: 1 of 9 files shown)


src/core/telegram/workspace.ts
```diff
@@ -688,27 +655,13 @@ async function writeCompositeWorkspaceFiles(options: {
...
-    agentContextFiles.push(
-      ...(await collectMemberAgentContextFiles({
-        workspacePath: options.workspacePath,
-        memberProjectId,
-        project: memberProject,
-      })),
-    );
-  }
-
+  const config = options.project.subagents ? { subagents: options.project.subagents } : {};
```
markustherobot, line 658: **[P2] Refresh generated metadata for preserved composites**

On restart, `reconnectSession` skips `prepareWorkspace` whenever the composite root still exists, so this writer never runs for preserved sessions. After upgrading, those workspaces therefore keep the old `agentContextFiles` config, and adding or changing `subagents.defaultLaunchModels` in the Telegram project also has no effect on the reconnected runtime. Refresh the generated root `AGENTS.md` and `.tau/config.json` during composite recovery (without recloning or rerunning provisioning), and cover the preserved-workspace path in a regression test.

markustherobot (commented):
I found one recovery/upgrade issue in the generated composite metadata path. I also launched three read-only subreviews covering config/model enforcement, workspace context, and docs/regression coverage; all three completed successfully.

PR #461, opened 2026-08-21
make session snapshots projection-owned

### LINKED ISSUE #460: a bug with delivering queued message
the TUI showed this error:

```
 failed to queue message
 failed to drain pending user message: session delta produced an invalid snapshot: messages: duplicate message id 'history-4b330767-cf79-4c45-b881-02e65303bb00'
```

the agent was writing its final response while I queued a follow-up. then this error showed right after the response streaming completed.

markustherobot:
## why

Queued user-message acceptance mutates agent history before its awaited semantic event is projected by the host. Ordinary snapshot refreshes rebuilt from that live history whenever the runtime appeared idle, so a refresh could commit the message without its turn and timeline state, then reject the event's canonical `message.append` as a duplicate.

Persistence failures could also roll protocol state back while runtime-owned rewind, compaction, interrupted-message, or configuration mutations remained applied. Serving snapshots after that divergence could combine stale transcript state with newer runtime revisions.

## what

- Make the host's committed protocol projection authoritative for every ordinary snapshot refresh.
- Initialize new hosted sessions with a canonical in-memory projection instead of treating projection state as optional.
- Keep full runtime reconciliation at explicit host-owned transitions, and reconcile restored runtime state eagerly at the recovery boundary.
- Fail closed when a failed runtime-owning transition leaves runtime state ahead of the committed projection.
- Add deterministic regressions for queued event projection and failed rewind/configuration persistence.

## details

New sessions now begin with a revision 1 system-only projection, so the first user-message event advances the snapshot to revision 2. Protocol revisions are opaque ordering values, and this gives snapshots a valid source of truth before any runtime event can race with them.

Runtime-owning operations now guard their failure boundaries. Configuration mutations invalidate the live handle when their queued commit fails. Rewind, compaction, and interrupted-assistant mutations compare agent state before and after failure and invalidate only when runtime state changed. Invalidated handles reject later snapshots and mutations, preserving the last committed store state for explicit disposal and recovery.

fixes #460

### DIFF
--- a/src/host/local_session_host.ts
+++ b/src/host/local_session_host.ts
@@ -261,9 +261,7 @@ export class LocalSessionHost implements TauSessionHost {
           ? { usageCheckpoint: { ...recovered.snapshot.agentState.usageCheckpoint } }
           : {}),
       });
-      if (recovered.changed || agentRecovery.recoveredToolResults.length > 0) {
-        await hostedSession.persistRecoveredAgentState(agentRecovery);
-      }
+      await hostedSession.reconcileRecoveredState(agentRecovery);
       return hostedSession;
     } catch (error) {
       if (hostedSession) {
@@ -670,7 +668,7 @@ function normalizeDeltaCause(cause: LocalDeltaCause): SessionProtocolDeltaCause 
 
 class LocalHostedSessionHandle implements LocalHostedSession {
   readonly session: ChatRuntime;
-  private committedSnapshot?: SessionProtocolSnapshot;
+  private committedSnapshot: SessionProtocolSnapshot;
   private persistedSnapshot?: SessionProtocolSnapshot;
   private draftAssistantMessage?: SessionProtocolMessage;
   private readonly messageStates = new Map<string, SessionProtocolMessage["state"]>();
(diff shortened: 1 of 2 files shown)


src/host/local_session_host.ts
```diff
@@ -1599,14 +1595,10 @@ class LocalHostedSessionHandle implements LocalHostedSession {
 
   async snapshot(): Promise<SessionProtocolSnapshot> {
     this.assertActive();
-    return await this.enqueueMutation(async () =>
-      this.runtime.isTurnRunning && this.committedSnapshot
-        ? await this.commitProjectedSnapshot()
-        : await this.commitSnapshot(),
-    );
+    return await this.enqueueMutation(() => this.commitProjectedSnapshot());
```
markustherobot, line 1603: **[P1] Invalidate the projection after failed runtime transitions**

When `history_rewound` truncates `AgentRuntime` and its `commitSnapshot()` then fails, `enqueueMutation()` restores only protocol-side state, so runtime history remains truncated. This new snapshot path subsequently returns the old committed transcript instead of reconciling or failing, and the next `user_message` patch can persist that old transcript together with the post-rewind agent revision and new message. The session cannot retry the rewind coherently, and recovery can resurrect removed context. Runtime-owning transitions (including configuration mutations) must either roll the runtime back on persistence failure or mark the handle inconsistent so no snapshot or patch is served until an explicit safe reconciliation succeeds.

markustherobot (commented):
I found one high-priority state-integrity regression in the failure path. I launched three Luna subreviews covering projection/persistence semantics, recovery compatibility, and event concurrency; the recovery pass found no separate issue.

markusylisiurunen:
@markustherobot please address the findings if they are clear bugs with clearly right way to fix them

src/host/local_session_host.ts
markustherobot: Fixed in 76b0e56. Failed configuration transitions now invalidate the handle, while rewind, compaction, and interrupted-assistant transitions invalidate it when agent state changed. Subsequent snapshots and mutations fail closed, with regressions covering rewind and configuration persistence failures.

PR #462, opened 2026-08-21
expose browser diff reviews through the SDK

markusylisiurunen:
## why

Integrations can create and control Tau sessions through the public SDK, but the browser diff-review experience is currently tied to the TUI's local `/diff` lifecycle. Services that provision their own workspaces need a supported way to host the same review UI and ephemeral review agents without importing Tau internals or exposing Tau credentials to the browser.

Long-lived hosted reviews also need crash-safe state and submission boundaries. Review annotations should survive a client restart without treating expired ephemeral agent thread IDs as durable, and downstream work must be accepted before the browser is told that its one allowed submission succeeded.

## what

Add `startTauSdkDiffReview()` to the public SDK. It captures a diff through the session execution environment, creates the standard review-scoped ephemeral agent context, starts the built-in HTTP UI on the SDK client, and returns explicit result and close lifecycle handles. The browser now uses document-relative API routes so an embedding service can proxy it below a capability-protected path.

The helper accepts an optional opaque storage adapter. Tau versions and validates the stored document, fingerprints it against the captured diff, persists durable mutations before HTTP success, and rolls failed mutations back. Durable state excludes loading indicators and ephemeral model thread IDs; restored follow-ups fork from a fresh bootstrap and include the saved transcript.

An optional awaited `onSubmit` callback gives the embedding service a durable acceptance point. Each running review permits one in-flight or successful submission, while callback failure leaves it available for retry. The callback receives the composed review, diff command, and reviewed files without exposing diff-tool internals.


markustherobot (commented):
Found several correctness issues around the advertised proxy embedding and the new persistence lifecycle/concurrency behavior.

I launched four read-only subreviews covering persistence and submission, SDK lifecycle and bridge integration, the execution-backend move, and browser routing/public API behavior. All four completed successfully, and I independently verified the findings below against the current head.

src/diff_tool/app/src/api.ts
```diff
@@ -13,7 +13,7 @@ import type {
 } from "./types.js";
 
 async function request<T>(path: string, options: RequestInit = {}): Promise<T> {
-  const response = await fetch(path, {
+  const response = await fetch(new URL(path, document.baseURI), {
```
markustherobot, line 16: **[P1] Make the entire app prefix-relative**

Changing the fetch targets is not enough to support the documented nested proxy mount. The production Vite build still emits root-absolute `/assets/...` script and stylesheet URLs, and `App.css` requests root-absolute `/fonts/...` files, so a review served at `/reviews/123/` leaves the protected prefix and cannot load its JavaScript. Configure a relative build base and relative public-asset URLs (and normalize the mounted URL's trailing slash), then cover a nested-prefix load in a test.

src/diff_tool/http_server.ts
```diff
@@ -112,6 +133,15 @@ export class DiffToolHttpServer {
...
       this.context = await this.client.getContext();
       this.files = (await this.client.listFiles()).files;
+      if (this.storage) {
+        const diff = await this.client.getDiff();
+        this.scopeFingerprint = createDiffReviewScopeFingerprint(
+          this.context,
+          this.files,
+          diff.patch,
+        );
+        await this.restoreReviewState();
```
markustherobot: **[P2] Do not listen after protocol shutdown**

If the protocol closes while a slow `storage.load()` is pending, `closeFromProtocol()` marks this server closed before the HTTP listener exists. Once the load resolves, `start()` still calls `listen()`; `setUiText()` then fails on the closed client, and the catch-path `close()` no-ops because `closed` is already true, leaving an orphan listener that keeps the SDK process alive. Check the shutdown state before binding and make cleanup close any listener that appears after an earlier close.

src/diff_tool/http_server.ts
```diff
@@ -170,6 +200,57 @@ export class DiffToolHttpServer {
...
+    });
+    await previousMutation;
+
+    const previousState = this.reviewState.getState();
+    try {
+      const result = mutation();
+      await this.persistReviewState();
+      return result;
+    } catch (error) {
+      this.reviewState.replaceState(previousState);
```
markustherobot: **[P2] Keep transient loading flags out of rollback**

`previousState` includes runtime-only `thread.loading` and `brief.loading` values even though the persisted document deliberately excludes them. If a failed save overlaps an agent request, this whole-state rollback can clear `loading` while the request is still active (allowing a duplicate request), or restore `loading: true` after the agent request has failed and cleared it (causing permanent 409s until restart). Roll back only durable fields, or serialize the transient lifecycle transitions with the same queue.

src/diff_tool/http_server.ts
```diff
@@ -170,6 +200,57 @@ export class DiffToolHttpServer {
...
+      throw error;
+    } finally {
+      releaseMutation();
```
markustherobot: **[P2] Return the snapshot that was actually persisted**

`releaseMutation()` wakes the next queued request before the awaiting handler reads `getState()`. That next request can mutate state and wait on a slow save, so the first request's 200 response includes the second request's uncommitted state; if that save fails, the server rolls it back while the browser has already installed it from the successful response. Capture and return the committed state inside this queued transaction before releasing the next mutation.

src/diff_tool/http_server.ts
```diff
@@ -311,15 +392,15 @@ export class DiffToolHttpServer {
         return;
       }
 
-      this.reviewState.addReply(id, text);
+      await this.mutateReviewState(() => this.reviewState.addReply(id, text));
```
markustherobot: **[P2] Validate thread mutations inside the queue**

The thread existence check runs before this request waits for earlier persistence work. A queued delete can therefore remove the thread before `addReply()` executes; its `false` result is ignored and this endpoint returns 200 even though the reply was discarded. The resolve and collapse handlers have the same race. Perform the lookup and mutation atomically inside the queue and propagate the boolean result as 404/409.

src/sdk/diff_review.ts
```diff
@@ -0,0 +1,173 @@
...
+  let removeAbortListener = (): void => undefined;
+  try {
+    await bridge.start();
+    options.signal?.throwIfAborted();
+    const client = new DiffReviewProtocolClient(
+      parseDiffToolLaunchEnvironment(bridge.launchEnvironment),
+    );
+    server = new DiffToolHttpServer({
+      client,
+      ...(options.host !== undefined ? { host: options.host } : {}),
```
markustherobot, line 121: **[P2] Format IPv6 hosts in the returned URL**

The new public `host` option accepts an IPv6 literal such as `::1`, and Node can bind it, but `DiffToolHttpServer` constructs `http://::1:<port>`, which is not a valid URL. Format the bound address with IPv6 brackets (for example, `http://[::1]:<port>`) before exposing `review.url`.

markusylisiurunen:
@markustherobot deslop

PR #464, opened 2026-08-21
allow ephemeral reasoning overrides

### LINKED ISSUE #463: allow overriding ephemeral agent's reasoning effort
if omitted when creating an ephemeral agent, should inherit from parent agent.

markustherobot:
## why

Ephemeral contexts currently resolve reasoning only from their persona defaults. Clients need to choose an initial effort and update an individual thread's effort as part of a submission, including a fork's first turn.

## what

- add optional `reasoning` fields to ephemeral context creation and thread submission across the protocol and Node SDK
- capture the parent agent's effective reasoning when context creation omits an override
- retain reasoning as thread spec state, copy it into forks, and apply submission overrides before the submitted turn
- document the lifecycle and cover protocol validation, host routing, SDK forwarding, inheritance, retention, and fork behavior

## details

The implementation treats context reasoning as the initial value for newly created threads, not as mutable shared context state. An omitted creation value captures the parent agent spec at context creation, so later parent setting changes do not alter that context. A submission override updates only the target thread; a fork copies the source thread spec first and then applies its own override without mutating the source.

fixes #463

### DIFF
--- a/src/protocol/session_protocol.ts
+++ b/src/protocol/session_protocol.ts
@@ -236,12 +236,14 @@ export type SessionProtocolEphemeralAgentTool = "bash" | "write" | "edit" | "vie
 export type SessionProtocolEphemeralCreateParams = SessionProtocolSessionIdParams & {
   instructions: string;
   tools: SessionProtocolEphemeralAgentTool[];
+  reasoning?: SessionProtocolReasoningEffort;
 };
 export type SessionProtocolEphemeralSubmitParams = SessionProtocolSessionIdParams & {
   contextId: string;
   threadId: string;
   forkFromThreadId?: string;
   message: string;
+  reasoning?: SessionProtocolReasoningEffort;
 };
 export type SessionProtocolEphemeralCloseParams = SessionProtocolSessionIdParams & {
   contextId: string;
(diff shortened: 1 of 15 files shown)


src/protocol/session_protocol.ts
```diff
@@ -236,12 +236,14 @@ export type SessionProtocolEphemeralAgentTool = "bash" | "write" | "edit" | "vie
 export type SessionProtocolEphemeralCreateParams = SessionProtocolSessionIdParams & {
   instructions: string;
   tools: SessionProtocolEphemeralAgentTool[];
+  reasoning?: SessionProtocolReasoningEffort;
```
markustherobot, line 239: **[P1] Advance the session protocol version**

`SESSION_PROTOCOL_VERSION` remains 13 even though this adds new request fields. A new v13 SDK can therefore connect to an older v13 host, whose Zod schemas strip the unknown `reasoning` fields and return success, so both creation and submission overrides are silently ignored. Increment the protocol to 14 and update the version-coupled docs and tests so the exact-version handshake rejects mixed implementations instead of accepting requests with different semantics.

markustherobot (commented):
The runtime and SDK wiring otherwise look coherent, but the session protocol compatibility boundary needs correction before this ships.

I launched three Luna subreviews covering runtime and host semantics, protocol and SDK contracts, and tests and integration behavior; all completed successfully.

PR #465, opened 2026-08-22
add guide mode to diff reviews

markusylisiurunen:
## why

Large diff reviews need orientation before line-by-line inspection. The previous reviewer brief was isolated from the main workflow, and the browser app had accumulated duplicated controls, state logic, and styling that made UI iteration difficult.

## what

Add a Guide mode that prepares orientation, focused topics, and reviewer questions as soon as shared review context is ready. Reviewers can request or revise explanations, ask additional questions, attach comments to guide content, and switch to the existing Diff mode for file and line-level review.

Reorganize the React app around diff, guide, review, and thread features with shared UI primitives. Canonicalize the palette and control styling, simplify the top bar and diff layout, persist guide state through the versioned review document, and include guide comments in returned reviews.

## details

Guide generation is deduplicated across eager startup and browser requests, and restored reviews recreate agent threads from fresh bootstrap context when follow-up work is requested. The v2 persisted state reader retains support for shipped v1 review documents while validating guide identities, targets, and bounds at the storage boundary.


src/diff_tool/http_server.ts
```diff
@@ -544,37 +692,77 @@ export class DiffToolHttpServer {
...
-      const payload = parseReviewPayload(await this.readJsonBody(request));
+      await this.readJsonBody(request);
       if (this.submissionState !== "active") {
         this.sendJson(response, 409, { error: "diff review has already been submitted" });
         return;
       }
 
       this.submissionState = "submitting";
-      const review = this.reviewState.buildReviewText(payload.message);
+      const review = this.reviewState.buildReviewText();
```
markustherobot, line 766: **[P1] Flush pending mutations before submission**

`/api/guide/comment` publishes its mutation only after the storage save completes, but `/api/review` reads `reviewState` immediately without waiting for `stateMutationQueue`. If a reviewer saves a guide comment and immediately clicks Submit (especially with a slow durable adapter), submission can overtake that save, return a review without the new comment, and close the session. Serialize submission behind pending state mutations, or prevent submission until client writes are acknowledged.

src/diff_tool/app/src/App.tsx
```diff
@@ -71,772 +20,90 @@ const LocalAgentation = import.meta.env.DEV
...
-          onCollapseAll={collapseAll}
+          mode={navigation.mode}
+          fileCount={session.files.length}
+          viewedCount={session.reviewState.viewedFileIds.length}
+          additions={fileState.totals.additions}
+          deletions={fileState.totals.deletions}
+          commentCount={fileState.unresolvedThreadCount}
+          diffStyle={session.reviewState.diffStyle}
+          overflowMode={session.reviewState.overflowMode}
+          finished={submission.finished}
```
markustherobot, line 62: **[P2] Restore a visible status and error surface**

The hooks still write request failures and submission progress to `session.status`, but the new component tree never renders it; `status` only feeds `emptyContent`, which is visible only when Diff mode has no files. With a normal diff or in Guide mode, guide generation, comment persistence, thread, submit, and cancel failures silently appear to do nothing. Render the status in the top bar or the relevant feature panel.

src/diff_tool/review_prompts.ts
```diff
@@ -1,69 +1,265 @@
...
+      return { kind: operation.kind, question: content.question };
+    }
+  }
+}
+
+function buildGuideOperationContext(currentGuide: DiffToolGuide): string[] {
+  return [
+    "Use this current guide as context:",
+    "",
+    JSON.stringify({
```
markustherobot, line 219: **[P2] Escape guide content inside the system wrapper**

`currentGuide` is model-generated from repository content, and this JSON is interpolated inside `<system>...</system>` without escaping `<`. A guide about Tau's own prompt code can legitimately contain `</system>` (and an adversarial diff can induce it), which closes the wrapper early and makes the remaining reference content look like instructions. The comment-thread path already applies `.replaceAll("<", "\\u003c")`; apply the same encoding to guide and queued-operation context before wrapping it.

src/diff_tool/review_prompts.ts
```diff
@@ -1,69 +1,265 @@
...
+        response,
+        z.object({ topic: topicSchema }).strict(),
+        "guide topic",
+      );
+      return { kind: operation.kind, topic: content.topic };
+    }
+    case "question.ask": {
+      const content = parseGuideAgentResponse(
+        response,
+        z.object({ question: questionSchema }).strict(),
```
markustherobot: **[P2] Preserve the reviewer's exact question**

For `question.ask`, the parser accepts whatever question text the model returns even though the prompt requires an exact copy, then persists that text with `source: "user"`. A harmless model paraphrase therefore changes what the reviewer ostensibly asked and what later comments target. Persist `operation.question` and parse only the answer, or reject responses whose question does not match exactly.

src/diff_tool/app/src/features/guide/guide_conversations.tsx
```diff
@@ -0,0 +1,144 @@
...
+          aria-pressed={view === "history"}
+          title="Conversation history"
+          onClick={conversations.showHistory}
+        >
+          <History size={16} />
+        </Button>
+        <div className="guide-conversation-header-actions">
+          {selected && !selected.resolved && (
+            <Button
+              variant="ghost"
```
markustherobot, line 42: **[P2] Let reviewers re-include excluded conversations**

This action is shown only for unresolved conversations. After it calls `resolveThread(..., true)`, history still lets the reviewer open the excluded thread, but the composer becomes only “Excluded from review” and no UI path calls `resolveThread(..., false)`. One accidental click permanently removes that conversation from the submitted review, so provide a matching Include/Reopen action for resolved conversations.

src/diff_tool/app/src/App.css
```diff
@@ -1,190 +1,44 @@
...
-  font: inherit;
+.app.diff-mode {
+  grid-template-columns: var(--sidebar-width) minmax(0, 1fr);
 }
 
-.app {
-  min-height: 100vh;
+.app.guide-mode {
+  --guide-conversation-width: 450px;
+  grid-template-columns:
```
markustherobot, line 15: **[P2] Collapse fixed side panels on narrow windows**

Diff mode always reserves a 320px sidebar and Guide mode always reserves a 450px panel plus padding, while the page disables overflow and defines no responsive breakpoint. On narrow browser windows the primary review content collapses to an unusably small column with no way to hide or scroll past the side panel. Add a narrow-layout breakpoint that stacks, overlays, or hides these panels.

markustherobot (commented):
I found several material issues in the new review workflow, mainly around asynchronous state publication, client feedback, and Guide interaction boundaries.

I launched five focused subreviews covering server concurrency, persistence contracts, Guide client state, Diff client regressions, and protocol/prompt behavior. All five completed, and I verified the retained findings against the current head.

src/diff_tool/app/src/App.tsx
markusylisiurunen: let's remove statuses and errors for now and create a new issue to remember to handle this holistically and in a unified way wrt statuses, errors, working indicators, etc.

src/diff_tool/review_prompts.ts
markusylisiurunen: no need for this kind of overengineering

src/diff_tool/review_prompts.ts
markusylisiurunen: I don't think it makes sense to ask the model to return the question itself at all for user-asked questions when it writes an answer

src/diff_tool/app/src/features/guide/guide_conversations.tsx
markusylisiurunen: ok, could maybe show an eye icon where now the exclude button is shown

markusylisiurunen:
@markustherobot address unresolved findings

src/diff_tool/http_server.ts
markustherobot: I now wait for the state mutation queue before composing and returning the review. The regression test holds a guide-comment persistence write open and verifies submission neither returns nor omits that feedback.

src/diff_tool/app/src/App.tsx
markustherobot: I removed the unrendered session status and error-message plumbing while preserving local loading rollback. #466 now tracks a unified design for statuses, errors, working indicators, completion feedback, and retries.

src/diff_tool/review_prompts.ts
markustherobot: User-question operations now ask the model to return only an answer. The parser reconstructs the persisted question from the original operation text for both single and queued requests.

src/diff_tool/app/src/features/guide/guide_conversations.tsx
markustherobot: Excluded conversations now show an eye action in the same header location, which marks the thread unresolved and includes it in the returned review again.

markusylisiurunen:
@markustherobot deslop

PR #469, opened 2026-08-25
add private history viewer

### LINKED ISSUE #468: read-only UI for history Worker
and the agent should be able to give urls and help find conversations.

markustherobot:
## why

The remote history service already owns stable session identities, cross-host search, generated digests, and ordered transcripts, but those conversations were only available through the agent-facing API. A private browser view makes the user-owned collection directly navigable without exposing session snapshots or adding a local history server.

## what

- add server-rendered, paginated index, search, and transcript pages to the history Worker
- protect browser routes with HTTP Basic auth using a dedicated viewer password, redirect external HTTP viewer traffic before authentication, and preserve bearer auth for the existing JSON API
- expose absolute `webUrl` values on remote session descriptors and teach the history tool to return them when a conversation link is requested
- install the separate viewer credential from `tau history setup`, printing generated secrets without echoing supplied credentials
- add a checked-in Wrangler development project with local-only D1 state, fixed development credentials, and an idempotent fixture
- make one SQL migration asset canonical for deployment and local development
- document the browser view, credentials, URL behavior, and local workflow

## details

The viewer renders remote history semantics rather than resumable session snapshots. Rewinds remove the replicated suffix, compaction keeps earlier transcript entries, replication can lag, and oversized payloads can contain truncation markers. Text, metadata, summaries, snippets, and tool payloads are escaped under restrictive browser security and no-store headers. Valid PNG, JPEG, GIF, and WebP content blocks render as inline `data:` images; other structured content falls back to escaped text. A small authenticated same-origin script provides bounded Markdown copy controls and bulk tool expansion.

The fixed viewer username is `tau`. Existing history service installations need to rerun `tau history setup` to install `VIEWER_PASSWORD`; setup reuses the existing D1 database. The browser credential remains separate from the host API key and never appears in session URLs. External HTTP viewer requests redirect to HTTPS before a Basic challenge or content response, while loopback HTTP remains available for local Wrangler development. Production conversation descriptors use HTTPS URLs. This remains scoped to the remote Worker, with no local-history HTTP service or shareable capability links.

fixes #468


package.json
```diff
@@ -53,7 +54,8 @@
     "@types/ws": "^8.18.1",
     "prettier": "^3.9.6",
     "typescript": "^7.0.2",
-    "vitest": "^4.1.11"
+    "vitest": "^4.1.11",
+    "wrangler": "^4.126.0"
```
markustherobot, line 58: **[P1] Allow Wrangler's required install scripts**

Adding Wrangler pulls in `esbuild@0.28.1` and `workerd@1.20260825.1`, but neither is present in `allowScripts`. Both CI and npm publishing run `npm ci --strict-allow-scripts`, and the current build already fails at that step, so no clean checkout can reach checks or tests. Add these exact resolved packages to the allowlist, or use a dependency graph that does not require the unapproved scripts.

src/core/history/setup.ts
```diff
@@ -77,6 +89,10 @@ export async function setupHistoryService(options: HistorySetupOptions): Promise
       stdout("Set TAU_HISTORY_API_KEY to:");
       stdout(apiKey);
     }
+    stdout("");
+    stdout(`Open the private history viewer at https://${domain}/`);
+    stdout("Sign in with username tau and this viewer password:");
+    stdout(viewerPassword);
```
markustherobot, line 95: **[P1] Do not echo a supplied viewer password**

When `--viewer-password` or `TAU_HISTORY_VIEWER_PASSWORD` supplies the credential, this still writes it to stdout, exposing access to every transcript in CI logs and other output capture. The existing API-key branch deliberately avoids echoing supplied secrets; likewise, print the viewer password only when setup generated it, and otherwise report that the supplied password was installed.

src/history/worker/index.ts
```diff
@@ -118,37 +155,52 @@ function invalidRequest(message: string): HistoryApiError {
...
+        }
 
-      if (new URL(request.url).pathname === "/v1/read") {
-        return json(await read(env.DB, await readJson(request)));
+        return error("not_found", "Not found", 404);
       }
 
-      return error("not_found", "Not found", 404);
+      if (!viewerRoute) return viewerError("Not found", 404);
+      if (!authorizeViewer(request, env)) return viewerUnauthorized();
```
markustherobot, line 188: **[P1] Redirect plaintext requests before Basic auth**

History endpoints explicitly permit `http://`, and the deployed Worker route does not configure an HTTPS redirect, so this branch can issue a Basic challenge and serve transcripts over plaintext HTTP. A user who opens that URL and authenticates sends the viewer password unencrypted. Reject or redirect non-HTTPS external viewer requests before challenging (while preserving loopback development), and ensure emitted production `webUrl` values are HTTPS.

markustherobot (commented):
Found three blocking issues across transport security, credential handling, and clean-install policy. I launched three Luna subreviews covering Worker viewer security, setup/packaging/migrations, and web URL/client integration; all completed, and I independently verified their candidates against the current head.

markusylisiurunen:
@markustherobot ci fails

markusylisiurunen:
@markustherobot review

src/history/worker/index.ts
```diff
@@ -118,37 +155,60 @@ function invalidRequest(message: string): HistoryApiError {
...
```
markustherobot, line 166: **[P1] Browser credentials and transcripts can travel over plaintext HTTP**

The viewer challenges and serves requests without checking their scheme. Calling the Worker at `http://history.example.com/` returns a Basic-auth 401, then a 200 with valid credentials. Setup installs a scheme-less `${domain}/*` route and does not configure an HTTPS redirect, so the documented HTTPS-only boundary depends on an unrelated zone setting. Opening an HTTP link can expose the collection-wide viewer password and transcript to the network; descriptors also preserve an HTTP request origin in `webUrl`.

**Suggested fix:** Redirect or reject external HTTP viewer requests before issuing an authentication challenge or serving content, retain an explicit loopback exception for local development, and emit HTTPS production conversation URLs. Cover the unauthenticated and authenticated HTTP paths in tests.

src/history/worker/index.ts
```diff
@@ -661,13 +757,395 @@ function descriptor(row: Record<string, unknown>): Record<string, unknown> {
...
+    .replaceAll("<", "&lt;")
+    .replaceAll(">", "&gt;")
+    .replaceAll('"', "&quot;")
+    .replaceAll("'", "&#39;");
+}
+
+const VIEWER_JS = `
+function inline(value) {
+  const tick = String.fromCharCode(96);
+  return value.trim().replace(/([\\\\*_{}[]<>#+.!|~-])/g, "\\$1").replaceAll(tick, "\\" + tick);
```
markustherobot: **[P1] Every copy and bulk tool-toggle button is nonfunctional**

`VIEWER_JS` is an ordinary template literal, so constructing the response consumes the escapes intended for the browser's JavaScript. For example, the emitted `replaceAll` string is unterminated, and the `"\n"` sequences later in the template become literal newlines inside quoted JavaScript strings. Fetching `/viewer.js` and compiling its body produces `SyntaxError: Invalid or unexpected token`. The browser therefore never installs the click listener, breaking every Copy, Copy conversation, Expand tools, and Collapse tools button despite the successful script-response test.

**Suggested fix:** Preserve valid JavaScript escapes in the emitted asset (for example, use a raw template with correctly authored script contents), then test parsing and executing the actual response, including clipboard conversion and the tool toggles.

src/core/history/setup.ts
markustherobot: **[P2] Setup exposes caller-supplied viewer passwords in captured output**

When `--viewer-password` or `TAU_HISTORY_VIEWER_PASSWORD` supplies the credential, these lines still print it unconditionally. An operator using a protected environment or secret manager therefore gets the collection-wide transcript password copied into setup logs or other captured stdout. The adjacent API-key branch deliberately avoids echoing supplied secrets, while the new test currently asserts that this password is printed.

**Suggested fix:** Mirror the API-key branch: print a viewer password only when setup generated it, otherwise report that the supplied credential was installed. Update the test to assert that a supplied password is absent from output.

markustherobot (commented):
Found three issues affecting browser credential protection and working viewer controls. The focused history suite passes (49 tests), but compiling the actual `/viewer.js` response exposes a failure the tests miss.

I launched three Luna subreviews covering viewer/browser behavior, setup and packaging, and remote history contracts. All completed; I independently checked their candidates against the complete current snapshot.

markusylisiurunen:
@markustherobot can you address the unresolved review comments? also run a deslop pass using the `deslop` skill.

src/history/worker/index.ts
markustherobot: Fixed in 9792b50. External HTTP viewer routes now redirect to HTTPS before authentication or content, loopback HTTP remains available for Wrangler, and production `webUrl` values are normalized to HTTPS. Tests cover unauthenticated and authenticated HTTP requests plus the loopback exception.

src/history/worker/index.ts
markustherobot: Fixed in 9792b50. The asset now uses `String.raw`, and the test executes the actual `/viewer.js` response before exercising message, tool, and conversation copying plus both bulk tool toggles.

src/core/history/setup.ts
markustherobot: Fixed in 9792b50. Setup now prints generated viewer passwords only; supplied CLI or environment credentials are installed without being echoed, and the setup test asserts the supplied value is absent from captured output.

PR #470, opened 2026-08-27
migrate Gemini speech transcription

markusylisiurunen:
## why

Gemini now has dedicated transcription models for live and file-based speech recognition. Tau's Gemini provider still used the general-purpose Flash model in a single prompted request, while the unused Mistral speech provider kept an unnecessary third configuration path.

## what

- stream terminal microphone audio to `gemini-3.5-transcribe-live` in smart mode
- derive bounded custom vocabulary from recent conversation context with Gemini 3.7 Flash
- use Files and Interactions APIs with `gemini-3.5-transcribe` for retained recordings and Telegram audio, including best-effort remote cleanup
- remove Mistral speech-to-text while retaining its unrelated PDF OCR support
- update speech configuration, documentation, and coverage for the two-provider contract


src/core/utils/gemini_transcription.ts
```diff
@@ -47,131 +94,544 @@ export type GeminiTranscriptionOptions = {
...
+        this.sendAudio(audio);
+      }
+      if (this.failure) return;
+      this.resolveReady?.();
+      this.resolveReady = undefined;
+      this.rejectReady = undefined;
+      return;
+    }
+
+    const transcript = event.data.serverContent?.inputTranscription?.text;
```
markustherobot, line 304: **[P2] Preserve interim text when no final frame arrives**

Gemini Live sends `interimInputTranscription` snapshots separately from the final `inputTranscription`, and a turn can complete with only the latest interim available. This handler ignores both the interim field and `generationComplete`/`turnComplete`, so that case waits for the 30-second timeout and reports a failed transcription despite having usable text. Track the latest interim and promote it on turn completion when no final transcript was received.

markustherobot (commented):
I launched three focused subreviews covering the Gemini Live state machine, the batch upload lifecycle, and the config/TUI/Telegram integrations. The main API shapes and provider wiring are consistent, but one live-completion reliability issue remains.

src/core/utils/gemini_transcription.ts
markusylisiurunen: I don't think we should have this kind of fallbacks, it will just hide issues in our implementation. I believe the Gemini API protocol promises the final response, right? we should assume it works as advertised.

markusylisiurunen:
@markustherobot address the unresolved review comments

src/core/utils/gemini_transcription.ts
markustherobot: Agreed. I rechecked the protocol: `interimInputTranscription` is a low-latency update, while `inputTranscription` is the final transcription and is independently ordered from turn completion. Promoting interim text would mask a missing final event, so I’m keeping the final-only behavior and the existing timeout as the failure path.

PR #473, opened 2026-08-28
raise diff review snapshot limit

### LINKED ISSUE #472: Diff review fails when patch exceeds hardcoded 2 MiB capture limit
## What happened

An interactive diff review failed to start for `markusylisiurunen/pico-labs#234` at head `9aa11cefd6d16ede5a181fdeee918a92a82239ee`.

Tau reported:

```text
git diff ad0c6833df3115d91d8303d5fda6cda19922b382...9aa11cefd6d16ede5a181fdeee918a92a82239ee output exceeded 2097152 bytes while capturing diff review snapshot
```

The generated diff is 2,109,009 bytes, only 11,857 bytes above the hardcoded 2 MiB `GIT_MAX_CAPTURE_BYTES` limit in `src/core/diff_review/snapshot.ts`.

## Reproduction

With the two commits available locally:

```bash
git diff ad0c6833df3115d91d8303d5fda6cda19922b382...9aa11cefd6d16ede5a181fdeee918a92a82239ee | wc -c
# 2109009
```

Starting a diff review for the same range fails while capturing the snapshot.

## Expected behavior

Diff reviews should support reasonably large repository changes without failing because the complete patch exceeds a small process-output capture limit. Possible approaches include streaming or writing the patch to a temporary file, or using a separate configurable snapshot-size limit rather than the generic Git command capture limit.

If a size limit is still required, the SDK should expose a structured/actionable failure so integrations can tell users that the diff is too large and report the applicable limit.

## Environment

- Tau: `0.3.64`
- Node.js: `24.19.0`
- Invocation: Tau SDK through Cowork running as a systemd service

markustherobot:
## why

Diff review snapshot capture used a generic 2 MiB Git output ceiling, which rejects otherwise reviewable repository changes such as the 2,109,009-byte patch reported in #472.

## what

- raise the diff-review snapshot patch limit to 16 MiB for Git diffs and patch-file sources
- enforce the limit again on the final assembled snapshot so combined tracked and untracked patches cannot bypass it
- raise the per-file untracked text limit to 4 MiB while retaining the aggregate snapshot backstop
- return an actionable size-limit message and document the limits for TUI and SDK reviews
- cover explicit diffs above the old ceiling, aggregate working-tree enforcement, and patch-file rejection

## details

The limit is measured in UTF-8 bytes and permits a patch exactly at the ceiling. Patch-file input is bounded cumulatively while reading and the canonical assembled patch is checked before constructing the snapshot.

The existing 30-second Git timeout, 24 MiB SDK execution-capture ceiling, diff-tool request-body limit, and persisted review-state limit remain unchanged because they protect separate paths. There are no remaining implementation questions.

fixes #472

### DIFF
--- a/src/core/diff_review/snapshot.ts
+++ b/src/core/diff_review/snapshot.ts
@@ -57,8 +57,8 @@ type DiffSnapshotDeps = Pick<CoreDeps, "spawn"> & {
 };
 
 const GIT_TIMEOUT_MS = 30_000;
-const GIT_MAX_CAPTURE_BYTES = 2 * 1024 * 1024;
-const MAX_UNTRACKED_TEXT_BYTES = 512 * 1024;
+const MAX_SNAPSHOT_PATCH_BYTES = 16 * 1024 * 1024;
+const MAX_UNTRACKED_TEXT_BYTES = 4 * 1024 * 1024;
 const WORKING_TREE_SCOPE_LABEL = "current working tree";
 
 export class DiffReviewSnapshot {
@@ -155,6 +155,7 @@ export async function captureDiffReviewSnapshot(
       : source.diffArgs.length > 0
         ? await captureExplicitDiffSnapshot(cwd, source.diffArgs, deps, signal)
         : await captureWorkingTreeSnapshot(repoRoot, deps, signal);
+  assertSnapshotPatchBytesWithinLimit(Buffer.byteLength(captured.patch, "utf-8"));
   const diffArgs = source.kind === "git_diff" ? [...source.diffArgs] : [];
 
   return new DiffReviewSnapshot({
@@ -178,6 +179,14 @@ type CapturedSnapshotData = {
 function throwIfSnapshotAborted(signal?: AbortSignal): void {
   if (signal?.aborted) {
     throw new Error("diff review start aborted");
+  }
+}
+
+function assertSnapshotPatchBytesWithinLimit(bytes: number): void {
+  if (bytes > MAX_SNAPSHOT_PATCH_BYTES) {
+    throw new Error(
+      `diff review snapshot patch exceeds the ${MAX_SNAPSHOT_PATCH_BYTES}-byte limit; narrow the review scope`,
+    );
   }
 }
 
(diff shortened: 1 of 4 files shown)


src/core/diff_review/snapshot.ts
```diff
@@ -57,8 +57,8 @@ type DiffSnapshotDeps = Pick<CoreDeps, "spawn"> & {
 };
 
 const GIT_TIMEOUT_MS = 30_000;
-const GIT_MAX_CAPTURE_BYTES = 2 * 1024 * 1024;
-const MAX_UNTRACKED_TEXT_BYTES = 512 * 1024;
+const MAX_SNAPSHOT_PATCH_BYTES = 16 * 1024 * 1024;
+const MAX_UNTRACKED_TEXT_BYTES = 4 * 1024 * 1024;
```
markustherobot, line 61: **[P2] Avoid argument spreading for newly accepted files**

Raising this limit newly sends 512 KiB–4 MiB text files through `createUntrackedFilePatch`, which calls `patch.push(...lines.map(...))`. On Node 24, a 1 MiB file made of `a,b,c,d\n` creates 131,072 arguments and throws `RangeError: Maximum call stack size exceeded`, so plain `/diff` now fails on a plausible CSV or log that was previously skipped. Append lines with a loop, or construct the patch without an argument spread, so files up to the advertised limit are actually supported.

src/core/diff_review/snapshot.ts
```diff
@@ -155,6 +155,7 @@ export async function captureDiffReviewSnapshot(
       : source.diffArgs.length > 0
         ? await captureExplicitDiffSnapshot(cwd, source.diffArgs, deps, signal)
         : await captureWorkingTreeSnapshot(repoRoot, deps, signal);
+  assertSnapshotPatchBytesWithinLimit(Buffer.byteLength(captured.patch, "utf-8"));
```
markustherobot, line 158: **[P2] Enforce the aggregate limit while collecting untracked patches**

The only aggregate check runs after `captureWorkingTreeSnapshot` returns, but `captureUntrackedFilePatches` has already read, synthesized, and retained every eligible patch in its array and map. With the new 4 MiB per-file allowance, a worktree containing many unignored generated text files can consume hundreds of MiB or OOM before it reaches the intended 16 MiB rejection. Pass the tracked-byte count or remaining budget into untracked capture and reject before retaining a patch that crosses it; keep this final assertion as defense in depth.

markustherobot (commented):
I found two reliability issues in the expanded untracked-file path that should be addressed. I launched three read-only subreviews covering snapshot capture boundaries, downstream transport and limit compatibility, and tests and resource behavior; all completed.

PR #475, opened 2026-08-29
preview diff review submissions

### LINKED ISSUE #471: a way to preview diff tool submission
- right now it's hard to see what comments are sent at submission.
- I'd like to be able to preview them before submitting.
- would be great to be able to perhaps "resolve" or remove comments from this preview too if something does not feel right. editing?
- I'd also like to be able to just copy the full submission text from the UI without having to submit.

markustherobot:
## why

Diff review feedback was submitted immediately, which made it difficult to inspect the final text or copy it before returning the review to Tau.

## what

Submit and Approve now open a Base UI dialog with the server-generated review preview. The dialog renders the exact canonical Markdown, provides a copy action, and shows an explicit approval state when no comments remain.

Included feedback can be excluded directly from the preview. Review threads are resolved, guide comments are removed through their existing state operation, and the preview is regenerated after each change. The mock server, public behavior documentation, and server regressions cover the same flow.

## details

The final Markdown remains read-only so review state has one canonical representation. Preview and submission use the same server-side formatter rather than assembling text in the browser. No inline message editing was added, following the issue discussion.

fixes #471


src/diff_tool/http_server.ts
```diff
@@ -766,7 +772,7 @@ export class DiffToolHttpServer {
       }
 
       this.submissionState = "submitting";
-      const submission = this.reviewState.buildReviewSubmission();
+      const submission = this.reviewState.buildReviewPreview().submission;
```
markustherobot: **[P1] Bind submission to the previewed state**

`GET /api/review` returns a point-in-time preview, but this line rebuilds the submission from mutable state. For example, `/api/thread-message` waits for the model outside `stateMutationQueue`, so its agent reply can commit after the dialog opens but before Submit is clicked; Tau then receives generated text that was neither shown nor copied. Return a state revision or preview token and reject/refresh a stale preview, or submit an immutable server-held preview snapshot.

src/diff_tool/app/src/features/review/use_review_submission.ts
```diff
@@ -1,15 +1,100 @@
...
+  );
+
+  const copy = useCallback(async () => {
+    if (preview?.submission.outcome !== "commented") {
+      return;
+    }
+
+    setError(null);
+    try {
+      await navigator.clipboard.writeText(preview.submission.review);
```
markustherobot: **[P2] Support copy on non-loopback HTTP origins**

The SDK explicitly supports binding the UI to a non-loopback `host`, and the server returns an `http://` URL for that documented trusted-network setup. Those origins are not secure contexts, so `navigator.clipboard` is unavailable and the new Copy action always errors there. Detect this case and provide a manual/selectable-text or copy fallback, or otherwise constrain the supported setup so the documented action works.

markustherobot (commented):
I found two material issues in the confirmation flow. I launched three focused subreviews covering server/state consistency, dialog async behavior, and contracts/mock/tests; all completed successfully.

PR #479, opened 2026-08-30
align voice transcription duration limits

### LINKED ISSUE #476: is there a max length in voice input duration?

markustherobot:
## why

The 20-minute voice-input limit is a product policy for Tau's clients, not a technical ceiling in the transcription provider interface. Provider adapters should accept long audio and hide provider-specific constraints from callers.

## what

- Enforced the 20-minute file policy at the TUI retry and Telegram boundaries instead of in the shared provider facade.
- Made the TUI stop and finish live transcription at 20 minutes without submitting the resulting editor contents.
- Removed duration metadata and artificial 20- and 30-minute guards from the shared interface and both provider adapters.
- Split decoded OpenAI file audio into independent WAV uploads below the provider's 25 MB file limit, with up to four requests in flight and results restored to audio order.
- Bounded OpenAI decoding with a progress-scaled watchdog that preserves support for arbitrarily long audio.
- Updated TUI and Telegram documentation and added boundary regressions.

## details

Telegram's required integer-second duration remains authoritative and oversized messages are rejected before download. The TUI retains measured duration with failed recordings, rejects an oversized retry before reading or provider work, and uses the same finish-and-insert behavior for manual stops and the 20-minute timer.

Gemini uses its native upload and realtime paths without a Tau-imposed duration ceiling. OpenAI decodes completed audio once, creates upload-safe chunks internally, and transcribes independent chunks through a four-worker pool using only the original conversation prompt. Results are joined by chunk index, and the first failure cancels sibling requests and prevents additional chunks from being scheduled. Streaming adapters likewise no longer count audio toward an artificial provider cap.

OpenAI decoding receives a 60-second base execution budget plus one second for every second of PCM produced. The deadline therefore scales with legitimate long input while stalled or slower-than-realtime decoding is cancelled with a distinct timeout error.

Exactly 20 minutes remains accepted at client file boundaries. Reaching the live TUI limit stops further recording, finishes transcription, and leaves the transcript in the editor for review without submitting it.

fixes #476


src/core/utils/openai_transcription.ts
```diff
@@ -24,10 +24,13 @@ const OPENAI_TRANSCRIPTION_COMPLETION_TIMEOUT_MS = 30_000;
 const OPENAI_TRANSCRIPTION_KEYWORD_TIMEOUT_MS = 15_000;
 const OPENAI_TRANSCRIPTION_MAX_KEYWORD_CHARACTERS_TOTAL = 1_024;
 const OPENAI_TRANSCRIPTION_CONTEXT_TOKENS = 1_024;
+const OPENAI_TRANSCRIPTION_MAX_DURATION_MS = 30 * 60 * 1_000;
 const OPENAI_TRANSCRIPTION_FFMPEG_TIMEOUT_MS = 5 * 60 * 1_000;
 const OPENAI_TRANSCRIPTION_FFMPEG_OUTPUT_LIMIT_BYTES = 20_000;
-const OPENAI_TRANSCRIPTION_MAX_PCM_BYTES =
-  OPENAI_FILE_SAMPLE_RATE * 2 * (OPENAI_TRANSCRIPTION_FFMPEG_TIMEOUT_MS / 1_000);
+const OPENAI_TRANSCRIPTION_MAX_FILE_PCM_BYTES =
+  OPENAI_FILE_SAMPLE_RATE * 2 * (OPENAI_TRANSCRIPTION_MAX_DURATION_MS / 1_000);
```
markustherobot: **[P1] OpenAI rejects accepted recordings longer than roughly 13 minutes**

This raises the decoded allowance to 57,600,000 PCM bytes, and `transcribeOpenAIAudio` wraps all decoded bytes in an uncompressed WAV before uploading it. OpenAI's [file transcription contract](https://developers.openai.com/api/docs/guides/speech-to-text#longer-inputs) caps files at 25 MB, so an accepted 20-minute recording becomes about 38.4 MB and is rejected by the endpoint; the default Telegram/OpenAI path and `/listen retry` therefore still cannot handle much of the new supported range.

**Suggested fix:** Keep the decoded PCM accounting for actual-duration validation, but encode the upload into a supported compressed format below 25 MB or split it into bounded requests before calling the transcription endpoint.

src/tui/session_chat_controller.ts
```diff
@@ -894,7 +895,7 @@ export class SessionChatController {
       recording.maxDurationTimeout = setTimeout(() => {
         if (this.listenRecording !== recording || this.listenTransition) return;
         void this.runListenTransition(() => this.stopListenCapture());
-      }, LISTEN_RECORDING_MAX_DURATION_MS);
+      }, SPEECH_TO_TEXT_CLIENT_MAX_DURATION_MS);
```
markustherobot, line 898: **[P1] Automatic 20-minute recordings fail instead of transcribing**

`startListenAudioCapture` forwards the first PCM chunk before `capture.started` resolves, but this full 20-minute timer is armed only after that await (and after retained-file cleanup). The shared streaming guard has therefore already counted audio when the timer starts, so a later chunk crosses its 20-minute byte ceiling and aborts the provider before this callback runs. The normal stop path then calls `finish()`, gets the limit error, and retains the recording as a failed transcription instead of inserting the transcript.

**Suggested fix:** Establish the auto-stop deadline at the first PCM chunk before forwarding it, or subtract already-streamed PCM duration when arming the timer so capture stops before the guard-crossing chunk. Add an auto-stop boundary regression that asserts successful transcription rather than retention.

markustherobot (commented):
Reviewed the complete 13-file snapshot. Two blocking duration-boundary regressions remain around provider upload sizing and TUI auto-stop coordination.

I launched three subreviews covering provider/shared speech limits, TUI capture and retry state, and Telegram duration handling, then independently verified their candidates against the source and tests. The focused affected suites pass and the pull request check is green.

markusylisiurunen:
@markustherobot I actually think we need to redefine the limits a bit. I want each of the transcription providers (OpenAI and Gemini) to allow effectively however long audio to be transcribed. the 20-minute limit should be a product choice instead of a technical limitation. the adapters may need to chunk the audio somehow when it goes past the limit, but that should be an internal implementation detail when needed. if the transcription provider interface API needs to evolve to support this cleanly, you should do that.

- for realtime transcription in TUI, should just trigger submit if the 20 minutes is hit.
- for files (e.g. /listen retry or Telegram), should return an error for audio over 20 mins.

markusylisiurunen:
@markustherobot review

src/core/telegram/adapter.ts
```diff
@@ -2708,6 +2721,14 @@ class TelegramAdapterImpl {
     message: TelegramAudioMessage,
     options: { silent?: boolean } = {},
   ): Promise<TelegramAudioTranscriptionResult> {
+    if (message.durationMs > SPEECH_TO_TEXT_CLIENT_MAX_DURATION_MS) {
```
markustherobot, line 2724: **[P1] Sender metadata lets oversized Telegram audio bypass the limit**

This check treats `voice.duration` and `audio.duration` as the actual media length, but the [Telegram Bot API](https://core.telegram.org/bots/api#audio) defines this field as “duration ... as defined by the sender.” A sender can therefore declare at most 1,200 seconds for a longer file and Tau will download and transcribe it. That bypasses the requested client policy and is especially risky on the OpenAI path now that decoding has no PCM cap: an allowed group participant can force substantial decoding memory, provider work, and cost even when `allowedUserIds` would prevent them from triggering a turn, because group audio is transcribed before that authorization check.

**Suggested fix:** Keep this metadata check as an early rejection, but after download independently derive and validate the actual media duration at the Telegram boundary before invoking either provider. Bound and cancel that probe/decoding work so a malformed file cannot become a resource-exhaustion path; keep the shared provider interface duration-agnostic.

src/core/utils/openai_transcription.ts
```diff
@@ -460,20 +476,14 @@ async function decodeOpenAIAudio(args: {
       ],
       {
         input: args.audio,
-        signal,
-        timeoutMs: OPENAI_TRANSCRIPTION_FFMPEG_TIMEOUT_MS,
+        signal: args.signal,
```
markustherobot: **[P2] A stalled decoder can block transcription indefinitely**

The decoder now receives only the lifecycle abort signal and no execution timeout, and the `timedOut` handling was removed with it. During ordinary TUI or Telegram transcription that signal is not aborted unless the client or runner shuts down, so a malformed or pathologically slow file can leave the speech operation (and Telegram’s per-chat update queue) stuck indefinitely. Removing the decoded-duration ceiling does not require removing a processing deadline.

**Suggested fix:** Restore a separate, generous ffmpeg execution timeout, ideally scaled from the accepted or probed media duration, and retain a distinct timeout error. This preserves support for long recordings without allowing a decoder process to run forever.

markustherobot (commented):
Reviewed the complete 10-file snapshot. Two material reliability issues remain at the external duration-validation and decoder-lifecycle boundaries.

I launched three subreviews covering provider/shared transcription, TUI recording and retry behavior, and Telegram audio validation, then independently verified their candidates against the source and tests. The focused 179-test affected suite passes, and the pull request check is green.

src/core/telegram/adapter.ts
markusylisiurunen: I think it's ok to trust this.

markusylisiurunen:
@markustherobot address unresolved review comments

src/core/utils/openai_transcription.ts
markustherobot: Implemented in `723f7f4`. OpenAI decoding now uses a combined owner/watchdog abort signal with a 60-second base budget plus one second per decoded second. This keeps the deadline proportional for arbitrarily long inputs while aborting stalled or slower-than-realtime decoding with a distinct timeout error. I added a progress-and-timeout regression, and the full test suite passes.

markusylisiurunen:
@markustherobot 

> but stalled or slower-than-realtime work is cancelled

what if the model or endpoint just simply is a bit slow while still working as intended? I think this kind of failure guards should always be quite lax so that they never fire under normal operations, even accidentally. they are meant for real failure modes. perhaps I'm misunderstanding something, though. no changes, just answer me.

PR #482, opened 2026-09-02
evict unobserved session runtimes

markusylisiurunen:
## why

Tau Serve retained every live session runtime after its WebSocket client disconnected. Durable session snapshots remained useful for recovery, but keeping histories, runtime state, prompts, and execution environments resident indefinitely caused memory use to grow with every Cowork workflow.

## what

Track observer references for hosted sessions and release them when protocol connections close. After the final observer disconnects, idle runtimes are snapshotted and disposed while their durable session remains recoverable. Active detached work is allowed to settle before eviction, and recovery races and multiple observers preserve the runtime until it is genuinely unused.

The remote-session documentation and host and protocol regression coverage now reflect these lifecycle semantics.

### DIFF
--- a/src/host/local_session_host.ts
+++ b/src/host/local_session_host.ts
@@ -162,9 +162,11 @@ export type LocalHostedSession = TauHostedSession & {
 
 export class LocalSessionHost implements TauSessionHost {
   private readonly sessions = new Set<LocalHostedSessionHandle>();
+  private readonly sessionReferences = new Map<LocalHostedSessionHandle, number>();
+  private readonly sessionEvictions = new Set<Promise<void>>();
   private readonly sessionRecoveryPromises = new Map<
     string,
-    Promise<LocalHostedSession | undefined>
+    Promise<LocalHostedSessionHandle | undefined>
   >();
   private readonly store: SessionStore;
   private readonly history: HistoryManager;
@@ -232,13 +234,13 @@ export class LocalSessionHost implements TauSessionHost {
       throw new Error("local session host is shut down");
     }
 
-    return hostedSession;
+    return this.retainSession(hostedSession);
   }
 
   private async createRecoveredSession(
     snapshot: SessionProtocolSnapshot,
     executionEnvironment: ExecutionEnvironment,
-  ): Promise<LocalHostedSession> {
+  ): Promise<LocalHostedSessionHandle> {
     let hostedSession: LocalHostedSessionHandle | undefined;
     try {
       const recovered = normalizeRecoveredSnapshot(snapshot);
@@ -371,7 +373,10 @@ export class LocalSessionHost implements TauSessionHost {
       committedSnapshot,
       forceNextSnapshotRevision,
       this.sessionOptions.recordUsage ?? appendUsageLogEntry,
-      (session) => this.sessions.delete(session),
+      (session) => {
+        this.sessions.delete(session);
+        this.sessionReferences.delete(session);
+      },
     );
     const historyFailure = this.history.registerSession(
       {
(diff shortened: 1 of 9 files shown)


markusylisiurunen:
@markustherobot review

markusylisiurunen:
@markustherobot diff

src/host/session_protocol_handler.ts
```diff
@@ -1658,7 +1664,9 @@ export class SessionProtocolHandler {
     }
 
     const session = await this.host.observeSession(sessionId);
-    if (!session || this.closed) {
+    if (!session) return undefined;
+    if (this.closed) {
+      this.host.releaseSession?.(session);
       return undefined;
     }
     return this.registerSession(session);
```
markustherobot, line 1690: **[P1] Concurrent requests can permanently pin a released runtime**

Each `host.observeSession()` call now acquires a retained handle. If two first requests for the same session overlap before either lookup registers state, both calls retain the same handle, but the second `registerSession()` returns the existing state and drops its separately acquired retain. Closing or unobserving the connection then releases only once, leaving the host's reference count positive and preventing the runtime from ever being evicted. The WebSocket server explicitly runs request handlers concurrently, so a client can trigger this with two ordinary session requests.

**Suggested fix:** After the asynchronous lookup, release the newly acquired handle when registration finds an existing state, or serialize lookup and registration so every retain has exactly one matching release.

src/host/session_protocol_handler.ts
```diff
@@ -767,7 +770,7 @@ export class SessionProtocolHandler {
   private async handleAttach(
     request: Extract<SessionProtocolRequestMessage, { method: "session.observe" }>,
   ): Promise<void> {
-    const wasObserved = this.sessionStates.has(request.params.sessionId);
+    const wasObserved = this.sessionStates.get(request.params.sessionId)?.observed ?? false;
```
markustherobot, line 790: **[P2] A failed overlapping observe can detach a successful observer**

`wasObserved` is captured before the asynchronous snapshot completes. If two `session.observe` requests overlap while the state is not yet marked observed, both capture `false`; after one succeeds and sets `state.observed = true`, a snapshot failure in the other still enters this cleanup branch, removes the shared listeners, and releases the host handle. The client therefore receives a successful observe response but stops receiving updates, and the runtime may be evicted underneath that successful observation.

**Suggested fix:** Base failure cleanup on the current shared state's observation or handshake state, and remove it only when no overlapping observe has succeeded or remains active.

src/host/local_session_host.ts
```diff
@@ -522,6 +529,45 @@ export class LocalSessionHost implements TauSessionHost {
...
+    const eviction = this.evictReleasedSession(hostedSession);
+    this.sessionEvictions.add(eviction);
+    void eviction.then(
+      () => this.sessionEvictions.delete(eviction),
+      () => this.sessionEvictions.delete(eviction),
+    );
+  }
+
+  private async evictReleasedSession(session: LocalHostedSessionHandle): Promise<void> {
+    await session.waitForActiveWork().catch(() => undefined);
```
markustherobot, line 558: **[P1] Disconnecting can interrupt an active ephemeral submission**

`LocalHostedSessionHandle.waitForActiveWork()` does not include promises from `submitEphemeralThread()`, which run independently inside `HostedEphemeralAgentSession`. When the final observer disconnects during `session.ephemeral.submit`, this wait can finish immediately, after which `session.dispose()` disposes the ephemeral context and `AgentRuntime.dispose()` interrupts the live thread. This makes a normal detach cancel host work even though the new lifecycle contract says active detached work is allowed to settle before eviction.

**Suggested fix:** Include ephemeral submissions in the hosted session's active-work accounting, or hold an operation lease for each in-flight ephemeral request until it settles before allowing eviction.

markustherobot (commented):
I found three lifecycle issues in the new retention and eviction path, centered on concurrent protocol ownership and detached work that is outside the host's active-work accounting. I launched three read-only subreviews covering host eviction concurrency, protocol reference ownership, and lifecycle regression coverage; all completed successfully.

markusylisiurunen:
@markustherobot address the review comments. think about whether these issues should be patched directly or whether they warrant an architecture or design change in order to address in a proper and clean way.

src/host/session_protocol_handler.ts
markustherobot: Fixed in 36234ba. Protocol registration now releases every redundant handle acquired by a concurrent lookup, and release is a required host contract. The regression test overlaps two first lookups and verifies both acquisitions are balanced.

src/host/session_protocol_handler.ts
markustherobot: Fixed in 36234ba. Observe snapshot handshakes are now serialized per session because they share buffering and listener state. A regression overlaps a failing observe with a successful one and verifies the successful observer keeps receiving deltas.

src/host/local_session_host.ts
markustherobot: Fixed in 36234ba. Ephemeral submissions now run through the hosted session active-work tracker and propagate interruption to the ephemeral AgentRuntime. Eviction therefore waits for detached ephemeral work to settle, with a real-runtime regression covering the lifecycle.

Issue #481, opened 2026-09-01
long-lived TUI sessions become increasingly sluggish


markusylisiurunen:
## problem

A TUI session becomes progressively less responsive as its visible transcript grows. Typing can lag on every key, controls such as reasoning selection take longer, and the slowdown is worse while the model is working. Starting a new TUI session restores normal performance, but compacting the current session does not.

## expected behavior

Typing and controls should stay responsive during long workflows. Compaction should also bound the amount of old conversation that the live UI must process, even when the earlier transcript remains available in terminal scrollback.

## investigation

This was reproduced with focused benchmarks against Tau 0.3.67 after a clean build. The benchmarks used a no-op terminal writer, so they measure application CPU work and exclude terminal output time. Each confirmed cause is recorded in a separate comment below.

Related architecture work: #406.

markusylisiurunen:
## finding 1: every key processes the full retained transcript

Every keyboard event asks pi-tui to render immediately. Tau caches the transcript's rendered lines while it is idle, but the main-screen renderer still copies, normalizes, compares, and scans every retained line before it can update the editor at the bottom.

With terminal writes disabled, a single editor-only update took:

- 10,000 retained lines: about 5 ms median
- 25,000 lines: about 13 ms
- 50,000 lines: about 25 ms, with a 40 ms p95

This is a lower bound because it excludes terminal output. It directly explains why typing gets slower even while the model is idle.

Relevant code: `src/tui/chat_view.ts:136`, pi-tui `tui.ts:235-243`, `tui.ts:890-900`, and `tui-main-screen.ts:262-273,361-387` at the installed 0.84.4 commit.

markusylisiurunen:
## finding 2: compaction does not release old UI history

Compaction starts a new timeline segment, but it does not clear old messages from `ChatContainerComponent`. It adds a “compacted context” divider and keeps all previous message models, components, and rendered lines alive.

This means compaction shrinks the model/session context but not the work done by the live TUI. Repeated compactions keep increasing the retained scrollback. In a synthetic example with roughly 480 characters per message, 10,000 retained messages produced about 70,000 rendered lines and kept about 40 MB of additional heap alive.

The earlier transcript can remain visible in terminal scrollback without remaining in the live component tree. Detaching or bounding frozen segments would let compaction recover UI performance.

Relevant code: `src/tui/session_chat_controller.ts:1793-1818` and `src/tui/ui/chat_container.ts:20,123-127`.

[...]

PR #484, opened 2026-09-02
target tui timeline updates

markusylisiurunen:
## why

Long-lived TUI sessions still rebuild every active transcript entry for common turn transitions. User messages, assistant completion, durable notices, and tool results can therefore pause in proportion to the retained timeline even though each patch changes only a small number of entries.

This continues the low-risk performance work from #481 without changing transcript retention, compaction, or rendering behavior.

## what

The session controller now applies recognized message and timeline patch combinations through a guarded targeted path. It updates only affected messages, derived outcome notices, tool cards, and newly appended notices, while preserving full reconciliation for resets, rewinds, removals, and unfamiliar patch shapes.

Timeline high-water updates also use the presentation-neutral state path. Regression coverage exercises user append, assistant finalization, durable notice, and tool-result patches against a 1,000-message transcript and verifies that unrelated entries are not updated.

### DIFF
--- a/src/tui/session_chat_controller.ts
+++ b/src/tui/session_chat_controller.ts
@@ -1362,7 +1362,7 @@ export class SessionChatController {
       if (this.tryApplyFastContentAppendDelta(delta)) {
         return true;
       }
-      if (this.tryApplyFastToolUiDelta(delta)) {
+      if (this.tryApplyFastTimelineDelta(delta)) {
         return true;
       }
       if (this.tryApplyFastAgentDelta(delta)) {
@@ -1406,6 +1406,7 @@ export class SessionChatController {
           case "cost.set":
           case "settings.set":
           case "turn.set":
+          case "timeline.advance":
           case "operation.set":
           case "operation.remove":
             return false;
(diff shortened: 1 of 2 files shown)


src/tui/session_chat_controller.ts
```diff
@@ -1466,6 +1467,119 @@ export class SessionChatController {
...
+
+    if (!hasTimelineChange) {
+      return false;
+    }
+
+    const nextSnapshot = applySessionProtocolDelta(previousSnapshot, delta);
+    this.snapshot = nextSnapshot;
+    this.observedSessionRevision = Math.max(this.observedSessionRevision, nextSnapshot.revision);
+
+    for (const messageId of affectedMessageIds) {
```
markustherobot, line 1538: **[P2] Preserve timeline order for targeted inserts**

The fast path processes affected messages first, then tools, then notices, while `upsertRenderedMessage` can only append a missing entry. This diverges from `getRenderableTimelineItems` whenever a new rendered record belongs before an existing one. For example, after a tool call has been admitted and its card rendered, an assistant `message.replace` with `stopReason: "error"` creates the derived failure notice here; the live path appends that notice after the tool card, but a fresh render places it immediately after the assistant message. The transcript therefore changes order after reattach or recovery. Valid mixed append batches can be grouped out of sequence for the same reason.

**Suggested fix:** Process newly rendered items in canonical timeline sequence with insertion support, or fall back to full reconciliation whenever a targeted change would add an item before an existing rendered entry.

markustherobot (commented):
I found one transcript-ordering regression in the targeted reconciliation path. I reviewed both changed files and the relevant host, protocol, runtime, and rendering paths. I also launched two Luna subreviews covering reconciliation semantics and protocol/test coverage; both completed.

markusylisiurunen:
@markustherobot address the review comment. the guiding principle should be that user-facing behavior should not change given this PR (unless there is a clear bug, of course). for example, rendering a notice after some earlier card is not correct as it may very well not even be visible to the user if the earlier card is above the currently visible tail of the timeline.

src/tui/session_chat_controller.ts
markustherobot: Updated the targeted path to visit affected records in timeline sequence, which preserves mixed-batch append order. Late derived outcome notices intentionally remain appended at the visible tail, matching the existing live behavior rather than inserting them beside an older card. I added regressions for both cases, and the full suite passes.

PR #486, opened 2026-09-02
upgrade to Gemini 3.8 Flash

markusylisiurunen:
## why

Gemini 3.8 Flash supersedes the existing 3.7 Flash integration. Its model metadata is supplied through Tau's remote catalog until bundled `pi-ai` support is available.

## what

Replace the built-in 3.7 Flash persona with a catalog-gated 3.8 Flash chat persona. Upgrade Gemini transcription hint extraction and speech rewriting to 3.8 Flash, including tests and user documentation.


src/core/personas.ts
```diff
@@ -213,10 +213,11 @@ type PersonaSpec = {
...
 const PERSONA_SPECS: PersonaSpec[] = [
   {
-    id: "gemini-3.7-flash",
-    description: "Gemini 3.7 Flash",
+    id: "gemini-3.8-flash",
+    description: "Gemini 3.8 Flash",
     provider: "google",
-    modelId: "gemini-3.7-flash",
+    modelId: "gemini-3.8-flash",
+    catalogOnly: true,
```
markustherobot, line 220: **[P1] The new Gemini persona is unavailable**

`catalogOnly` removes this persona unless the merged model resolver can find `google:gemini-3.8-flash`. The live `https://pi.dev/api/models/providers/google` shard fetched with Tau's exact `pi/0.84.4` user agent currently contains `gemini-3.7-flash` but not `gemini-3.8-flash`; reproducing `createBuiltinPersonas` with that shard yields neither the old nor the new Gemini persona. A recently refreshed pre-3.8 cache can also retain this state for the four-hour freshness window even after the shard changes. As a result, the headline 3.8 chat persona is absent for normal users.

**Suggested fix:** Publish and verify a compatible Google catalog shard containing the complete 3.8 model metadata before shipping this change, accounting for existing fresh caches, or retain the 3.7 persona until 3.8 metadata is available through the supported catalog path.

markustherobot (commented):
I found one availability issue in the catalog-gated persona path. I launched two read-only subreviews covering persona/catalog resolution and Gemini speech/transcription paths; both completed successfully, and the direct speech changes were otherwise consistent.

PR #489, opened 2026-09-05
allow Escape to cancel manual compaction

### LINKED ISSUE #487: pressing esc while manual compaction is running does not cancel it

markustherobot:
## why

Escape could not cancel manual compaction: the TUI ignored interruption while only compaction was active, and the protocol handler queued cancellation behind the compaction itself.

## what

Allow Escape during either manual compaction command and reset interruption deduplication when compaction settles. Send the host's active-work abort before waiting for the protocol mutation queue so pending compaction can unwind.

Add controller and protocol regressions covering both commands, repeated Escape presses, subsequent compactions, and cancellation while compaction holds the mutation queue. Add a real hosted-session regression covering interruption during final compaction persistence, snapshot and delta consistency, subsequent turns, and shutdown. Clarify manual compaction cancellation in the TUI documentation.

## details

Keep compaction serialized as a context mutation. Only delivery of the existing abort signal moves outside the queue; pending-input handling, interruption bookkeeping, and the protocol response remain serialized. Cancellation before history replacement retains the existing compaction error and operation-settlement behavior without introducing a new cancellation result contract. Once runtime history has been replaced and persistence succeeds, a late abort no longer converts committed compaction into a failed mutation or invalidates the live session.

fixes #487

### DIFF
--- a/src/host/session_protocol_handler.ts
+++ b/src/host/session_protocol_handler.ts
@@ -1132,8 +1132,9 @@ export class SessionProtocolHandler {
       return;
     }
 
+    // Cancellation must reach maintenance work while it holds the mutation queue.
+    const interrupted = state.session.interruptActiveWork();
     const result = await this.enqueueMutation(state, async () => {
-      const interrupted = state.session.interruptActiveWork();
       if (interrupted) {
         state.live.interrupting = true;
       }
--- a/docs/tui.md
+++ b/docs/tui.md
@@ -51,7 +51,7 @@ When Tau is idle, Enter and Ctrl+Enter both start a normal turn. While a turn is
 
 Pending input is session state shared by attached clients while the host remains alive. It is not durable across host restart or session recovery. A queued turn captures the persona, reasoning, tools, and model settings when that turn actually starts. Steering remains part of the active logical turn and keeps the settings captured when that turn began.
 
-Escape interrupts foreground client work or the main session’s active work. If a local diff review, recording, or speech playback task owns the foreground, Escape stops that task first; otherwise it requests main-session interruption from the host, including cancellation of a running manual compaction. It does not stop independently running supervised subagents; select one with Alt+Down and use Ctrl+G. Press Escape twice to clear the current editor text.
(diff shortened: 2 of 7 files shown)


src/host/session_protocol_handler.ts
```diff
@@ -1132,8 +1132,9 @@ export class SessionProtocolHandler {
       return;
     }
 
+    // Cancellation must reach maintenance work while it holds the mutation queue.
+    const interrupted = state.session.interruptActiveWork();
```
markustherobot, line 1136: **[P1] Escape during final persistence can leave the session unusable**

Moving the abort outside the queue also makes it possible to interrupt while `compaction_end` is persisting an already-replaced runtime history. After that persistence completes, `LocalHostedSession.compact()` executes its post-call `signal.throwIfAborted()`, and `runAgentStateMutation()` sees the changed runtime state and permanently invalidates the live projection. I reproduced this by pausing the final snapshot commit, calling `interruptActiveWork()`, and releasing persistence: compaction rejected with `AbortError`, subsequent `snapshot()` calls rejected with `session state is inconsistent after a failed mutation`, and host shutdown failed too. The new fake-session regression cannot exercise this commit boundary.

**Suggested fix:** Distinguish cancellation before the compaction commit from cancellation after it. Once runtime compaction has successfully committed, do not turn a late abort into a failed state mutation (for example, remove the redundant post-commit abort check while retaining cancellation before history replacement). Add a real hosted-session regression that interrupts during final compaction persistence and verifies that the session remains usable.

markustherobot (commented):
I found one cancellation race that can leave the live session unusable after compaction commits. TypeScript compilation and all 271 focused tests passed; I reproduced the failure separately with the real hosted session by pausing final snapshot persistence.

I launched one review subagent covering host/protocol cancellation and queue semantics, and independently verified the reported race.

markusylisiurunen:
@markustherobot can you address the unresolved review comment

src/host/session_protocol_handler.ts
markustherobot: Fixed in b8ff17a. I removed the redundant host-side post-commit abort check; the runtime still checks cancellation before replacing history. The new real hosted-session regression pauses final compaction persistence, interrupts, and verifies successful completion, matching persisted and delta-applied snapshots, a subsequent turn, and clean shutdown. It reproduced the AbortError before the fix. Formatting/typechecks, build, and all 1,276 tests pass.

PR #493, opened 2026-09-08
add saved prompt selection to Telegram

### LINKED ISSUE #491: allow using saved prompts on Telegram

markustherobot:
## why

Saved prompts are available to Telegram sessions but previously had no Telegram interaction. The agreed flow in #491 adds a prompt to the conversation without immediately starting or interrupting an agent turn.

## what

Add `/prompt` with a paginated inline picker backed by the active session’s prompt catalog. Selection lazily resolves the current body and records it through the existing SDK operation, only while idle. After recording succeeds, the picker is replaced with the prompt and an instruction to send another message when ready; long text continues in additional messages.

Pickers are chat- and session-bound and become non-actionable after success. Duplicate taps and pickers from replaced sessions cannot append another message. Regression coverage exercises pagination, single-use selection, stale sessions, resolution failures, and work starting during lazy resolution.

## details

The interaction uses `/prompt` without an ID argument, matching the agreed button-selection flow. Only the newest picker per chat remains active, pages contain up to 20 prompts, and runner restarts expire pickers. Recorded prompts remain ordinary durable user messages, not drafts. Selection leaves pending attachments and group context for the next normal submission.

The manager checks both runner activity and host snapshot lifecycle/maintenance state before recording. This reuses the existing session protocol rather than introducing a new recording mode.

Registering the eleventh built-in command reduces the maximum visible projects per bot from 90 to 89 under Telegram’s 100-command limit. Configuration validation and documentation reflect that change. The detailed interaction documentation lives on the existing prompts page, with a link from the Telegram command table, to stay within the packaged documentation size limit.

fixes #491

### DIFF
--- a/src/core/telegram/adapter.ts
+++ b/src/core/telegram/adapter.ts
@@ -1,3 +1,4 @@
+import { randomUUID } from "node:crypto";
 import { mkdtemp, readdir, rm, writeFile } from "node:fs/promises";
 import { tmpdir } from "node:os";
 import { basename, extname, join } from "node:path";
@@ -120,6 +121,12 @@ export type TelegramApi = {
     allowedUpdates: TelegramAllowedUpdates;
   }): Promise<TelegramUpdate[]>;
   sendMessage(chatId: number, text: string, options: TelegramSendOptions): Promise<void>;
+  editMessage(
+    chatId: number,
+    messageId: number,
+    text: string,
+    options: TelegramSendOptions,
+  ): Promise<void>;
   sendRichMessage(chatId: number, markdown: string, options: TelegramSendOptions): Promise<void>;
   sendVoice(chatId: number, voice: Buffer, options: TelegramSendOptions): Promise<void>;
   sendChatAction(chatId: number, action: string): Promise<void>;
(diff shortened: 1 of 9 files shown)


src/core/telegram/adapter.ts
```diff
@@ -1804,7 +1837,11 @@ class TelegramAdapterImpl {
       return;
     }
 
-    const callbackHandled = await this.handleCallback(chat.id, callbackData);
+    const callbackHandled = await this.handleCallback(
+      chat.id,
+      callbackData,
+      callbackQuery.message?.message_id,
```
markustherobot, line 1843: **[P1] Every real picker tap is treated as expired**

`TelegramCallbackQuerySchema` parses the callback's `message` with only `{ chat }`, and `telegramPartialObject` strips unknown fields. Consequently, Telegram's `message_id` is discarded before this handler runs. This argument is always `undefined` for real updates, so `handlePromptCallback` rejects both selection and pagination as expired: no saved prompt can actually be recorded. The new tests inject updates through a mocked `getUpdates`, bypassing that schema. I reproduced the field loss using the compiled production parser.

**Suggested fix:** Preserve `message_id` in the callback-message schema and add a regression test that passes a Telegram callback payload through the normal API parsing path before exercising selection and pagination.

markustherobot (commented):
I found one blocking integration issue. The build and 133 targeted tests pass, but the picker tests bypass the production callback parser.

I reviewed all eight changed files and used two subreviews covering picker lifecycle/API handling and idle-recording/session semantics.

markusylisiurunen:
@markustherobot address the review comment

src/core/telegram/adapter.ts
markustherobot: Fixed in 85a3129. I preserved message_id in the callback-message schema and moved the picker regression through the production HTTP/API parser. It verifies pagination, selection, the outgoing message ID, duplicate-tap rejection, and no turn submission. The regression fails without the schema fix and passes with it; all 121 targeted Telegram tests pass.

PR #497, opened 2026-09-15
add session-owned background Bash jobs

### LINKED ISSUE #495: background bash
currently the agent can use something like tmux to e.g. run a dev server in the background. this works, but the issue is that the agent can easily leave resources running. we need to trust the agent to clean up any started resources during its turn, every time. this is also wasteful if e.g. iterating on UI if the agent has to always start and destroy dev server.

to address this, we could allow the agent to put bash to background and the Tau session can then clean up shit when it closes.

markustherobot:
## why

Long-running commands should be reusable across agent turns without relying on tmux or requiring the agent to clean up every process manually. The host needs to own their lifecycle, independently of client connections.

## what

Adds `bash({background: true})` and `list_bash_jobs`, `read_bash_job`, `stop_bash_job`, and `wait_for_bash_jobs`. Launch returns a job ID, not an application-readiness guarantee. Foreground execution keeps its existing behavior; background execution has no lifetime deadline and rejects timeout at launch.

Jobs share one live session registry across main agents, supervised agents, and ephemeral threads. They survive turn cancellation and observer release, retain bounded logs, and keep released sessions alive without blocking subsequent turns. Orderly disposal stops managed process groups before waiting for their completion. Job records are not persisted or recovered. The tool description tells agents to inspect current process and output state before restarting an unknown job, rather than assuming it never ran or is no longer running.

Waits return when any requested job finishes or the deadline expires. The default is 60,000ms, with positive-integer validation and a 300,000ms cap. Expiry is a successful observation; neither expiry nor interruption stops the jobs.

Adds backend launch/output callbacks, dedicated Cloudflare command sessions so servers cannot block foreground commands in the same directory, and local/Sprite process-group cleanup coverage. Also fixes Sprite cancellation losing its SIGKILL escalation when the shell exits before a TERM-ignoring child. Updates tool/session documentation and contributor ownership guidance.

## details

The four management tools are bundled with the `bash` capability rather than independently configurable persona entries. The registry retains up to 64 records, evicts older completed records when needed, and rejects launches when all slots are running. Each job retains a 64 KiB output tail; observations return up to 2,048 estimated output tokens per job within an 8,192-token shared output budget. Background calls reject `timeout` and `maxOutputTokens` rather than silently ignoring them. Read, stop, and wait accept `includeOutput: false` for metadata-only responses; exit status, termination reasons, and operational errors remain visible regardless of output selection. Output defaults to included for these calls. Management cards use Bash-scoped operation headings with stable subjects and conventional seven-line middle-truncated result previews. Launch retains `started background <command>`.

Coverage includes shared-agent access and cwd scoping, turn cancellation, timed waits, rejected background launch deadlines, bounded output and records, cleanup errors, session isolation, observer release/reconnect, natural exit, and shutdown with child processes. Local and Sprite tests execute real processes; Cloudflare coverage uses a mocked bridge, not a live deployment. Job-management failures expose bounded diagnostics in tool cards, while ordinary cancellation stays quiet. Regression coverage includes validation, stale IDs, cleanup diagnostics, cancellation, and persisted TUI facets with ordered deltas. All 1,320 tests pass in the final single-worker run. Earlier full runs intermittently hit timing-sensitive scratch-file and Telegram attachment tests; those tests were left unchanged.

As agreed, cleanup is limited to orderly closure. There is no crash supervisor, restart reattachment, or guarantee for independently daemonized processes or arbitrary tmux sessions. No product questions remain open.

fixes #495


markustherobot (commented):
I reviewed the complete snapshot at `34ec3a6` and found one non-blocking error-reporting issue in the new job-management tools.

I launched two review subagents covering registry/host lifecycle and process backends, then verified their candidates against the production call paths. Typechecking, build, and all 1,306 tests passed.

src/core/tools/bash_jobs.ts
```diff
@@ -0,0 +1,331 @@
...
+                  subject: "Bash jobs",
+                  details: text.split("\n").map((text) => ({ text })),
+                }),
+              },
+            };
+          } catch (error) {
+            return createTextToolOutcome(
+              error instanceof Error ? error.message : String(error),
+              context.signal.aborted ? "cancelled" : "failed",
+            );
```
markustherobot, line 326: **[P2] Job-management failures hide their diagnostics in the TUI**

When `read_bash_job`, `stop_bash_job`, or `wait_for_bash_jobs` encounters a stale ID or backend cleanup failure, this catch returns the diagnostic only as model-facing text without emitting a terminal presentation event. The runner has already installed the empty `describe()` presentation, and `getToolPresentation()` in `session_chat_controller.ts` uses that current-version facet rather than falling back to the result text. Consequently, the terminal shows only “failed Bash jobs,” hiding both the affected ID and the reason, including whether process cleanup failed. Invalid arguments have the same problem in the earlier return.

**Suggested fix:** Emit a bounded failure/blocked presentation containing the diagnostic on these paths, following the existing management tools, and cover a failed job operation through the host/TUI projection.

markusylisiurunen:
@markustherobot regarding the "Job-management failures hide their diagnostics in the TUI" review comment, how do e.g. subagent tools behave in similar situations?

markusylisiurunen:
@markustherobot address the finding so that you try to make these new bash job tools behave similarly to how subagent tools behave in similar situations. also, make sure these new bash tools use similar casing format (e.g. all lowercase or sentence case) as other existing tools in UI and tool output.

src/core/tools/bash_jobs.ts
markustherobot: Fixed in 1be0bbd. Validation now emits a blocked presentation with the reason; operational failures emit a finished error presentation with bounded diagnostics and the requested job IDs. Cancellation stays quiet, matching subagent tools. Added coverage for stale IDs, validation, cleanup-error truncation, cancellation, and persisted TUI facets reconstructed from ordered deltas. Check, build, and all 1,311 tests pass.

